PR #2, opened 2018-03-07
test: Setup dev server for rendering JS and Sass

lynnjepsen:
Allows each new package to setup it's own page in the dev server (aka http://localhost:8080/test/screenshot/foo-package/). Developers just need to update test/screenshot/webpack.config.js with their input Sass and JS, and output CSS and JS. Then they reference that output CSS and JS in an index.html


moog16:
I think having an index.html file for each component screenshot test will be a lot of unnecessary duplicate work. I think if we utilize React it'll greatly simplify adding/removing screenshot tests. Yes we will have only 1 bundle.css and bundle.js file, but the end user (the team) all have fast connections and computers. We should really be driving for faster/easier development. 

Also the screenshot tests lend themselves to using a SPA framework like React, similar to the demo pages. But we can chat about this more tomorrow :)

kfranqueiro:
I would suspect that the decision for separate bundles per test is VERY intentional; IIRC @acdvorak intends to do similar for our screenshot tests in the main repo. (Actually, I'm at least a little bit confused here... we ARE still doing screenshot tests in the main repo, right?)

Look at how long it takes the dev server in the main repo to initially start up and compile CSS, then talk to me again about fast computers. (It takes a full minute for me, which is frankly unacceptable, and I look forward to the day we alleviate that with better-organized bundles for screenshot test pages.)

test/screenshot/index.html
```diff
@@ -0,0 +1,3 @@
+<html>
+  <a href="temporary-package">Temporary Package</a>
+</html>
```
kfranqueiro: You're missing newlines at end of files FYI

moog16:
Well the start up for webpack does take a while, but the load time in a browser of a large bundle.js(~1MB) vs a small bundle.js(~100kB) served locally I would argue is not perceivable. If it was served across the wire to more people, then yes it would be a problem. But I think the use case is, download the repo and run the instance locally. The webpack bundling time is another issue altogether.

What I was saying is having an html file per screenshot test doesn't make sense to me. The purpose of React is to minimize the amount of scaffolding and setup you have. I suspect that in the Catalog pages we wouldn't have an individual html file per component, so I don't think we should here. The piece of info I'm missing and cannot speak to is how the screenshot tests are setup.

kfranqueiro:
Webpack rebuild time is especially important for developer experience, i.e. not having to wait 30 seconds just to check a change during development. Browser load time also matters for screenshot tests, since I think we're already wary of how long even our unit tests in the core repo take, and IIUC CrossBrowserTesting is _not_ a local network use case.

Meanwhile, I'm pretty sure I do remember us talking about intentionally having an html file per catalog page... I realize that's an atypical application of React, but I got the impression that using React for the catalog was more of (a) a templating solution and (b) a gateway to showcasing MDC React in the catalog, rather than an intentional direction toward making the catalog a SPA. The catalog's pages should load as quickly as possible. That all is tangential to the organization of this repo though.

lynnjepsen:
I updated this PR so it actually uses React (and added some new lines).

The start up time of this dev server is still almost 6 seconds, mostly because of babel-ifying react-dom. 

The reload time is fine right now....but theres not a lot of code. And after some investigations we cant really figure out when webpack decides to recompile stuff /shrug.

Matt is going to work on a separate PR which does the exact same thing as this PR....except it uses React Router instead of creating new index.html for each "page" in our dev server. 

We'll compare and contrast PRs once that is ready.

test/screenshot/webpack-bundles.js
```diff
@@ -0,0 +1,63 @@
+module.exports.sassBundle = function(sassInput, cssOutput) {
```
moog16: I think cutting this down from sassBundle and javaScript bundle into just 1 xyzBundle method will cut down the compile time. When you pass a new config into the array in multicompiler mode, webpack thinks it is a new bundle. Yes we want the css file to be separate, but I think by passing the `name` prop to the the `file-loader` will create a separate css file. 

The only drawback(up for debate) to having 1 bundle is that you must import sass files into the JS files. But I think this improves build times.

package.json
```diff
@@ -3,6 +3,7 @@
   "description": "Material Components React",
   "license": "Apache-2.0",
   "scripts": {
+    "start": "webpack-dev-server --config test/screenshot/webpack.config.js --content-base test/screenshot",
```
moog16: I don't think we can do the --content-base option here. I think we need the webpack.config.js file to live in the root. I don't think it transform the files in the packages dir. Try importing a React Component into the screenshot tests, and I don't think you'll get JSX errors.

lynnjepsen: nope!

PR #3, opened 2018-03-09
test: capture screenshots with capture-chrome, and diff them with resemble

lynnjepsen:


.babelrc
```diff
@@ -0,0 +1,5 @@
+{
+  "presets": [
+    "es2015"
```
moog16: can we also add "react" here? We don't have any jsx files yet, so we can add it then.

test/screenshot/screenshot.js
```diff
@@ -0,0 +1,62 @@
...
+    test(this.urlPath_, () => {
+      const capturePromise = capture({
+        url: 'http://localhost:8080/' + this.urlPath_,
+      });
+      const readPromise = readFilePromise(
+          './test/screenshot/' + this.imagePath_);
+      return Promise.all([capturePromise, readPromise])
+      .then(function([newScreenshot, oldScreenshot]) {
+        return new Promise(function(resolve) {
+            const resolve2 = function(data) {
```
moog16: more appropriate name is `onComplete`

test/screenshot/screenshot-suite.js
```diff
@@ -0,0 +1,22 @@
+export default class ScreenshotSuite {
```
moog16: This class might be better described as `Screenshots`, what do you think?

.babelrc
lynnjepsen: I think i want to wait, because this babelrc is actually affecting our babel translator for the mocha. So basically this is running on like capture-suite.js

moog16: Ya, but I don't think it affects .js files that don't have JSX in it. It basically compiles the JSX -> JS. Although I'm testing the branch again right now and having issues.

PR #5, opened 2018-03-12
test/puppeteer

moog16:


test/screenshot/screenshot.js
```diff
@@ -59,4 +52,23 @@ export default class Screenshot {
       });
     });
   }
+
+  createScreenshotTask_(url, path) {
```
patrickrodee: You could make this an `async` method (`async createScreenshotTask_(url, path) {}`) then remove the returned Promise and the wrapped async method. It would clean it up a lot.

```
async createScreenthotTask_(url, path) {
  let image;
  const browser = await puppeteer.launch();
  const page = await browser.newPage();
  await page.goto(url);
  if (path) {
    image = await page.screenshot({path});
  } else {
    image = await page.screenshot();
  }
  await browser.close();
  return image;
}

moog16: Ya good suggestion. I thought I could do something like that but am not familiar with async/await syntax. Any other places that async/await could clean up?

patrickrodee: Yeah, quite a few spots actually. I can make a PR later to fix it in those spots.

moog16:
Closing this since we did this change in another PR

PR #9, opened 2018-03-15
test: add Travis CI

patrickrodee:


.travis.yml
```diff
@@ -0,0 +1,15 @@
+language: node_js
+node_js: 8
+dist: trusty
+sudo: required
+git:
+  depth: false
+branches:
+  only:
+    - experimental/travis
```
moog16: Should this be _all`?

patrickrodee: Not yet! Just want to test this branch for now. It'll get changed to "all" once this is stable.

PR #12, opened 2018-03-22
feat: remote hosted screenshot testing

patrickrodee:
Remove repo screenshots in favor of Google Cloud Storage-hosted images. Add `golden.json` file to keep track of the golden image per tested URL. Increase line length to 120. Remove image upload step from Travis CI.


test/screenshot/golden.json
```diff
@@ -0,0 +1,4 @@
+{
+  "temporary-package/index.html": "57de91e42bd22a845074bae80f71eca3902ab7c6719f129960bc9018c79db95a",
+  "temporary-package/foo.html": "91f0795700eaba345092dd7507e046daa0d33e1136bad7992011ab454ddb1faf"
+}
```
acdvorak: Nit: Add trailing newline to end of file (and probably update your editor config to do this automatically)

package.json
```diff
@@ -5,7 +5,7 @@
   "scripts": {
     "start": "webpack-dev-server --config test/screenshot/webpack.config.js --content-base test/screenshot",
     "stop": "./test/screenshot/stop.sh",
-    "capture": "./node_modules/.bin/mocha --compilers js:babel-core/register --ui tdd test/screenshot/capture-suite.js",
+    "capture": "COMMIT_HASH=$(git rev-parse --short HEAD) BRANCH_NAME=$(git rev-parse --abbrev-ref HEAD) mocha --compilers js:babel-core/register --ui tdd --timeout 30000 test/screenshot/capture-suite.js",
```
acdvorak: If we want this to run on Windows, we'll need to use `cross-env` (see MDC Web's package.json for examples).

Also, keep in mind that there may be unstaged changes in the working directory that won't be reflected in `COMMIT_HASH`. This may end up biting us down the road if we forget that fact and assume that screenshots with a particular commit hash are from that exact commit with no other changes.

test/screenshot/screenshot.js
```diff
@@ -1,60 +1,220 @@
...
-import {readFile, writeFile} from 'fs';
 import {assert} from 'chai';
+import Storage from '@google-cloud/storage';
 
 import comparisonOptions from './screenshot-comparison-options';
 
 const readFilePromise = promisify(readFile);
 const writeFilePromise = promisify(writeFile);
 
+const SERVICE_ACCOUNT_KEY = process.env.SERVICE_ACCOUNT_KEY;
```
acdvorak: I would suggest prefixing all of our env vars with `MDC_`. This makes it easier to tell which part of the stack is using the var, and makes it easier to search/replace.

Also, we might want to rename this var to indicate that it's used by GCP.

E.g., `MDC_GCP_SERVICE_ACCOUNT_KEY` or `MDC_GCLOUD_SERVICE_ACCOUNT_KEY`.

test/screenshot/screenshot.js
```diff
@@ -1,60 +1,220 @@
...
+const GOLDEN = 'golden';
+const SNAPSHOT = 'snapshot';
+const DIFF = 'diff';
+
+let credentials;
+try {
+  credentials = JSON.parse(SERVICE_ACCOUNT_KEY);
+} catch (err) {
+  console.error('Service account key could not be parsed.');
+  process.exit(1);
```
acdvorak: This will break any future unit tests. Maybe just rethrow the error instead?

test/screenshot/screenshot.js
```diff
@@ -1,60 +1,220 @@
...
+  }
+
+  /**
+   * Generates a unique hash from an image's contents
+   * @param {!Buffer} imageBuffer The image buffer to hash
+   * @return {string}
+   * @private
+   */
+  generateImageID_(imageBuffer) {
+    return createHash('sha256').update(imageBuffer).digest('hex');
```
acdvorak: Why not sha512?

Changing your hashing algorithm in the future is painful (just ask git or npm), so we should pick the "right" one from the start.

test/screenshot/screenshot.js
```diff
@@ -1,60 +1,220 @@
...
+    return imageBuffer;
+  }
+
+  /**
+   * Generates a unique hash from an image's contents
+   * @param {!Buffer} imageBuffer The image buffer to hash
+   * @return {string}
+   * @private
+   */
+  generateImageID_(imageBuffer) {
```
acdvorak: Can we call these an "image hash" instead of "image ID"?

test/screenshot/screenshot.js
```diff
@@ -1,60 +1,220 @@
...
+  /**
+   * Returns the correct image path
+   * @param {string} imageId The image ID
+   * @param {string} imageType The image type
+   * @private
+   */
+  getImagePath_(imageId, imageType) {
+    switch (imageType) {
+      case GOLDEN:
+      return `${this.urlPath_}/${imageId}.golden.png`;
```
acdvorak: Nit: Indent (and maybe add an eslint rule for it?)

test/screenshot/screenshot.js
```diff
@@ -1,60 +1,220 @@
...
+
+      case SNAPSHOT:
+      case DIFF:
+      return `${this.urlPath_}/${COMMIT_HASH}/${imageId}.${imageType}.png`;
```
acdvorak: Nit: Indent

test/screenshot/screenshot.js
```diff
@@ -1,60 +1,220 @@
...
+    }
+  }
+
+  /**
+   * Saves the given image to Google Cloud Storage with optional metadata
+   * @param {string} imagePath The path to the image
+   * @param {!Buffer} imageBuffer The image buffer
+   * @param {Object} customMetadata Optional custom metadata
```
acdvorak: Nit: `@param {!Object=}`

test/screenshot/screenshot.js
```diff
@@ -1,60 +1,220 @@
...
+
+  /**
+   * Captures a screenshot of the test URL
+   */
   capture() {
     test(this.urlPath_, async () => {
-      const url = `http://localhost:8080/${this.urlPath_}`;
-      const imagePath = `./test/screenshot/${this.imagePath_}`;
-      await this.createScreenshotTask_(url, imagePath);
+      const golden = await this.createScreenshotTask_();
```
acdvorak: Rename to `goldenTask` or `screenshotTask`?

test/screenshot/screenshot.js
```diff
@@ -1,60 +1,220 @@
...
+   * @return {Buffer}
+   * @private
+   */
+  async readImage_(gcsFilePath) {
+    const data = await bucket.file(gcsFilePath).download();
+    return data[0];
+  }
+
+  /**
+   * Captures a screenshot of the test URL
```
acdvorak: `Captures a screenshot of the test URL and marks it as the new golden image`

test/screenshot/screenshot.js
```diff
@@ -1,60 +1,220 @@
...
+   */
   capture() {
```
acdvorak: The name `capture` is misleading. It would be clearer to name it something like `updateGolden`.

test/screenshot/screenshot.js
patrickrodee: No reason, I just chose the first one that came to mind

test/screenshot/screenshot.js
patrickrodee: Could rename the method to `takeScreenshot_` since it returns a Buffer, not a task.

test/screenshot/screenshot.js
```diff
@@ -1,60 +1,213 @@
...
-import {readFile, writeFile} from 'fs';
 import {assert} from 'chai';
+import Storage from '@google-cloud/storage';
 
 import comparisonOptions from './screenshot-comparison-options';
 
 const readFilePromise = promisify(readFile);
 const writeFilePromise = promisify(writeFile);
 
+const SERVICE_ACCOUNT_KEY = process.env.MDC_GCLOUD_SERVICE_ACCOUNT_KEY;
```
lynnjepsen: dont need all caps constants

test/screenshot/screenshot.js
```diff
@@ -1,60 +1,213 @@
...
-    this.imagePath_ = `${urlPath}.golden.png`;
-    this.snapshotImagePath_ = `${urlPath}.snapshot.png`;
-    this.diffPath_ = `${urlPath}.diff.png`;
     // TODO allow clients to specify capture-chrome options, like viewport size
   }
 
+  /**
+   * Returns the golden hash
+   * @return {string|undefined}
+   * @private
```
lynnjepsen: sort private methods to the bottom of the file

test/screenshot/screenshot.js
```diff
@@ -1,60 +1,213 @@
...
+   */
+  async getGoldenHash_() {
+    const goldenFile = await readFilePromise(GOLDENS_FILE_PATH);
+    const goldenJSON = JSON.parse(goldenFile);
+    return goldenJSON[this.urlPath_];
+  }
+
+  /**
+   * Saves the golden hash
```
lynnjepsen: hehehe, golden hash

test/screenshot/screenshot.js
acdvorak: Are you sure?

The [Google style guide](https://google.github.io/styleguide/jsguide.html#naming-constant-names) says that constants should be all uppercase.

test/screenshot/screenshot.js
```diff
@@ -1,60 +1,209 @@
...
+  async getGoldenHash_() {
+    const goldenFile = await readFilePromise(goldenFilePath);
+    const goldenJSON = JSON.parse(goldenFile);
+    return goldenJSON[this.urlPath_];
+  }
+
+  /**
+   * Returns the correct image path
+   * @param {string} imageHash The image hash
+   * @param {string} imageType The image type
```
acdvorak: Can we make an `@enum` for this? It can have string values so that your existing usages will continue to work.

test/screenshot/screenshot.js
```diff
@@ -1,60 +1,209 @@
...
+      const diffPath = this.getImagePath_(snapshotHash, 'diff');
+      const metadata = {golden: goldenHash};
+
+      // Save the snapshot and the diff
+      await Promise.all([
+        this.saveImage_(snapshotPath, snapshot, metadata),
+        this.saveImage_(diffPath, diff, metadata),
+      ]);
+
+      return assert.isBelow(Number(data.misMatchPercentage), 0.01);
```
acdvorak: Make a constant for `.01`.

test/screenshot/screenshot.js
```diff
@@ -1,60 +1,209 @@
...
+    const [exists] = await file.exists();
+    if (exists) {
+      console.log('✔︎ No changes to', imagePath);
       return;
+    }
+
+    // Create a new stream from the image buffer
+    let stream = new Readable();
+    stream.push(imageBuffer);
+    stream.push(null);
```
acdvorak: Maybe add a comment on this line explaining that [`null` indicates EOF](https://nodejs.org/api/stream.html#stream_readable_push_chunk_encoding)

test/screenshot/upload-screenshots.js
```diff
@@ -1,59 +0,0 @@
-/**
```
acdvorak: Don't forget to remove or update `npm run upload:screenshots` in `package.json`

PR #14, opened 2018-04-02
feat(fab): Add fab

williamernest:
Based off of the top app bar branch, so this PR is adjusted to that branch to show my diffs. Will be merged to master some time after top app bar is merged.


packages/fab/README.md
```diff
@@ -0,0 +1,27 @@
+# Top App Bar React
```
moog16: Change this to FAB

packages/fab/index.js
```diff
@@ -0,0 +1,46 @@
...
+    } = this.props;
+
+    return classnames('mdc-fab', Array.from(classList), className, {
+      'mdc-fab--mini': mini,
+    });
+  }
+
+  render() {
+    this.children = React.Children.map(this.props.children, (item) => {
+      const className = `${item.props.className} mdc-fab__icon`;
```
moog16: I think pulling this method out into a class method would make the logic clearer. Something like `this.addClassesToChild` or something like that.

packages/fab/index.js
```diff
@@ -0,0 +1,46 @@
...
+      const props = Object.assign({}, item.props, {className: className});
```
moog16: You can just write `{className}`

packages/fab/index.js
```diff
@@ -0,0 +1,46 @@
...
+          {this.children}
+        </button>
+    );
+  }
+}
+
+Fab.propTypes = {
+  mini: PropTypes.bool,
+  className: PropTypes.string,
+  children: PropTypes.element,
```
moog16: I looked more into this and I think we should have a default for it. Can you try this: ```
propTypes: {
    children: React.PropTypes.oneOfType([
      React.PropTypes.arrayOf(React.PropTypes.node),
      React.PropTypes.node
    ])
  },
  getDefaultProps: function() {
    return {
      children: null // or [] I guess
    };
  },
```

File: test/screenshot/fab/standard.js [hunk 5]
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
+import ReactDOM from 'react-dom';
+import Fab from '../../../packages/fab';
+
+import '../../../packages/fab/index.scss';
+import './index.scss';
+
+ReactDOM.render((
+    <div>
+      <Fab className="demo-button"><i className="material-icons">favorite</i></Fab>
```
moog16: These would be more readable to me as:

```
<Fab ..>
  <i ....>whatever</i>
</Fab>
```

What do you think. If so we should add a eslint rule or something

packages/fab/README.md
```diff
@@ -0,0 +1,27 @@
...
+```
+
+
+## Props
+
+Prop Name | Type | Description
+--- | --- | ---
+className | String | Classes to be applied to the root element.
+mini | n/a | Enables the mini variant.
+
```
moog16: Please add link to sass mixins https://github.com/material-components/material-components-web/blob/v0.34.1/packages/mdc-fab/README.md#advanced-sass-mixins

packages/fab/README.md
```diff
@@ -0,0 +1,32 @@
...
+## Installation
+
+```
+npm install @material/react-fab
+```
+
+## Usage
+
+### JSX Structure
```
lynnjepsen: JSX Structure is not necessary when it is this self-evident

packages/fab/index.js
```diff
@@ -0,0 +1,54 @@
...
+          {children}
+        </button>
+    );
+  }
+}
+
+Fab.propTypes = {
+  mini: PropTypes.bool,
+  className: PropTypes.string,
+  children: PropTypes.oneOfType([
```
lynnjepsen: I think Fab's propTypes should be similar to TopAppBar's propTypes, which allowed our clients to pass in their icons, or a MaterialIcon

packages/fab/package.json
```diff
@@ -0,0 +1,24 @@
+{
+  "name": "@material/fab-react",
+  "version": "0.0.0",
+  "description": "",
+  "main": "index.js",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
```
lynnjepsen: I don't think you need a scripts section, or most of this package.json, now that we have lerna setup

test/screenshot/golden.json
```diff
@@ -6,4 +6,4 @@
   "top-app-bar/shortCollapsed.html": "d95df9e94851ffc70ccc38daed159e274209c54f3d1d18961f4512504defda6f",
   "top-app-bar/standardWithNavigationIconElement.html": "f2171a6e85650530d4809dae75d3509f0bbdc7d4c5caf77af88dc9bc2437415d",
   "material-icon/menu.html": "790b2076760612c6c91ee0f6bb42aa3603b30c2a60f302de22429c2572645640"
-}
+}
```
lynnjepsen: there should be a new screenshot golden in here

packages/fab/index.js
williamernest: So pass it in using a prop, or continue to use a child element? The reason we used a prop for top app bar was the icons had to be placed 2 `div` elements down (inside a section that was within a row). Since this is such a simple element, does it make sense to make to use a prop when a direct child element does the same thing?

packages/fab/package.json
williamernest: What should these components be called? Currently we have `@material/top-app-bar-react` and `@material/fab-react`.

lynnjepsen: @material/fab-react

packages/fab/index.js
lynnjepsen: I guess I just dont understand the difference between PropTypes.node and PropTypes.element. Why does TopAppBar have elements, and Fab has nodes?

In TopAppBar we have this code:
```
  addClassesToElement(classes, element) {
    const propsWithClasses = {
      className: classnames(classes, element.props.className),
    };
    return React.cloneElement(element, propsWithClasses);
  }
```

Maybe we should pull that out of top app bar so we can use it in multiple components

kfranqueiro: FYI, `element` is more restrictive than `node`, IIUC. From prop-types' README:

```
  // Anything that can be rendered: numbers, strings, elements or an array
  // (or fragment) containing these types.
  optionalNode: PropTypes.node,

  // A React element.
  optionalElement: PropTypes.element,
```

moog16: We chose element since we need it to be an element that we can add a class to. As a FAB, it should always be an element (<a>, <i>, <svg>, etc). 

We may have issues when we get to extended Fab, but with regards to Fab `element` is what we want.

packages/fab/index.js
```diff
@@ -0,0 +1,53 @@
...
+      mini,
+      className,
+    } = this.props;
+
+    return classnames('mdc-fab', Array.from(classList), className, {
+      'mdc-fab--mini': mini,
+    });
+  }
+
+  setClassOnChildren = () => {
```
lynnjepsen: betterName: addIconClassToAllChildren

packages/fab/index.js
```diff
@@ -0,0 +1,53 @@
...
+  mini: PropTypes.bool,
+  className: PropTypes.string,
+  children: PropTypes.oneOfType([
+    PropTypes.arrayOf(PropTypes.element),
+    PropTypes.element,
+  ]).isRequired,
+};
+
+Fab.defaultProps = {
+  mine: false,
```
lynnjepsen: mini, not mine

PR #15, opened 2018-04-09
chore: Use `includePaths` instead of `importer` in `sass-loader` config

moog16:


test/screenshot/webpack-bundles.js
```diff
@@ -22,18 +22,7 @@ module.exports.bundle = function(testPath, outputPath) {
...
-                    let nodeModulePath = `./node_modules/@material/${filePath}`;
-                    return {
-                      file: require('path').resolve(nodeModulePath),
-                    };
-                  }
-                  return {
-                    file: url,
-                  };
-                },
+                includePaths: ['./node_modules']
```
acdvorak: Linter is complaining about a missing trailing comma

PR #16, opened 2018-04-11
feature(button): added new component with ripple HOC

moog16:


packages/button/README.md
```diff
@@ -0,0 +1,22 @@
...
+
+## Props
+
+Prop Name | Type | Description
+--- | --- | ---
+className | String | Classes to be applied to the root element.
+raised | n/a | Enables raised variant.
+unelevated | n/a | Enables unelevated variant.
+stroked | n/a | Enables stroked variant.
+dense | n/a | Enables dense variant.
```
lynnjepsen: lets not do dense yet...I think that API might change

packages/button/README.md
```diff
@@ -0,0 +1,22 @@
...
+icon | Element | Icon to render within root element.
+children | String | Text to be displayed within root element.
```
lynnjepsen: And then this should be like "textLabel"

packages/button/README.md
```diff
@@ -0,0 +1,22 @@
...
+## Props
+
+Prop Name | Type | Description
+--- | --- | ---
+className | String | Classes to be applied to the root element.
+raised | n/a | Enables raised variant.
+unelevated | n/a | Enables unelevated variant.
+stroked | n/a | Enables stroked variant.
+dense | n/a | Enables dense variant.
+icon | Element | Icon to render within root element.
```
lynnjepsen: I think the icon should be the children, since it is an element

packages/ripple/index.js
```diff
@@ -0,0 +1,174 @@
...
+        deregisterDocumentInteractionHandler: (evtType, handler) =>
+          document.documentElement.removeEventListener(evtType, handler, util.applyPassive()),
+        registerResizeHandler: (handler) => window.addEventListener('resize', handler),
+        deregisterResizeHandler: (handler) => window.removeEventListener('resize', handler),
+        updateCssVariable: this.updateCssVariable,
+        computeBoundingRect: () => instance.getBoundingClientRect(),
+        getWindowPageOffset: () => ({x: window.pageXOffset, y: window.pageYOffset}),
+      };
+
+      this.instance_ = instance;
```
lynnjepsen: saving this.instance_ within a "create" method is weird. Its a side effect of the "createAdapter" method that is unintuitive

packages/ripple/index.js
```diff
@@ -0,0 +1,174 @@
...
+      this.adapter_ = adapter;
```
lynnjepsen: Also, saving this.adapter_ within a "create" method is weird, also an unintuitive side effect

packages/ripple/index.js
```diff
@@ -0,0 +1,174 @@
...
+          break;
+        case 'onKeyDown':
+          action = onKeyDown;
+          break;
+      }
+
+      if (eventType !== 'onMouseUp') {
+        this.activateRipple(e);
+      }
+      action(e);
```
lynnjepsen: instead of saving the function in another variable (action) and then calling it...just call the original function. Much simpler.

packages/button/README.md
moog16: It is nonintuitive - I don't think being an element should dictate if its `children` or through a `prop`. We should go with what makes sense.

I like switching to `textLabel` and `icon` as props. Its more declarative. What do you think?

packages/ripple/index.js
moog16: Will remove

packages/ripple/index.js
moog16: What if...I called it `initializeFoundation`, which did this:

```
this.foundation_ = new MDCRippleFoundation({ ... adapter stuff ... });
this.foundation_.init();
```

packages/ripple/index.js
```diff
@@ -31,8 +27,14 @@ const withRipple = (WrappedComponent) => {
...
 
-    createAdapter = (instance) => {
-      const adapter = {
+    initializeFoundation_ = (instance) => {
+      const adapter = this.getAdapter_(instance);
+      this.foundation_ = new MDCRippleFoundation(adapter);
+      this.foundation_.init();
+    }
+
+    getAdapter_ = (instance) => {
```
lynnjepsen: createAdapter is more accurate as a name

packages/ripple/README.md
```diff
@@ -0,0 +1,19 @@
+# React Ripple
```
lynnjepsen: Show an example of how to wire ref to initRipple

packages/material-icon/index.js
```diff
@@ -2,34 +2,80 @@ import React from 'react';
 import classnames from 'classnames';
 import PropTypes from 'prop-types';
 
+import withRipple from '@material/react-ripple';
+
 export default class MaterialIcon extends React.Component {
```
lynnjepsen: i think it is weird to have `export default` but then later in the file have a non-default `export`

packages/material-icon/index.js
```diff
@@ -2,34 +2,80 @@ import React from 'react';
...
       icon,
-      className,
+      hasRipple,
       ...otherProps
     } = this.props;
-    const classes = classnames('material-icons', {
-      [className]: className.length > 0,
-    });
+
+    if (hasRipple) {
```
lynnjepsen: so this is happening in the render method...does that mean the entire MaterialIcon component will re-render *with* a ripple if someone changes the `hasRipple` prop to true at runtime?

packages/material-icon/index.js
```diff
@@ -2,34 +2,80 @@ import React from 'react';
...
-  className: PropTypes.string,
+  hasRipple: PropTypes.bool,
 };
 
 MaterialIcon.defaultProps = {
+  icon: '',
+  hasRipple: false,
+};
+
+MaterialIconDefault.propTypes = {
```
lynnjepsen: You don't actually export MaterialIconDefault, so I dont know if you need to specify propTypes.

packages/material-icon/index.scss
```diff
@@ -1 +1,9 @@
 @import url("https://fonts.googleapis.com/icon?family=Material+Icons");
+@import "@material/ripple/common";
+@import "@material/ripple/mixins";
+
+.material-icons {
+  @include mdc-ripple-surface;
```
lynnjepsen: I don't think we should add mdc-ripple-surface to all material-icons. We should have another `material-icons--ripple-surface` class which this is limited to. And then update the MaterialIcon.js to apply this `material-icons--ripple-surface` as needed

packages/ripple/README.md
```diff
@@ -17,3 +17,43 @@ unbounded | boolean | Ripple is unbounded if true.
...
+      {...otherProps}>
+      {children}
+    </div>
+  );
+};
+
+const RippleIcon = withRipple(Icon);
+```
+
+Wrap your Icon component with the HOC `withRipple`, which returns a new "rippled"
```
lynnjepsen: "which returns a component with a ripple capable surface"

..."rippled" :P

packages/top-app-bar/README.md
```diff
@@ -48,7 +48,8 @@ If you decide to use a React Component please see [Integrating with Components](
 
 > NOTE: `actionItems` prop is expecting an array of elements.
 
-Please see [MDC Web Top App Bar Readme](https://github.com/material-components/material-components-web/tree/master/packages/mdc-top-app-bar) for more information.
+> NOTE: If you use [Material Icon's](../material-icon/README.md) as Action Items or Navigation Icons, you will get the Ripple for free. Using custom elements will require you to
+> wrap the custom element with the [`withRipple` HOC](../ripple/README.md).
```
lynnjepsen: nope nope, If you use custom elements then you have to implement the hasRipple prop.

packages/top-app-bar/index.js
```diff
@@ -48,11 +48,17 @@ export default class TopAppBar extends React.Component {
...
-  addClassesToElement(classes, element) {
-    const propsWithClasses = {
+  addClassesAndEnableRippleOnElement(classes, element) {
+    const updatedProps = {
       className: classnames(classes, element.props.className),
     };
-    return React.cloneElement(element, propsWithClasses);
+
+    // If `element` is a Native React Element, we don't want to throw
+    // a warning.
```
lynnjepsen: I think we should have a failure if you give us a Native React Element. The child needs to implement the `hasRipple` property, so it has to be a proper React.Component

packages/top-app-bar/index.js
```diff
@@ -48,11 +48,17 @@ export default class TopAppBar extends React.Component {
     this.foundation_.init();
   }
 
-  addClassesToElement(classes, element) {
-    const propsWithClasses = {
+  addClassesAndEnableRippleOnElement(classes, element) {
```
lynnjepsen: methods are more likely to have bugs if they have the word "and" in their name.

Split this into two methods.

packages/material-icon/index.js
moog16: I need the other export for a test, otherwise its not necessary. I could put it in another file if that's better.

packages/material-icon/index.js
moog16: That's correct. It will be re-rendered, but that's ok since the underlying HTML will not change. Also in any other case we would expect it to re-render since its a new component. 

With React we shouldn't worry about re-rendering, since every time `props` or `state` changes, it goes through a render cycle. There are of course edge case, but this is what React was built for.

packages/material-icon/index.js
moog16: Lint yells at me otherwise :|

packages/top-app-bar/README.md
moog16: They will need the `withRipple` HOC. Even if they have `hasRipple` prop, it won't be attached to anything. In the example below, `hasRipple` won't do anything:

```js
const Icon = ({hasRipple}) => (
  <i className='icon-class-name' hasRipple >menu</i>
);
```

packages/top-app-bar/index.js
moog16: The Native React Components are still React Components. I don't think we want to be that restrictive here. We spent time in the TopAppBar PR to not limit this.

packages/top-app-bar/index.js
moog16: I believe that...where'd you hear that?

packages/material-icon/README.md
```diff
@@ -8,18 +8,17 @@ MDC React Material Icon is a component for material icons.
...
 icon | string | Type of icon to be displayed.
 className | string | Classes to pass on to the root `<i>` element.
+hasRipple | n/a | If present on element, it will enable the ripple on the icon.
+
+
+## CSS Classes
+
+CSS Class | Description
+--- | ---
+`material-icons--ripple-surface` | Adds ripple to the icon.
```
lynnjepsen: This is an internal implementation detail. you dont need to add it to the README

packages/top-app-bar/index.js
```diff
@@ -48,11 +48,23 @@ export default class TopAppBar extends React.Component {
     this.foundation_.init();
   }
 
-  addClassesToElement(classes, element) {
-    const propsWithClasses = {
+  addClassesOnElement(classes, element) {
```
lynnjepsen: addClassesToElement was better

PR #17, opened 2018-04-17
feature(cards): new component

moog16:


packages/card/README.md
```diff
@@ -0,0 +1,133 @@
+# React Card
+
+MDC React Card is a component for material design cards.
+
+## Installation
+
+```
+npm install @material/react-card
```
acdvorak: Are we publishing this to NPM right now? Or are you just planning ahead?

Also, the package names seem inconsistent; there's `react-card` and `react-fab`, but also `material-icon-react` and `top-app-bar-react`. Which one are we going with? (I vote for `react-foo`)

packages/card/README.md
```diff
@@ -0,0 +1,133 @@
...
+
+#### Props
+
+Prop Name | Type | Description
+--- | --- | ---
+fullBleed | n/a | Enables full bleed card actions row by removing all padding.
+
+### CardContent
+
+This component is used as the container for primary tappable content.
```
acdvorak: If this is intended to be the primary tappable area, it should be named "primary action" for consistency with MDC and the spec.

packages/card/README.md
```diff
@@ -0,0 +1,133 @@
...
+import Card, {CardContent} from '@material/react-card';
+
+<Card>
+  <CardContent>
+    <p>Content</p>
+  </CardContent>
+</Card>
+```
+
+### Card
```
acdvorak: Shouldn't this be the first component listed in the README?

I would probably reorder these components to something like:

1. Card
2. CardPrimaryAction
3. CardMedia
4. CardActions
5. CardActionButtons
6. CardActionIcons

packages/card/README.md
moog16: Just planning ahead.

And yes this PR is inconsistent, and is fixed in #16 for `react-foo` :)

packages/card/README.md
moog16: Good idea. I was going in order of how it displayed in my filesystem.

PR #19, opened 2018-04-20
feature(floating label): new component

moog16:


packages/floating-label/README.md
```diff
@@ -0,0 +1,68 @@
...
+onShakeEnd | Function | Callback method called when the shake animation ends.
+setWidth | Function | Callback method to pass the width to a parent Component.
+shouldFloat | Boolean | Floats label depending on value passed.
+shouldShake | Boolean | Shakes label depending on value passed.
+
+## Usage
+
+### Positioning
+
+React Floating Label is meant to be used with other Components, specifically input type elements. Floating label is positioned absolute, which requires the parent element to be position _relative_ or _absolute_.
```
lynnjepsen: Just say
which requires the parent element to be position _relative_

You don't have to explain that position: absolute makes the elements children

packages/floating-label/README.md
```diff
@@ -0,0 +1,68 @@
...
+```
+
+## Props
+
+Prop Name | Type | Description
+--- | --- | ---
+className | String | Classes to be applied to the root element.
+onShakeEnd | Function | Callback method called when the shake animation ends.
+setWidth | Function | Callback method to pass the width to a parent Component.
+shouldFloat | Boolean | Floats label depending on value passed.
```
lynnjepsen: just float, not "shouldFloat".

same for "shouldShake"

packages/floating-label/README.md
```diff
@@ -0,0 +1,68 @@
...
+shouldShake | Boolean | Shakes label depending on value passed.
+
+## Usage
+
+### Positioning
+
+React Floating Label is meant to be used with other Components, specifically input type elements. Floating label is positioned absolute, which requires the parent element to be position _relative_ or _absolute_.
+
+### Float Label
```
lynnjepsen: I dont think you really need this Float Label section at all....its already covered by the description of the property

packages/floating-label/README.md
```diff
@@ -0,0 +1,68 @@
...
+```
+
+### Shake Label
+
+Toggling the `shouldShake` prop to true will start the shake animation. Toggling the `shouldShake` prop to false while animation is animating, will cause animation to end. Once animation ends, the `onShakeEnd()` callback will be called if the parent component needs to be notified. Ex:
+
+```js
+<FloatingLabel
+  shouldShake={this.state.shouldShake}
+  onShakeEnd={() => this.setState({shouldShake: false})}
```
lynnjepsen: discussed offline that shake is better as a method than as a property

test/screenshot/floating-label/index.js
```diff
@@ -0,0 +1,83 @@
...
+              {shouldUsePrimaryLabel ? labelText[0] : labelText[1]}
+            </FloatingLabel>
+          </div>
+          {this.renderControls()}
+        </div>
+      </div>
+    );
+  }
+
+  renderControls() {
```
lynnjepsen: instead of having a button to test float, just have multiple instances of FloatingLabel on the same screenshot page.

test/screenshot/floating-label/index.js
```diff
@@ -0,0 +1,83 @@
...
+    const {width} = this.state;
+    const buttons = [{
+      text: 'Float',
+      action: () => this.setState({shouldFloat: !this.state.shouldFloat}),
+    }, {
+      text: 'Shake',
```
lynnjepsen: Don't cover this in screenshot test

test/screenshot/floating-label/index.js
```diff
@@ -0,0 +1,83 @@
...
+      action: this.floatingLabelElement_.current && this.floatingLabelElement_.current.shake,
+    }, {
+      text: 'Label Text',
```
lynnjepsen: test the "handleWidthChange" with a JS test

packages/floating-label/README.md
```diff
@@ -0,0 +1,71 @@
...
+
+  // ... later in code ...
+
+  this.floatingLabelElement.current.shake();
+}
+```
+
+### Label Width
+
+Label width is set during mount, and is calculated from the `offsetWidth` property. The Component will call `setWidth()` when mounted. `setWidth()` will execute with a new width value if `this.props.children` changes.
```
lynnjepsen: update this to say `handleWidthChange`

test/screenshot/floating-label/index.js
```diff
@@ -0,0 +1,52 @@
...
+
+import '../../../packages/floating-label/index.scss';
+import './index.scss';
+
+class Input extends React.Component {
+  state = {
+    shouldUsePrimaryLabel: true,
+    float: false,
+    width: 0,
+  };
```
lynnjepsen: You shouldnt have all this logic in your screenshot tests. Just render two components...then take a screenshot. No state necessary.

PR #22, opened 2018-04-24
feature(notched outline): new component

moog16:


packages/notched-outline/index.js
```diff
@@ -0,0 +1,106 @@
...
+    this.foundation_.destroy();
+  }
+
+  componentWillReceiveProps(nextProps) {
+    const hasToggledNotch = this.props.notch !== nextProps.notch;
+    const hasToggleRtl = this.props.isRtl !== nextProps.isRtl;
+    const notchWidthUpdated = this.props.notchWidth !== nextProps.notchWidth;
+    const shouldUpdateNotch = notchWidthUpdated || hasToggleRtl || hasToggledNotch;
+
+    if (shouldUpdateNotch) {
```
lynnjepsen: nit, return quickly.

So:
```
if (!shouldUpdateNotch) {
  return;
}

// other code
```

It reduces the amount of nesting

PR #27, opened 2018-05-08
feature(text-field): Add new component

moog16:


packages/text-field/Input.js
```diff
@@ -0,0 +1,220 @@
...
+    const {value} = e.target;
+
+    foundation.autoCompleteFocus();
+    handleValueChange(value);
+    onChange(e);
+  }
+
+  render() {
+    const {
+      className, // eslint-disable-line no-unused-vars
```
patrickrodee: Replace all these with a leading `/* eslint-disable no-unused-vars */` and a trailing `/* eslint-enable no-unused-vars */`

packages/text-field/Input.js
```diff
@@ -0,0 +1,220 @@
...
+    if (this.props.id !== nextProps.id) {
+      nextProps.setInputId(nextProps.id);
+    }
+
+    this.updateValue(nextProps);
+  }
+
+  handleValidationAttributeUpdate = (nextProps) => {
+    const {foundation} = nextProps;
+    VALIDATION_ATTR_WHITELIST.some((attr) => {
```
patrickrodee: This will only update the first prop that changed and ignore the others. Is that what we want?

packages/text-field/Input.js
```diff
@@ -0,0 +1,220 @@
...
+  // The end developer and the foundation change the value at the same time,
+  // the end developer's value should always win.
+  updateValue = (nextProps) => {
+    const valueChanged = this.props.value !== nextProps.value;
+    const foundationValueChanged = this.props.foundationValue !== nextProps.foundationValue;
+
+    if (foundationValueChanged) {
+      this.setState({value: nextProps.foundationValue});
+    }
+    if (valueChanged) {
```
patrickrodee: I'd change the order of these and add an early-exit `return` statement in the `valueChanged` case since that's an exception. Relying on control flow fall-through here doesn't make sense.

packages/text-field/Input.js
```diff
@@ -0,0 +1,220 @@
...
+
+    foundation.autoCompleteFocus();
+    handleValueChange(value);
+    onChange(e);
+  }
+
+  render() {
+    const {
+      className, // eslint-disable-line no-unused-vars
+      disabled,
```
patrickrodee: Move this below the `/* eslint-enable no-unused-vars */` line.

packages/text-field/README.md
```diff
@@ -0,0 +1,69 @@
...
+  }
+}
+```
+
+> NOTE: In order to get access to the value, you must add an onChange handler, which accepts an event and updates the value of the input as shown above.
+
+
+### Input Props
+
+Prop Name | Type | Description
```
patrickrodee: Fill this out or drop it.

packages/text-field/helper-text/index.js
```diff
@@ -12,6 +12,7 @@ export default class HelperText extends React.Component {
     this.state = {
       'aria-hidden': props['aria-hidden'],
       'role': props.role,
+      'classList': new Set(),
```
patrickrodee: Are all these attributes are wrapped in single quotes to satisfy ESLint?

packages/text-field/icon/index.js
```diff
@@ -63,7 +63,7 @@ export default class Icon extends React.Component {
 
 Icon.propTypes = {
   children: PropTypes.element.isRequired,
-  disabled: PropTypes.boolean,
+  disabled: PropTypes.bool,
```
patrickrodee: Good catch

packages/text-field/index.js
```diff
@@ -0,0 +1,367 @@
...
+      value: null,
+      classList: new Set(),
+      inputId: props.children.props.id,
+      isFocused: false,
+      isBadInput: false,
+      dir: 'ltr',
+      disabled: false,
+
+      // floating label state
+      shouldFloatLabel: false, // needs to read off of props if should float
```
patrickrodee: If this should be read off props, should it be initialized via props?

packages/text-field/index.js
```diff
@@ -0,0 +1,367 @@
...
+      const dir = window.getComputedStyle(this.textFieldElement.current).getPropertyValue('direction');
+      return dir === 'rtl';
+    }
+  }
+
+  /**
+  * adapter methods
+  */
+
+  get adapter() {
```
patrickrodee: Nit: order the getters together (move this up)

packages/text-field/index.js
```diff
@@ -0,0 +1,367 @@
...
+      hasLabel: () => !!this.props.label,
+      getLabelWidth: () => this.state.labelWidth,
+    };
+  }
+
+  get lineRippleAdapter() {
+    return {
+      activateLineRipple: () => this.setState({activeLineRipple: true}),
+      deactivateLineRipple: () => this.setState({activeLineRipple: false}),
+      setLineRippleTransformOrigin: (lineRippleCenter) => this.setState({rippleCenter: lineRippleCenter}),
```
patrickrodee: Nit: rename `lineRippleCenter` to `rippleCenter` so you can use the variable-as-key assignment `this.setState({rippleCenter})`

packages/text-field/index.js
```diff
@@ -0,0 +1,367 @@
...
+
+    if (helperText) {
+      return ([
+        textField, this.renderHelperText(),
+      ]);
+    }
+    return textField;
+  }
+
+  inputProps = (props) => {
```
patrickrodee: Nit: Move instance variables like this closer to the top of the class. Also, I don't think this needs to be an arrow function and can instead be an instance method.

packages/text-field/index.js
```diff
@@ -0,0 +1,367 @@
...
+    return (
+      <Icon disabled={disabled}>
+        {icon}
+      </Icon>
+    );
+  }
+}
+
+TextField.propTypes = {
+  'box': PropTypes.bool,
```
patrickrodee: Remove the single quotes from these unless they're necessary.

test/screenshot/text-field/index.js
```diff
@@ -0,0 +1,95 @@
+import React from 'react';
+import ReactDOM from 'react-dom';
+
+import TextField, {Input} from '../../../packages/text-field';
+import MaterialIcon from '../../../packages/material-icon/index';
+
+import '../../../packages/text-field/index.scss';
+import './index.scss';
+
+class TestField extends React.Component {
```
patrickrodee: lololol, you're funny 😆

packages/text-field/index.js
```diff
@@ -0,0 +1,367 @@
...
+      trailingIcon,
+      textarea,
+    } = this.props;
+
+    const textField = (
+      <div
+        className={this.classes}
+        onClick={() => this.foundation_ && this.foundation_.handleTextFieldInteraction()}
+        onKeyDown={() => this.foundation_ && this.foundation_.handleTextFieldInteraction()}
+        key='text-field-container'
```
patrickrodee: Why does this need a `key`?

packages/text-field/Input.js
moog16: Yes that's right. Its actually similar to what we do in MDC-Web. https://github.com/material-components/material-components-web/blob/master/packages/mdc-textfield/foundation.js#L187

packages/text-field/Input.js
moog16: Good call

packages/text-field/helper-text/index.js
moog16: correct...only because of aria-hidden

packages/text-field/index.js
moog16: I think I had an idea a while ago, but don't need to do it.

packages/text-field/Input.js
patrickrodee: Awesome, thanks for the explanation!

packages/text-field/helper-text/index.js
patrickrodee: Good to know, thanks

packages/text-field/index.js
patrickrodee: Ok. We'll leave this as is.

packages/text-field/index.js
moog16: I don't like them either, but they are necessary for `children.props`

packages/text-field/index.js
moog16: Needs a key if its part of an array...so if the render method returns `[<TextField />, <HelperText />]`, they both need a key.

packages/text-field/index.js
patrickrodee: Gotcha. We can leave them then.

packages/text-field/Input.js
```diff
@@ -0,0 +1,223 @@
...
+      }
+    });
+  }
+
+  // There are 2 sources that can update the input's value. 1) The end developer
+  // 2) the mdc.textfield.foundation through `this.getNativeInput().value = value`.
+  // Because of this, the input's value needs to be on this.state. If
+  // The end developer and the foundation change the value at the same time,
+  // the end developer's value should always win.
+  updateValue = (nextProps) => {
```
patrickrodee: I don't think there's a convention here but I prefer the class instance fields (the `foo = ...` on the class body) to be declared above the constructor since that's closer to where they'd normally go.

packages/text-field/index.js
```diff
@@ -0,0 +1,367 @@
...
+      // validity.valid.
+      setBadInputHandler: (getBadInput) => this.getBadInput = getBadInput,
+      setIsValidHandler: (getIsValid) => this.getIsValid = getIsValid,
+      setDisabled: (disabled) => this.setState({disabled}),
+      setInputId: (id) => this.setState({inputId: id}),
+    });
+  }
+
+  componentDidMount() {
+    const foundationMap = {
```
patrickrodee: You can just pass this directly to the `MDCTextFieldFoundation` constructor.

packages/text-field/index.js
```diff
@@ -0,0 +1,367 @@
...
+      const dir = window.getComputedStyle(this.textFieldElement.current).getPropertyValue('direction');
+      return dir === 'rtl';
+    }
+  }
+
+  /**
+  * getters
+  */
+
+  get classes() {
```
patrickrodee: Nit: Move the getters above the constructor but below class field declarations to follow the MDCWeb structure. This would apply to all the new files in this PR.

moog16: Spoke offline: this change would be inconsistent with other components. We will have a follow up change/PR to update all classes to follow same structure. When we do this, we should also implement some lint rules to enforce this going further.

packages/text-field/Input.js
```diff
@@ -0,0 +1,223 @@
...
+  handleValueChange: PropTypes.func,
+  id: PropTypes.string,
+  onBlur: PropTypes.func,
+  onChange: PropTypes.func,
+  onFocus: PropTypes.func,
+  onMouseDown: PropTypes.func,
+  onTouchStart: PropTypes.func,
+  setBadInputHandler: PropTypes.func,
+  setDisabled: PropTypes.func,
+  setInputId: PropTypes.func,
```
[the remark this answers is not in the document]
moog16: I have setInputId to set `id` on the TextField, which uses it for the FloatingLabel. FloatingLabel needs it for the `for` attribute, eg.

```
  <input id='my-id' />
  <label for='my-id>label</label>
```

packages/text-field/index.js
```diff
@@ -0,0 +1,367 @@
...
+      isBadInput: false,
+      dir: 'ltr',
+      disabled: false,
+
+      // floating label state
+      shouldFloatLabel: false,
+      labelWidth: 0,
+
+      // line ripple state
+      lineRippleCenter: null,
```
lynnjepsen: This variable could be more clear

packages/text-field/index.js
```diff
@@ -0,0 +1,367 @@
...
+
+      // notched outline state
+      shouldNotchOutline: false,
```
lynnjepsen: outlineIsNotched is more in sync with activeLineRipple

packages/text-field/index.js
```diff
@@ -0,0 +1,367 @@
...
+
+      // helper text state
+      helperTextShowToScreenReader: false,
```
lynnjepsen: showHelperTextToScreenReader is better

packages/text-field/index.js
```diff
@@ -0,0 +1,367 @@
...
+    const {className, box, dense, outlined, fullWidth, textarea, trailingIcon, leadingIcon} = this.props;
+    return classnames('mdc-text-field', Array.from(classList), className, {
+      'mdc-text-field--outlined': outlined,
+      'mdc-text-field--textarea': textarea,
+      'mdc-text-field--fullwidth': fullWidth,
+      'mdc-text-field--disabled': disabled,
+      'mdc-text-field--with-trailing-icon': trailingIcon,
+      'mdc-text-field--with-leading-icon': leadingIcon,
+      'mdc-text-field--box': box,
+      'mdc-text-field--dense': dense,
```
lynnjepsen: ooo like this

packages/text-field/index.js
```diff
@@ -0,0 +1,367 @@
...
+      this.inputAdapter,
+      this.labelAdapter,
+      this.lineRippleAdapter,
+      this.notchedOutlineAdapter,
+    );
+  }
+
+  get inputAdapter() {
+    // value: string,
+    // disabled: boolean, --> doesn't need to be implemented since the <INPUT> handles it
```
lynnjepsen: File a GitHub issue to remove disabled from MDC Web interface?

packages/text-field/index.js
```diff
@@ -0,0 +1,367 @@
...
+        return input;
+      },
+    };
+  }
+
+  get labelAdapter() {
+    return {
+      shakeLabel: (shakeLabel) => {
+        const {floatingLabelElement: floatingLabel} = this;
+        if (shakeLabel && floatingLabel && floatingLabel.current) {
```
lynnjepsen: Nit: early return if shakeLabel is false

packages/text-field/index.js
```diff
@@ -0,0 +1,367 @@
...
+      notchOutline: () => this.setState({shouldNotchOutline: true}),
+      closeOutline: () => this.setState({shouldNotchOutline: false}),
+      hasOutline: () => !!this.props.outlined,
+    };
+  }
+
+  get helperTextAdapter() {
+    return {
+      showToScreenReader: () =>
+        this.setState({helperTextShowToScreenReader: !this.state.helperTextShowToScreenReader}),
```
lynnjepsen: You should set 'helperTextShowToScrernReader' to true, always

packages/text-field/index.js
```diff
@@ -0,0 +1,367 @@
...
+        onKeyDown={() => this.foundation_ && this.foundation_.handleTextFieldInteraction()}
+        key='text-field-container'
+        ref={this.textFieldElement}
+      >
+        {leadingIcon ? this.renderIcon(leadingIcon) : null}
+        {this.renderInput()}
+        {label && !fullWidth ? this.renderLabel() : null}
+        {outlined ? this.renderNotchedOutline() : null}
+        {!fullWidth && !textarea && !outlined ? this.renderLineRipple() : null}
+        {trailingIcon ? this.renderIcon(trailingIcon) : null}
```
lynnjepsen: Oooo I like this too

packages/text-field/index.js
```diff
@@ -0,0 +1,367 @@
...
+      value: null,
+      classList: new Set(),
+      inputId: props.children.props.id,
+      isFocused: false,
+      isBadInput: false,
+      dir: 'ltr',
+      disabled: false,
+
+      // floating label state
+      shouldFloatLabel: false,
```
lynnjepsen: 'labelIsFloated' is more in sync with 'activeLineRipple'

packages/text-field/index.js
```diff
@@ -0,0 +1,367 @@
...
+      </FloatingLabel>
+    );
+  }
+
+  renderLineRipple() {
+    const {lineRippleClassName} = this.props;
+    const {activeLineRipple, lineRippleCenter} = this.state;
+    return (
+      <LineRipple
+        rippleCenter={lineRippleCenter}
```
lynnjepsen: Okay.. maybe 'lineRippleCenter' is as clear as 'rippleCenter'...not sure what units this is in though

packages/text-field/index.js
```diff
@@ -0,0 +1,367 @@
...
+  }
+
+  renderIcon(icon) {
+    const {disabled} = this.state;
+    // Toggling disabled will trigger icon.foundation.setDisabled()
+    return (
+      <Icon disabled={disabled}>
+        {icon}
+      </Icon>
+    );
```
lynnjepsen: I love all these render methods for sub-elements. They're adorable!

packages/text-field/index.js
```diff
@@ -0,0 +1,367 @@
...
+  'className': PropTypes.string,
+  'dense': PropTypes.bool,
+  'floatingLabelClassName': PropTypes.string,
+  'fullWidth': PropTypes.bool,
+  'helperText': PropTypes.string,
+  'helperTextAriaHidden': PropTypes.bool,
+  'helperTextClassName': PropTypes.string,
+  'helperTextIsValidationMessage': PropTypes.bool,
+  'helperTextPersistent': PropTypes.bool,
+  'helperTextRole': PropTypes.string,
```
lynnjepsen: Should we consider having the end developer pass us a full Helper text component? Instead of these 6 props?

test/screenshot/golden.json
```diff
@@ -14,5 +14,6 @@
   "floating-label/index.html": "67d9b6caad3a494563e2584ac6af071bbb3b73afc3e945f8299ca720b4015256",
   "notched-outline/index.html": "73394b6d7dcd94bc930dc740db8199000b94d365d7d703381ef01b3318b97d8c",
   "text-field/icon/index.html": "ce761ca24e2d5ca425ad1783c9e2d476b84f65232d86c4fdc0e8c3e756311c8f",
-  "text-field/helper-text/index.html": "722a8eb27cbde5e32fc92c70e6c107f8ef75bff1abb3914fd2ef58fd49b678bb"
+  "text-field/helper-text/index.html": "722a8eb27cbde5e32fc92c70e6c107f8ef75bff1abb3914fd2ef58fd49b678bb",
+  "text-field/index.html": "8cac264e77fe54efcc659edeb8ef1e530963f9efbe8a26edd2e54780beafcb71"
```
lynnjepsen: This screenshot is adorable

packages/text-field/Input.js
```diff
@@ -0,0 +1,195 @@
...
+  render() {
+    const {
+      disabled,
+      /* eslint-disable no-unused-vars */
+      className,
+      foundation,
+      value,
+      handleFocusChange,
+      handleValueChange,
+      setBadInputHandler,
```
lynnjepsen: Should this prop be 'handleBadInputChange'?

moog16: I don't think so, because this method is passed a function that sets it on the Text Field. That function is what returns validity.badInput. We don't actually do any actions if badInput updates.

packages/text-field/index.js
moog16: I think so...good idea!

PR #28, opened 2018-05-08
feature: update package.json versions to most recent

moog16:


patrickrodee:
Are the failing changes still appearing as expected (looking correct in 0.35.1?)

PR #40, opened 2018-05-09
docs: update descriptions

lynnjepsen:
Updated the descriptions to be more consistent.


packages/material-icon/README.md
```diff
@@ -1,6 +1,6 @@
 # Material Icon React
```
williamernest: React Material Icon

PR #52, opened 2018-05-18
feature: Update MDC Web dependencies to  v0.35.1

### LINKED ISSUE #42: Upgrade MDC Web to ^0.35.0
Currently all packages are on 0.34.1, but should be using latest (v0.35.0).

Known issues:
Ripple is broken, and needs to wrap the `foundation.activate()` call in a `requestAnimationFrame`. Not sure why this is the case, so further investigation is needed.

moog16:
Fixes #42 

Updated all versions of MDC Web to v0.35.1. Also fixed ripple by wrapping it in a requestAnimationFrame(). We verified fix by checking on IE11 and Edge.

Added polyfill, since Object.assign does not work on IE11.


moog16:
I wasn't aware of the card/button/text-field changes. I do have an issue for the top-app-bar change. This is the start of a series of PRs to sync with 0.36.x. I wanted to upgrade all of these to the latest, then fix the ripple. 

RE tests: How does Wiz catch these changes with their tests? At the very least, I think the screenshot tests should catch a rename from stroked --> outline.

moog16:
Update: 

_RE: Button_
https://github.com/material-components/material-components-web-react/commit/ab1dd138f90d5cf565947835a3ced731372f4a92 merged in an incorrect screenshot of stroked button. I overlooked the stroked button, and thought all these changes were due to typography changes. 


_RE: Card_
Card is the flakiest test since there is a lot of text. I didn't catch the difference between the `--outlined` styles when fixing this one. 

_RE: Text Field_
I don't see a `removeAttr` adapter method. Is that the correct name?

kfranqueiro:
Whoops... the CHANGELOG comment about `removeAttr` might've been a little overly vague, I think it was on the icon subcomponent specifically.

moog16:
Ah gotcha - created https://github.com/material-components/material-components-web-react/issues/59

PR #60, opened 2018-05-21
fix: update all packages to current v0.1.0

moog16:
First release errored when pushing to git, but I didn't realize until trying to use lerna to create a CHANGELOG. Fix is to `git merge v0.1.0`.


moog16:
closing this as I rebased instead.

PR #62, opened 2018-05-22
Remove scss extensions from @material/mdc-* imports

johnelliott:
Hi, I would like to help with material-components-web-react.

The sass imports are handy, but I was having trouble importing SCSS from Button. I needed to directly import the css from the non-react package, basically duplicating the intent of the index.scss of react-button:
image

When I looked at [the icon component sass import](https://github.com/material-components/material-components-web-react/blob/a45c75a1fd8bc7b83fbe0775fc629c928bb1df0f/packages/material-icon/index.scss#L2) I noticed that it didn't have an extension. Going into node_modules and removing it from the button index.scss file fixed the issue and allowed me to just import the button styles with `@import "@material/button/mdc-button";`, so perhaps this was just an oversight.

I've never used learn before, so I may need help figuring out how to contribute to this codebase :/

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,6 +29,7 @@
           "button",
           "card",
           "fab",
+          "line-ripple",
           "ripple",
           "top-app-bar",
           "infrastructure"
--- a/package.json
+++ b/package.json
@@ -29,6 +29,7 @@
           "button",
           "card",
           "fab",
+          "floating-label",
           "line-ripple",
           "ripple",
           "top-app-bar",
--- a/packages/button/index.scss
+++ b/packages/button/index.scss
@@ -1,1 +1,1 @@
-@import "@material/button/mdc-button.scss";
+@import "@material/button/mdc-button";
--- a/packages/card/index.scss
+++ b/packages/card/index.scss
@@ -1,1 +1,1 @@
-@import "@material/card/mdc-card.scss";
+@import "@material/card/mdc-card";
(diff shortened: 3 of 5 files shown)


johnelliott:
I signed it!

moog16:
@johnelliott Thanks for opening the PR. You're right it's not necessary and we can remove the extension from the filename. Can you also remove it from card, floating-label, line-ripple, and text-field Sass files for consistency?

Also something I realize is that you do not have the creds for the screenshot tests. I pushed a commit to see if it works for me.

johnelliott:
@moog16 thanks Matt. I like this lib; want to show my support :)

johnelliott:
Is it OK to add line-ripple, floating-label to the allowed scopes for commit messages?

johnelliott:
Perhaps this [test failure](https://github.com/material-components/material-components-web-react/pull/62l) is also permission related for the screenshot job.

moog16:
RE allowed scopes: YES definitely

RE Testing: Looking into Docker. Might be able to help us here.

johnelliott:
Ok, I use Docker and maybe that'd be the easiest way to get many people up and running those tests. I  would be OK helping to set that up but I'd need initial direction.

moog16:
Opened this PR: https://github.com/material-components/material-components-web-react/issues/71. I'm just gonna run the tests locally and merge this for you. Thanks!

johnelliott:
Thanks @moog16

PR #68, opened 2018-05-24
feat(top-app-bar): Add top app bar fixed variant

### LINKED ISSUE #54: Add Top App Bar's new Fixed Variant
After upgrading MDC Web to v0.35.1, the fixed variant must be added to the React Component.

moog16:
fixes #54
Related to changes from v0.35.x web sync


packages/top-app-bar/index.js
```diff
@@ -67,6 +77,17 @@ export default class TopAppBar extends React.Component {
...
 
+  setStyle = (varName, value) => {
+    const updatedStyle = Object.assign({}, this.state.style);
+    updatedStyle[varName] = value;
+    this.setState({style: updatedStyle});
+  }
+
+  getMergedStyles = () => {
+    const {style} = this.state;
+    return Object.assign({}, style, this.props.style);
```
bonniezhou: Is there a reason we don't want to add `props.style` to the `state`? What if we add `props.style` to the state in the constructor, and just return `state.style` here?

.travis.yml
```diff
@@ -31,5 +31,5 @@ matrix:
       install: npm i
       script:
         - ./test/screenshot/start.sh
-        - sleep 190s
+        - sleep 200s
```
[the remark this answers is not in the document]
moog16: Every time a new screenshot test is added I need to increase the time. Its a guess and check time increase, but its been averaging about 10 - 20 ms per new test.

test/screenshot/top-app-bar/prominent.js
```diff
@@ -8,7 +8,7 @@ import './index.scss';
 import MaterialIcon from '../../../packages/material-icon';
 
 ReactDOM.render((
-  <div>
+  <div className='top-app-bar-container'>
```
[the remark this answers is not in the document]
moog16: oh ya - I would like to get Stackblitz or something like it working here. Created this card https://github.com/material-components/material-components-web-react/issues/84

packages/top-app-bar/index.js
```diff
@@ -67,6 +77,17 @@ export default class TopAppBar extends React.Component {
     return React.cloneElement(element, {hasRipple: true});
   }
 
+  setStyle = (varName, value) => {
```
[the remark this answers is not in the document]
moog16: I thought it cleaned up the adapter, but no it won't be used anywhere else. Will move it

packages/top-app-bar/index.js
moog16: I thought this was the simplest approach. We update this.state.style in the setStyle adapter method. So every time we call setStyle, we would need to merge this.props.style. Also if this.props.style every updates, we would need to merge it again with this.state.style.

I think doing it in this method is the most straightforward.

bonniezhou: Oh I see, this makes sense if we don't want to have to keep track of whether or not props.style was merged into the state.

test/unit/top-app-bar/index.test.js
```diff
@@ -162,6 +173,34 @@ test('#adapter.getTotalActionItems returns false with no actionItem ' +
...
+
+test('#adapter.getTopAppBarHeight should return clientHeight', () => {
+  const div = document.createElement('div');
+  // needs to be attached to real DOM to get width
+  // https://github.com/airbnb/enzyme/issues/1525
+  document.body.append(div);
+  const options = {attachTo: div};
+  const wrapper = mount(<TopAppBar title='Hi' />, options);
+  const topAppBarHeight = wrapper.instance().adapter.getTopAppBarHeight();
+  assert.equal(topAppBarHeight, 18);
```
bonniezhou: Nit: Add a comment about where 18 comes from

PR #72, opened 2018-05-25
fix(ripple): Trigger on key events

### LINKED ISSUE #41: Ripple doesn't trigger on key down event
Steps to reproduce:

1. Start screenshot tests
1. Go to button screenshots
1. Bring to focus a button using the keyboard
1. While focused on a button, press space bar or enter key

Actual Results:
No Ripple is activated

Expected Results:
Ripple should activate, similar to MDC Web

bonniezhou:
Fixes #41.

Re-implement the `isSurfaceActive()` adapter method to check for active state directly on the ripple surface element, instead of creating and updating an `isActivated` state variable. 
Wire up the appropriate event handlers to call `deactivate(e)` in the foundation.

### DIFF
--- a/packages/ripple/index.js
+++ b/packages/ripple/index.js
@@ -34,10 +34,14 @@ const withRipple = (WrappedComponent) => {
     }
 
     createAdapter_ = (instance) => {
+      const MATCHES = [
+        'webkitMatchesSelector', 'msMatchesSelector', 'matches'
+      ].filter((p) => p in HTMLElement.prototype).pop();
+
       return {
         browserSupportsCssVars: () => util.supportsCssVariables(window),
         isUnbounded: () => this.props.unbounded,
-        isSurfaceActive: () => this.state.isActivated,
+        isSurfaceActive: () => instance[MATCHES](':active'),
         isSurfaceDisabled: () => this.props.disabled,
         addClass: (className) =>
           this.setState({classList: this.state.classList.add(className)}),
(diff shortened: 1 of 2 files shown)


packages/ripple/index.js
```diff
@@ -34,10 +34,14 @@ const withRipple = (WrappedComponent) => {
     }
 
     createAdapter_ = (instance) => {
+      const MATCHES = [
```
moog16: can we import the ./utils method instead of copy/pasting here

test/unit/ripple/index.test.js
```diff
@@ -85,6 +93,24 @@ test('touchStart event triggers activateRipple with no onTouchStart handler', ()
...
+  const touchEndHandler = td.func();
+  const wrapper = mount(<DivRipple onTouchEnd={touchEndHandler}/>);
+  const foundation = wrapper.instance().foundation_;
+  foundation.deactivate = td.func();
+  wrapper.simulate('touchEnd');
+  td.verify(foundation.deactivate(td.matchers.isA(Object)), {times: 1});
+  td.verify(touchEndHandler(td.matchers.isA(Object)), {times: 1});
+});
+
+test('touchEnd event triggers deactivateRipple with no onTouchEnd handler', () => {
```
moog16: does this test do the same thing as line 102? I think remove line 102 & line 100

test/unit/ripple/index.test.js
```diff
@@ -109,27 +135,32 @@ test('keyDown event triggers activateRipple with no onKeyDown handler', () => {
...
+  const wrapper = mount(<DivRipple onKeyUp={keyUpHandler}/>);
+  const foundation = wrapper.instance().foundation_;
+  foundation.deactivate = td.func();
+  wrapper.simulate('keyUp');
+  td.verify(foundation.deactivate(td.matchers.isA(Object)), {times: 1});
+  td.verify(keyUpHandler(td.matchers.isA(Object)), {times: 1});
 });
 
-test('#adapter.isUnbounded returns false prop is not set', () => {
+test('keyUp event triggers deactivateRipple with no onKeyUp handler', () => {
```
moog16: same with this test, you probably don't need to test it in the above test

kfranqueiro:
The enter key typically does not cause `:active` to be applied. Only space does. Test it with a native button: https://codepen.io/kfranqueiro/pen/KRjGZO

This should ostensibly already be handled by the MDC Web ripple foundation, so I'm not sure why it needs to be reimplmented in React... https://github.com/material-components/material-components-web/blob/v0.35.2/packages/mdc-ripple/foundation.js#L340

moog16:
Gotcha...I thought space and enter keys were supposed to do the same with buttons, etc. 

RE Events: It is handled by the foundation, but we chose to not use the native html events, but instead use React synthetic events. The result of this is having to setup those event in the React Component.

test/unit/ripple/index.test.js
bonniezhou: These are two different tests, one to check that the ripple deactivates when there's no handler and one for when there is. The `DivRipple` that's mounted in each of these tests are different.

PR #76, opened 2018-05-27
Render label even if the input is full width

patrick91:


patrick91:
It probably needs some tweaks still, as it is missing the padding:

patrick91:
My bad, I didn't notice the full width version doesn't have labels even in the web version, sorry for the noise!

Issue #58, opened 2018-05-21
Stroked card rename to outlined


moog16:
All references to stroked were renamed to outlined. Rename and fix all screenshot tests regarding card.

PR #79, opened 2018-05-29
refactor(card): renamed stroked card to outlined

moog16:
fixes https://github.com/material-components/material-components-web-react/issues/58


packages/card/README.md
```diff
@@ -51,7 +51,7 @@ import Card from '@material/react-card';
 #### Props
 Prop Name | Type | Description
 --- | --- | ---
-stroked | n/a | Enables stroked outline of card.
+outlined | n/a | Enables outlined outline of card.
```
bonniezhou: nit: "Enables an outline on the card"

Issue #66, opened 2018-05-23
Add autoprefixer to build pipeline


moog16:
MDC Web uses autoprefixer to build Css files from Sass files. MDC React should also do this to support devs using Css.

Issue #67, opened 2018-05-24
Add Getting Started Readme


moog16:
Should provide a start to finish on how to build and app using our components. Similar to https://github.com/material-components/material-components-web/blob/master/docs/getting-started.md.

Instead ours will start with `create-react-app` and then move into building a button on the page.

jimbojetlag:
This is great and much needed!

create-react-app 1.x does not support sass. Are you planning to base the guide on CRA 2.0 which is in alpha?

If it's going to be by importing css, and not sass, in CRA 1.x, how can customizing theme happen?

PR #81, opened 2018-05-29
infrastructure: add autoprefixer to webpack build step

moog16:
fixes https://github.com/material-components/material-components-web-react/issues/66


docs/guidelines.md
```diff
@@ -54,3 +54,7 @@ class MyApp extends React.Component {
   }
 }
 ```
+
+## Building Sass Files
+
+If you use the Sass files instead of the built CSS files found in the `/dist` directory, we recommend you use a package like [Autoprefixer](https://github.com/postcss/autoprefixer). Autoprefixer adds vendor prefixes to your CSS Output (ie. -moz-xyz, -webkit-xyz, etc). We use it in our [webpack config](../packages/webpack.config.js) as well in our [MDC-Web repository](https://github.com/material-components/material-components-web).
```
williamernest: This sounds overly descriptive of Autoprefixer. Maybe 
```
If using Sass instead of the pre-compiled CSS, we recommend you use [Autoprefixer](https://github.com/postcss/autoprefixer) for improved browser support. 
```
or something like that? 

Also, won't they need instructions for setting up `sass-loader` in order to compile?

moog16: Ya good call...I'm going to cover sass loader in the getting-started.md as part of issue #67.

Issue #82, opened 2018-05-30
bug: add error handling for npm run build


moog16:
When running `npm run build`, and there is an empty directory in `/packages`, throw an error to indicate that a JS file is required.

PR #85, opened 2018-05-31
fix(infrastructure): fixed npm run build error for empty package directories

moog16:
fixes https://github.com/material-components/material-components-web-react/issues/82

Error occurs when an empty directory exists in /packages ie. Chips.

### DIFF
--- a/packages/webpack.config.js
+++ b/packages/webpack.config.js
@@ -20,10 +20,12 @@ const path = require('path');
 const ExtractTextPlugin = require('extract-text-webpack-plugin');
 
 const isDirectory = (source) => lstatSync(source).isDirectory();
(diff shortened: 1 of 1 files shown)


packages/webpack.config.js
```diff
@@ -20,10 +20,12 @@ const path = require('path');
 const ExtractTextPlugin = require('extract-text-webpack-plugin');
 
 const isDirectory = (source) => lstatSync(source).isDirectory();
+const containsJsFile = (source) => readdirSync(source).some(file => path.parse(file).ext === '.js');
```
kfranqueiro: I think you can simplify `path.parse(file).ext` to `path.extname(file)`?

PR #87, opened 2018-05-31
chore(infrastructure): removed deprecated componentWillReceiveProps

### LINKED ISSUE #65: Migrate from deprecated lifecycle componentWillReceiveProps
Some of the packages use deprecated react lifecycle method `componentWillReceiveProps` (Please see [deprecation note](https://reactjs.org/docs/react-component.html#unsafe_componentwillreceiveprops)).

It should use `getDerivedStateFromProps` instead.

moog16:
As of React 16.x, the lifecycle method componentWillReceiveProps is being deprecated. React now suggests to replace the method with getDerivedStateFromProps (a static method).

I renamed/moved the logic from componentWillReceiveProps to componentDidUpdate, since all the logic is calling foundation methods not updating this.state (the main function of getDerivedStateFromProps).

https://reactjs.org/docs/react-component.html#unsafe_componentwillreceiveprops

Fixes #65


moog16:
Surprisingly no...I moved everything out of the componentWillReceiveProps
--> componentDidUpdate. In our tests we never explicitly call those
lifecycle methods, and they both get called just at different times.

bonniezhou:
Ohh I see, then I think a lot of tests can be renamed to something that doesn't refer to `#componentWillReceiveProps` or any lifecycle methods in general since that seems to be an implementation detail.

moog16:
ah...ya good call. I'll make those updates

moog16:
Ya it is a detail, but I think its nice just to know what part of the code is being tested and its location in the file. Gonna merge!

PR #97, opened 2018-06-09
docs: build-css script, edit for output directory.

jsuraj:
modified:   docs/getting-started.md
The -o option of node-sass-chokidar is for output directory. For import './App.css' to work in App.js, in build-css script it should be -o ./src otherwise a directory named App.css is created an App.css file is created inside it.

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -54,7 +54,7 @@ with:
 
 ```js
 "scripts": {
-  "build-css": "node-sass-chokidar --include-path ./src --include-path ./node_modules ./src/App.scss -o ./src/App.css",
+  "build-css": "node-sass-chokidar --include-path ./src --include-path ./node_modules ./src/App.scss -o ./src",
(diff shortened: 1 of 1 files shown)


moog16:
@jsuraj can you explain the issue you were having?

jsuraj:
@moog16 In Getting Started guide the script to build css is
```
"build-css": "node-sass-chokidar --include-path ./src --include-path ./node_modules ./src/App.scss -o ./src/App.css"
```
This creates App.css inside a directory named App.css. This happens because the -o option of node-sass-chokidar is to specify output directory (https://github.com/michaelwayman/node-sass-chokidar#options).

Whereas, in **Step 3: Use MDC React Button**: (Getting started guide) we are importing css file as ```import './App.css';```.
So, for this import statement to work the css file should be created in src directory itself. 
Hence the build script should be 
```
"build-css": "node-sass-chokidar --include-path ./src --include-path ./node_modules ./src/App.scss -o ./src"
```

PR #99, opened 2018-06-10
fix(line-ripple): Convert style name to camel case

lucasecdb:
fixes #86

PR #106, opened 2018-06-19
fix(line-ripple): Convert style name to camel case

moog16:
related https://github.com/material-components/material-components-web-react/pull/99


moog16:
Closing for https://github.com/material-components/material-components-web-react/pull/99

Issue #100, opened 2018-06-11
Why '.../dist/' ?


ebeloded:
Dear maintainers,

I find it a bit off-putting that to use a component one has to import the compiled version from `@material/react-button/dist` (as opposed to simply `@material/react-button`). Is there a reason why you decided to make it so?

Apart from the packages in this repo I have never had to reference `/dist` - it is not obvious for me as a component consumer. What's more, in the docs to every component, the imports in demo look as one would expect :`import Button from '@material/react-button'`. Yet this won't work without having to transpile imported modules or in create-react-app context.

Please, let me if there is a good reason behind this decision I may not understand.

ben-mckernan:
Have you considered the impact using the compiled version will have on your bundle size? Using the uncompiled version allows for tree shaking which will lead to smaller bundle sizes.

moog16:
@ebeloded sorry for the late reply. We thought that the uncompiled components would be used more than the compiled components. Like @ben-mckernan referenced, we saw it as a bundle savings. We also thought that having the transpiled code in a `/dist` or `/build` directory as being more semantic. However it seems that our main target audience are developers using CRA, which was short-sighted on our part.

After installing a few other packages and seeing what is distributed from the entry point it seems like we have this backwards, and in fact `import Button from '@material/react-button` should have the compiled code. And as reference in #107 we are causing more issues.

PR #108, opened 2018-06-20
feat(infrastructure): default npm entry point to ES5

### LINKED ISSUE #107: Please don’t publish experimental syntax on npm
Hi, I help maintain React and some other projects like Create React App. 👋

I recently saw in this issue (https://github.com/facebook/create-react-app/issues/4648#issuecomment-398540069) that this project uses experimental syntax (such as class properties) as part of the npm entry point.

I strongly encourage you to reconsider this decision. If the class properties transform ends up changing (which is very likely) or abandoned (not as likely, but could happen), having uncompiled packages on npm using it will create a huge amount of churn for everyone (including maintainers of this project). Not to mention this makes the project unusable in any environment that respects the spec (such as Node.js).

Please compile any experimental syntax away before publishing. If you’re convinced that keeping `import` uncompiled is worth it (which is a whole separate can of worms as you can see in https://github.com/graphql/graphql-js/issues/1248), please **compile at least the non-standard syntax away when publishing**. That includes JSX too. 

I understand this might seem like an inconvenience now. But it will be much better to do now than deal with pain for months when build tools and the spec changes. Thanks.

moog16:
fixes #107 #100

### DIFF
--- a/packages/button/package.json
+++ b/packages/button/package.json
@@ -3,6 +3,7 @@
   "version": "0.2.0",
   "description": "Material Components React Button",
   "license": "Apache-2.0",
+  "main": "dist/index.js",
   "keywords": [
     "mdc web react",
     "material components react",
--- a/package-lock.json
+++ b/package-lock.json
@@ -5728,7 +5728,6 @@
           "version": "1.0.5",
           "bundled": true,
           "dev": true,
-          "optional": true,
           "requires": {
             "delayed-stream": "~1.0.0"
           }
@@ -5791,8 +5790,7 @@
         "delayed-stream": {
           "version": "1.0.0",
           "bundled": true,
-          "dev": true,
-          "optional": true
+          "dev": true
         },
         "delegates": {
           "version": "1.0.0",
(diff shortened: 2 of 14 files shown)


packages/button/package.json
```diff
@@ -3,6 +3,7 @@
   "version": "0.2.0",
   "description": "Material Components React Button",
   "license": "Apache-2.0",
+  "main": "dist/index.js",
```
hzoo: Probably better as another pr: if you still want an esm build like Dan mentioned, you could run Babel with `["env", { "modules": false" }]` into a new dist type folder and add the `module` field like with main so it just doesn't touch the import/exports.

moog16: @hzoo I just pushed some changes that doesn't touch the import/exports by adding the {modules: false} options. I think its working correctly as I'm now seeing lines like this:

```js
/* harmony import */ var __WEBPACK_IMPORTED_MODULE_0_react__ = __webpack_require__(0);```

moog16:
Discussed offline: renamed <package_name>.map.js --> index.map.js and created .es.js files as part of dist directory.

PR #112, opened 2018-06-22
fix(ripple): setState warning when unmounting the component

### LINKED ISSUE #111: RippledComponent warning: Can't call setState (or forceUpdate) on an unmounted component
When Button is unmounted, React will produce warning as bellow:

```
Warning: Can't call setState (or forceUpdate) on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in the componentWillUnmount method.
    in RippledComponent (created by App)
```

Here is minimal test case that would produce warning when button is clicked:
https://codesandbox.io/s/9zw5v67kn4

lucasecdb:
fixes #111.

I don't really like this solution, but this is what I could do. I looked into the source for the foundation of the ripple and the problem is that it's calling `removeClass` and `updateCssVariable` from the `destroy` method, which I don't really know why.

Another better alternative would require submitting a pr to `material-components-web` and changing the implementation of the `destroy`, but this fixes it in the meantime.


kfranqueiro:
I don't have the most comprehensive understanding of React's lifecycle, so take this with a grain of salt... but it seems weird to me that the warning is issued within `componentWillUnmount` - ostensibly the component isn't actually unmounted yet, and this should be the right place to perform cleanup, right?

The console warning seems dubious in this case - it's not a memory leak, it's us performing cleanup intended _before_ unmounting. I wonder if there's a deeper issue or misunderstanding of the lifecycle.

lucasecdb:
Yeah, `componentWillUnmount` should be only used to do clean up and remove listeners and the like, to avoid updating the component unnecessarily.

The problem is because on this specific lifecycle method, you are not supposed to call `setState` (as you can see [here](https://reactjs.org/docs/react-component.html#componentwillunmount) on the docs) and, because the foundation removes the classes, which translates to us updating the state, we receive the console warning.

The thing is that it doesn't really make sense to update the state when the component is being unmounted because it's being removed from the DOM in the next render.

bmihelac:
I haven't tested the code in PR, but it looks right thing to do and what @lucasecdb stated was what I suspected is going on - foundation class want to update styles and React component is unmounted at the time.

moog16:
@lucasecdb I did some investigating and read this: https://reactjs.org/blog/2015/12/16/ismounted-antipattern.html. Seems like your pattern is correct.

Also after my learnings I think that this is not a system wide issue. This is just happening in ripple. We will need this pattern in future components if they call setState from `destroy()`, but right now ripple is the only one.

This warning makes sense as well. If the component is unmounting, then logically you would not want the component to update. Hence calling setState is a waste at this point. In our case, removing css variables and classes are useless if removed out of the DOM.

test/unit/ripple/index.test.js
```diff
@@ -272,3 +272,12 @@ test('throws error if no foundation', () => {
   const DivNoRefRipple = withRipple(DivNoRef);
   assert.throws(DivNoRefRipple.prototype.componentDidMount);
 });
+
+test('doesn\'t call #Component.setState after unmount', () => {
```
moog16: If I comment out 
```js
// this.isMounted_ = false;
```
in the componentWillUnmount method, the tests still passes...so effectively this is not testing anything. Looking at a fix.

moog16: I'm convinced there is no way to write a test for this with enzyme. Once you call unmount, you cannot get `wrapper.instance()` to check whether its true or false. I added tests to make sure that addClass/removeClass/updateCssVars methods do not update classLists/cssVars if `isMounted_` is false.

moog16:
@lucasecdb please sign this PR if you're ok with it: https://github.com/material-components/material-components-web-react/pull/134

lucasecdb:
@googlebot I give my consent

moog16:
Thanks! Once #134 passes, we can merge this one and close #134

moog16:
#134 passed CI tests. Merged!

bmihelac:
Thanks for elaboration and solving this issue.

PR #117, opened 2018-06-25
feat(chips): Add new component

bonniezhou:
Add basic working component for Chips, including Chip and ChipSet. Does not include any variants.


packages/chips/chip-set/index.js
```diff
@@ -0,0 +1,69 @@
...
+  renderInputChip(chip) {
+    return (
+      <Chip key={chip.id}>{chip.label}</Chip>
+    );
+  }
+
+  render() {
+    return (
+      <div className={this.classes}>
+        {this.state.chips.map(this.renderInputChip.bind(this))}
```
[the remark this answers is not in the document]
bonniezhou: Oops, copied this over from an experimental branch. Later when we add input chips `this` will be used in the render method so we can bind it then.

package.json
```diff
@@ -45,6 +45,7 @@
     "@google-cloud/storage": "^1.6.0",
     "@material/button": "^0.36.0",
     "@material/card": "^0.36.0",
+    "@material/chips": "^0.36.0",
```
moog16: does chips have a version 0.36.0? It is 0.35.x in package-lock

packages/chips/README.md
```diff
@@ -0,0 +1,46 @@
...
+```
+
+## Props
+
+### Chip Set
+
+Prop Name | Type | Description
+--- | --- | ---
+className | String | Classes to be applied to the chip set element.
+labels | Array | Array of text to be displayed in each chip.
```
amsheehan: Language in the description is a little ambiguous. Maybe "An array of strings. Each string has a corresponding chip whose label will be set to the value of that string."

packages/chips/chip-set/index.js
```diff
@@ -0,0 +1,67 @@
+import React, {Component} from 'react';
+import classnames from 'classnames';
+import PropTypes from 'prop-types';
+import {MDCChipSetFoundation} from '@material/chips';
+
+import Chip from '../chip';
+
+export default class ChipSet extends Component {
```
amsheehan: In the `README`'s example, you use a different import style for `Component`, e.g.: 

```
import React from 'react';

class Foo extends React.Component {...}
```

Suggest changing the example in the `README` to use this code style instead.

packages/chips/chip-set/index.js
```diff
@@ -0,0 +1,67 @@
...
+
+  foundation_ = null;
+
+  constructor(props) {
+    super(props);
+    this.maxId = 0;
+    this.state = {
+      chips: props.labels.map((label) => {
```
amsheehan: It is an anti-pattern to set `props` that will never change to `state`. Consider using props directly in the `render` method.

packages/chips/chip-set/index.js
```diff
@@ -0,0 +1,67 @@
...
+
+import Chip from '../chip';
+
+export default class ChipSet extends Component {
+
+  foundation_ = null;
+
+  constructor(props) {
+    super(props);
+    this.maxId = 0;
```
amsheehan: What is this used for?

packages/chips/chip-set/index.js
```diff
@@ -0,0 +1,67 @@
...
+    this.state = {
+      chips: props.labels.map((label) => {
+        return {
+          label,
+          id: this.maxId++,
```
amsheehan: Same question as above. Where are we using this? Is it just to assign a `key` to the iterator? If so we can pass a second argument to `map` and use that. E.g.: in `render`, we could do something like
```
{this.props.labels.map(label, index) => (
  <div key={index}>
    {label}
  </div>
)}`
```

Further, if chip labels are unique, we can just use those.

packages/chips/chip-set/index.js
```diff
@@ -0,0 +1,67 @@
...
+  }
+}
+
+ChipSet.propTypes = {
+  className: PropTypes.string,
+  labels: PropTypes.array.isRequired,
+};
+
+ChipSet.defaultProps = {
+  className: '',
```
amsheehan: Is this necessary?

packages/chips/chip/index.js
```diff
@@ -0,0 +1,78 @@
+import React, {Component} from 'react';
+import classnames from 'classnames';
+import PropTypes from 'prop-types';
+import withRipple from '@material/react-ripple';
+import {MDCChipFoundation} from '@material/chips';
+
+export class Chip extends Component {
+
```
amsheehan: Did eslint complain about starting this with an empty line?

packages/chips/chip/index.js
```diff
@@ -0,0 +1,78 @@
...
+}
+
+Chip.propTypes = {
+  className: PropTypes.string,
+  initRipple: PropTypes.func,
+  unbounded: PropTypes.bool,
+  children: PropTypes.string,
+};
+
+Chip.defaultProps = {
```
amsheehan: I'm not sure you'll need all of these. Possibly initRipple, because I don't know the internals of it.

packages/chips/package.json
```diff
@@ -0,0 +1,29 @@
...
+    "materialchips"
+  ],
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/material-components/material-components-web-react.git"
+  },
+  "dependencies": {
+    "@material/chips": "^0.36.0",
+    "@material/react-ripple": "^0.2.0",
+    "classnames": "^2.2.5",
```
amsheehan: Do we have a way of making sure that minor version changes to these will not cause problems for the components?

test/unit/chips/Chip.test.js
```diff
@@ -0,0 +1,31 @@
...
+});
+
+test('renders chip', () => {
+  const wrapper = mount(<Chip>Hello world</Chip>);
+  assert.exists(wrapper.find('.mdc-chip'));
+});
+
+test('#adapter.addClass adds class to state.classList', () => {
+  const wrapper = mount(<Chip>Jane Doe</Chip>);
+  wrapper.instance().foundation_.adapter_.addClass('test-class-name');
```
amsheehan: Should we be accessing private properties in a test? Is there a way to spy on `addClass`? I'm seeing this pattern here, in the next test, and one of the `ChipSet` tests.

packages/chips/chip-set/index.js
bonniezhou: Thanks for taking a look, Alex! I set things up this way to anticipate adding/removing input chips. Originally, I thought the ChipSet could internally manage updating the list of chip labels to be rendered, including assigning IDs. After talking to @moog16 it sounds more reasonable for the end developer to manage the list of chip labels, and for ChipSet to just handle rendering whatever's passed into it. Made a bunch of changes, PTAL.

test/unit/chips/Chip.test.js
bonniezhou: Changed the `ChipSet` test to use its public adapter getter, but I'm not sure if there's a way to do the same for `Chip` since it's an HOC wrapped in `withRipple`.

amsheehan: Are you using Jest to run the tests? If so, there are some handy automatic mocking functions you can use and pass arbitrary props down to Chip and check its `className` property.

http://jestjs.io/docs/en/es6-class-mocks#automatic-mock

bonniezhou: We're not using Jest :( we're using Karma like in the MDC Web repo

packages/chips/chip-set/index.js
```diff
@@ -0,0 +1,51 @@
...
+      <Chip key={index}>
+        {label}
+      </Chip>
+    );
+  }
+
+  render() {
+    return (
+      <div className={this.classes}>
+        {this.props.labels.map((label, index) => this.renderChip(label, index))}
```
moog16: this can reduce to: `this.props.labels.map(this.renderChip)`

packages/chips/README.md
```diff
@@ -0,0 +1,46 @@
+# React Chips
+
+A React version of an [MDC Chips](https://github.com/material-components/material-components-web/tree/master/packages/mdc-chips).
+
+## Installation
+
+```
+npm install @material/react-chips
```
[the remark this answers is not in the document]
kfranqueiro: No, `--save` has not existed for several major versions of npm.

PR #119, opened 2018-06-28
Updates readme for top-app-bar to include the css needed for the components

diondree:
Updating the Readme with accurate usage of the top app bar component.


diondree:
I've signed it!

moog16:
Thanks for opening this PR. I'm not sure that this belongs here. This is a generic import statement that would go into a higher level doc. I think something like that would belong in the [getting started readme](https://github.com/material-components/material-components-web-react/blob/master/docs/getting-started.md). 

What do you think?

diondree:
Hey, very happy to do so. I agree that this suited for the getting started readme but as a user of this project, I feel as though it's up to the developers to guess where the css for the particular component. The documentation for each component should point to where the css can be found.

moog16:
@diondree I pushed some changes to be more aligned with MDC-Web readmes. What do you think about that instead? If so, I think all packages' readmes should be update. Would you be willing to update them? 

Created this https://github.com/material-components/material-components-web-react/issues/123 to keep track of this.

diondree:
Oh, I see now. Yes, this would be much better and yes for sure I'd be happy to update the readmes. Can I get a list or is it all but the MDC-Button?

moog16:
@diondree its all directories in the `/packages` folder. 

Also your most recent changes include formatting a table. I find doing that increases maintenance by quite a bit. Can you please undo that change?

diondree:
@moog16 Alright not a problem and yea i just noticed the change after I pushed, just undid it. Sorry about that.

diondree:
Oh that's great! Thank you! If there's anything else I can help out with, please feel free to let me know.

PR #121, opened 2018-07-02
feat(chips): Add selection to chips

### LINKED ISSUE #130: Add choice/filter chips

bonniezhou:
Add support for single selection (choice chips) and multi-selection (filter chips). Fixes #130.

Due to the unidirectional flow in React, the user will manage the selection state of the chips. I included an example of how a user might implement choice chips and filter chips in the [screenshot tests](https://github.com/material-components/material-components-web-react/blob/ed5c0a35a89f9d9feb4c8f16ed38a6486692f2b8/test/screenshot/chips/index.js).

This means that `MDCChipSetFoundation`, whose main job is to manage selection state, is not used at all in the MDC React `ChipSet` 😱😱😱!!! This also means that the `mdc-chip-set--choice` and `mdc-chip-set--filter` classes serve no purpose, and are therefore omitted from `ChipSet`. We may want to rethink how `MDCChipSetFoundation` is implemented in MDC Web.

`ChipSet` is in charge of overriding ripple adapter methods for the `Chip` and rendering the `ChipCheckmark` for filter chips.

This PR does not include support for leading/trailing icons yet.


packages/chips/Chip.js
```diff
@@ -34,13 +36,21 @@ export class Chip extends Component {
         classList.delete(className);
```
mario-jerkovic: I don't think direct state mutation is the right solution. What about creating `new Set()`

moog16: there could be other classes that we don't want to remove, which is what would happen if we created another Set.

bonniezhou: I think @mario-jerkovic means something like this:
```
const classList = new Set(this.state.classList);
```
@moog16 We should probably do this for all the other components too to avoid mutating state.

moog16: Gotcha so we'd do:
```
const classList = new Set(this.state.classList);
classList.delete(className);
this.setState({classList}));
```

moog16: This is very immutablejs-ish, are you saying this because React doesn't do a deep comparison of the Set's values?

mario-jerkovic: I was thinking about two potential problems with this approach. First is that this.setState is asynchronous and if you try to grab the value of this.state.classList it will be mutated before state update, and second is that it would be a pain to implement shouldComponentUpdate with mutable state.

moog16: That makes sense. Potentially we could call removeClass/addClass more than once, and the classList might not be updated. I realize that it could update with incorrect classes. The code I wrote above also has this issue.

packages/chips/README.md
```diff
@@ -43,14 +43,19 @@ class MyApp extends Component {
...
+filter | Boolean | Indicates that the chips in the set are filter chips, which allow multiple selection from a set of options
+
 
 ### Chip
 
 Prop Name | Type | Description
 --- | --- | ---
-className | String | Classes to be applied to the chip element.
+className | String | Classes to be applied to the chip element
+id | Number | Unique identifier for the chip
```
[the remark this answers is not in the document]
bonniezhou: This would cause issues in the case of input chips because chips would re-render as indices are reassigned.

For example, we have Chip A with index 0 and Chip B with index 1. As a result of removing Chip A, Chip B's index is reassigned to 0. This would cause Chip B to re-render, which would trigger the entry animation, even though nothing has happened to Chip B.

packages/chips/Chip.js
mario-jerkovic: Yes, that's right. And like you said, because React. js uses shallow comparison, if you just mute the state with a `delete` call on the state the component won't be re-rendered. I have made a simple example with `add/remove`
https://codesandbox.io/s/9y9pp9l7vy . 

At first glance I can tell you that you have to make it like this:
`
const classList = new Set([...this.state.classList]);
`

moog16: Wow thanks for creating that codesandbox, really helpful. I like it, but question about this syntax:

```
this.setState(() => ({classList:newclassList}))
```

Why the function?

mario-jerkovic: No problem :) There is nothing wrong with just passing an object inside setState this is just another way of setting state and its more expressive. At the first glance you can see that the setState is asynchronous because of the callback and one added benefit is that function accepts the previous state and current props of the component so you can recalculate new state from the previous state.

moog16: gotcha - @bonniezhou want to make this change? I will make an issue to update the other components.

bonniezhou: Done!

packages/chips/Chip.js
```diff
@@ -21,26 +21,39 @@ export class Chip extends Component {
...
+        const classList = new Set(this.state.classList);
         classList.delete(className);
         this.setState({classList});
       },
+      hasClass: (className) => this.classes.split(' ').includes(className),
     };
   }
 
+  handleInteraction = () => {
+    this.props.handleSelect(this.props.id);
```
moog16: Its possible that the users of this chips component will want to do something onClick. Also I think changing from `handleInteraction` --> `handleClick` is more clear. Try:

```
handleClick = (e) => {
  if (typeof this.props.onClick === 'function) {
    this.props.onClick(e);
  }
  this.props.handleSelect(this.props.id);
}
```

packages/chips/Chip.js
```diff
@@ -21,26 +21,39 @@ export class Chip extends Component {
...
+  }
+
   render() {
     const {
       className, // eslint-disable-line no-unused-vars
-      children,
+      label,
+      handleSelect, // eslint-disable-line no-unused-vars
+      chipCheckmark,
+      computeBoundingRect, // eslint-disable-line no-unused-vars
```
moog16: Also include this.props.onClick here, so the user cannot override our method.

packages/chips/ChipCheckmark.js
```diff
@@ -0,0 +1,25 @@
+import React, {Component} from 'react';
+
+export default class ChipCheckmark extends Component {
+  width = null;
+
+  init = (element) => {
+    // The checkmark's width may initially be set to 0, so use the checkmark's height as a proxy since the
+    // checkmark should always be square.
+    this.width = element.getBoundingClientRect().height;
```
moog16: have you tried unmounting this component? the `ref` is called with `null` on unmount. So it might throw a null error here.

packages/chips/README.md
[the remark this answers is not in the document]
moog16: Looking forward to entry chips, if we wanted to create a new chip, what guidance will we provide? How will the id be generated?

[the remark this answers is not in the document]
bonniezhou: Still need to experiment with this more, but we'll likely provide guidance for the user write most of the logic for adding a input chip. The `Chip` will handle all animations. The user will generate the id's since they'll be managing selected chips using their id's.

PR #134, opened 2018-07-10
fix(ripple): setState warning when unmounting the component TESTS

moog16:
related https://github.com/material-components/material-components-web-react/pull/112


moog16:
Closing for #112

Issue #114, opened 2018-06-25
The README is missing a section on how to install material-components-web-react


mugwump:
the [README](https://github.com/material-components/material-components-web-react/blob/master/README.md) is pretty elaborate on how to install react and sass - but its missing the crucial part of how to actually install ```material-components-web-react``` - the simple ```npm i --save material-components-web-react``` only brings up an outdated version (0.0.10), which is missing most of the components except button and checkbox.

moog16:
@mugwump this repo is not the same as this repo: https://github.com/MihirNS/material-components-web-react. Our npm url is https://www.npmjs.com/package/@material/react-text-field. 

Also the instructions for each package is within each directory of the package, ie. https://github.com/material-components/material-components-web-react/blob/master/packages/text-field/README.md.

mugwump:
I figured that out after some trial&error that you had to install individual components instead of installing the whole package and then referencing components. I understand the rationale behind this, but a short note in the Readme of the main-project would probably help people to get started...

moog16:
Are you talking about the MDC-Web repo: https://github.com/material-components/material-components-web?

nickmccurdy:
Does this help? #137

mugwump:
yes, this makes it much much clearer and avoids the confusion with all the likely-sounding libs out there. +1 for the commit!

moog16:
Thanks for opening the PR!

PR #137, opened 2018-07-11
Move Getting Started to readme for discoverability

nickserv:
As discussed in [Reactiflux](https://www.reactiflux.com/), I didn't notice the Getting Started link in the readme until it was posted elsewhere. I think this would make it easier to find, and it's more consistent with MCW which has some basic getting started info in the readme.


moog16 (approved):
This change makes sense. I think many other libs have a "Quick Start" on their main README, but this is our equivalent. 

I also updated a broken link and fixed a line about including ES5 in our dist directory (not accurate anymore).

fixes #114

moog16:
Just doc changes, so I won't open another PR to run tests. Merging.

Issue #143, opened 2018-07-18
Create icon button


lynnjepsen:
See the MDC Web version we will wrap: 
https://github.com/material-components/material-components-web/tree/master/packages/mdc-icon-button

kfranqueiro:
Just FYI, I would point out https://github.com/material-components/material-components-web/issues/3107 will affect this component's DOM structure when we pursue it (and we will likely want to pursue it sometime soon to support clients that want SVG-powered toggle buttons).

Not saying it can't be wrapped before then, just that it may need an update a couple of months from now.

PR #154, opened 2018-07-18
feat(icon-button): create component

moog16:
fixes https://github.com/material-components/material-components-web-react/issues/143


packages/icon-button/README.md
```diff
@@ -0,0 +1,78 @@
...
+}
+```
+
+You can use any other icon here such as Font Awesome, which is documented more in detail [here](https://github.com/material-components/material-components-web/tree/master/packages/mdc-icon-button#icon-button-toggle-with-font-awesome). If you're using [Google Font's Material Icons](https://material.io/tools/icons/?style=baseline), we recommend using our [Material Icon Component](../material-icon) as shown in the example above.
+
+## Variants
+
+### Icon Button Toggle
+
+If you need to use this component as an Icon Button Toggle, please read [this documentation](https://github.com/material-components/material-components-web/blob/v0.37.1/packages/mdc-icon-button/README.md#icon-button-toggle). The following is an example using data attributes to toggle between on/off states:
```
lynnjepsen: Did you intend to lock this link to version 0.37.1?

packages/icon-button/README.md
```diff
@@ -0,0 +1,78 @@
...
+import MaterialIcon from '@material/react-material-icon';
+
+class MyApp extends React.Component {
+  render() {
+    return (
+      <IconButton
+        className='mdc-icon-button material-icons'
+        aria-label='Add to favorites'
+        aria-pressed='false'
+        data-demo-toggle
```
lynnjepsen: I do not think we should use DOM data attributes here. We should use React props. Is there some weird logic in MDC Web that assumes there are data attributes in the DOM?

packages/icon-button/index.js
```diff
@@ -0,0 +1,129 @@
+import React, {Component} from 'react';
+import classnames from 'classnames';
+import PropTypes from 'prop-types';
+import withRipple from '@material/react-ripple';
+import {MDCIconButtonToggleFoundation} from '@material/icon-button';
+
+const {strings} = MDCIconButtonToggleFoundation;
+
+export class IconButtonBase extends Component {
```
lynnjepsen: Why do you have IconButtonBase and IconButton? Is this because of the ripple? Why dont you have a nice clean `withRipple` call like Button and Fab have?

packages/icon-button/index.js
```diff
@@ -0,0 +1,129 @@
...
+      [strings.ARIA_PRESSED]: props[strings.ARIA_PRESSED],
+      [strings.ARIA_LABEL]: props[strings.ARIA_LABEL],
+    };
+  }
+
+  componentDidMount() {
+    this.initializeFoundation();
+  }
+
+  initializeFoundation = () => {
```
lynnjepsen: You only ever call this method from one place (line 22 above). So why have a separate method at all? Just put these two lines of code at line 22

packages/icon-button/index.js
```diff
@@ -0,0 +1,129 @@
...
+  get adapter() {
+    return {
+      addClass: (className) =>
+        this.setState({classList: this.state.classList.add(className)}),
+      removeClass: (className) => {
+        const classList = new Set(this.state.classList);
+        classList.delete(className);
+        this.setState({classList});
+      },
+      // the foundation needs data attributes that will be passed in as props if the user wants this feature
```
lynnjepsen: are these the data-toggle-on stuff I saw before?

packages/icon-button/index.js
```diff
@@ -0,0 +1,129 @@
...
+const {strings} = MDCIconButtonToggleFoundation;
+
+export class IconButtonBase extends Component {
+
+  constructor(props) {
+    super(props);
+    this.state = {
+      classList: new Set(),
+      childContent: props.children,
+      [strings.ARIA_PRESSED]: props[strings.ARIA_PRESSED],
```
lynnjepsen: As far as I can tell, you never update the ARIA_PRESSED part of this.state

So why have ARIA_PRESSED in this.state at all?

packages/icon-button/index.js
```diff
@@ -0,0 +1,129 @@
...
+      // the foundation needs data attributes that will be passed in as props if the user wants this feature
+      getAttr: (attr) => this.props[attr],
+      setAttr: (attr, value) => this.setState({[attr]: value}),
+      setText: (childContent) => this.setState({childContent}),
+    };
+  }
+
+  handleClick_ = (e) => {
+    this.props.onClick(e);
+    this.foundation_.toggle();
```
lynnjepsen: its kinda weird that MDCIconButtonFoundation doesn't have a handleClick method. Instead you are calling `toggle()`. Why doesn't MDCIconButtonFoundation follow our `handleX` pattern?

packages/icon-button/index.js
```diff
@@ -0,0 +1,129 @@
...
+        {childContent}
+      </button>
+    );
+  }
+}
+
+IconButtonBase.propTypes = {
+  children: PropTypes.node,
+  className: PropTypes.string,
+  className: PropTypes.string,
```
lynnjepsen: duplicate?

packages/icon-button/README.md
moog16: No, I think it should just be master or auto updated to whatever version it actually is. A lot of our readmes are out of sync in this section so we might want to just change them all to master to lower our maintenance.

packages/icon-button/README.md
moog16: Ya the icon button foundation is looking for 6 different data attributes. These are copied verbatim:

```
  DATA_TOGGLE_ON_LABEL: 'data-toggle-on-label',
  DATA_TOGGLE_ON_CONTENT: 'data-toggle-on-content',
  DATA_TOGGLE_ON_CLASS: 'data-toggle-on-class',
  DATA_TOGGLE_OFF_LABEL: 'data-toggle-off-label',
  DATA_TOGGLE_OFF_CONTENT: 'data-toggle-off-content',
  DATA_TOGGLE_OFF_CLASS: 'data-toggle-off-class',
```

@williamernest was telling me that there was a suggestion to toggle classes instead of toggle between these data attributes, but those changes haven't been made yet. Do you want to update icon button foundation, and then make appropriate changes here?

packages/icon-button/index.js
moog16: It is mainly because the ripple needs to be "unbounded". I'm not sure if there is a cleaner syntax than the one I have now. I'll try look for something better.

packages/icon-button/index.js
moog16: yup this is referring tot he 6 different data attrs

packages/icon-button/index.js
moog16: #adapter.setAttr method updates state. It really is supposed to only update aria-pressed and aria-label according to the foundation code.

packages/icon-button/index.js
moog16: There is a private `toggleFromEvt_` method which is the equivalent `handleX` method. In the web component, it directly calls this `toggle` method. Want to change the private `toggleFromEvt_` method to a public and rename to `handleClick`?

packages/icon-button/index.js
moog16: This is what the change looks like. There is a minor change to withRipple.

https://github.com/material-components/material-components-web-react/pull/154/commits/7496e445de33d206fdba012f4fee93b4fb7e72fc

bonniezhou: Seems like we can rename `IconButtonBase` to `IconButton` now, should we do that to keep things consistent?

packages/ripple/index.js
```diff
@@ -4,7 +4,7 @@ import classnames from 'classnames';
 
 import {MDCRippleFoundation, util} from '@material/ripple';
 
-const withRipple = (WrappedComponent) => {
+const withRipple = (WrappedComponent, isUnbounded = false) => {
```
bonniezhou: Add unit tests?

moog16:
spoke offline with @lynnjepsen about remove the toggle feature of the icon-button component, since icon button will be rewritten: see https://github.com/material-components/material-components-web/issues/3107. Looking at MDC Web foundation logic, its difficult to turn off without losing a11y support. See:

```
    this.on_ = this.adapter_.getAttr(ARIA_PRESSED) === 'true';
```

The only line we need to remove is the #adapter.getAttr line.

moog16:
created 2 new issues:

https://github.com/material-components/material-components-web-react/issues/160
https://github.com/material-components/material-components-web/issues/3162

packages/icon-button/index.js
```diff
@@ -0,0 +1,107 @@
...
+
+const {strings} = MDCIconButtonToggleFoundation;
+
+class IconButtonBase extends Component {
+
+  constructor(props) {
+    super(props);
+    this.state = {
+      classList: new Set(),
+      [strings.ARIA_PRESSED]: props[strings.ARIA_PRESSED],
```
bonniezhou: Do we need to do the same thing for `ARIA_LABEL`?

moog16: aria-label was removed from mdc web's icon-button

packages/ripple/index.js
```diff
@@ -26,7 +26,7 @@ import classnames from 'classnames';
 
 import {MDCRippleFoundation, util} from '@material/ripple/dist/mdc.ripple';
 
-const withRipple = (WrappedComponent) => {
+const withRipple = (WrappedComponent, isUnbounded = false) => {
```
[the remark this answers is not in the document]
moog16: ya you might be right. let me try that

packages/icon-button/index.js
bonniezhou: This can be updated now

test/unit/icon-button/index.test.js
```diff
@@ -0,0 +1,102 @@
...
+  assert.isFalse(wrapper.instance().adapter.hasClass('test-class'));
+});
+
+test('#adapter.setAttr sets aria-pressed', () => {
+  const wrapper = shallow(<IconButton />);
+  wrapper.instance().adapter.setAttr('aria-pressed', true);
+  assert.isTrue(wrapper.state()['aria-pressed']);
+});
+
+test('#adapter.setAttr sets aria-label', () => {
```
bonniezhou: This test can be removed

packages/icon-button/index.js
moog16: I think this comment no longer applies since the MDC Web component has a setter that calls foundation.toggle():
```
  set on(isOn) {
    this.foundation_.toggle(isOn);
  }```

moog16:
@bonniezhou which `handleFocus` method are you talking about?

bonniezhou:
The `handleFocus` method in `RippledComponent`. Filed #265

Issue #98, opened 2018-06-09
MaterialIcon component should not load the material icon font (blocks page load)


johnelliott:
Hello, I'm using the code in an app and totally enjoying the library:

## about app
- Static site
- Hosted on firebase
- Bundled/transpiled with webpack/babel-loader
- Bundled using webpack rules to import SCSS directly from node_modules/@material/*

## issue: slow rendering (fonts)

I noticed this when I ran Lighthouse (with emoji commentary)
😑 Opportunities
￼
![image](https://user-images.githubusercontent.com/5633265/41195764-b9f892ae-6c01-11e8-83f2-09df514eccca.png)

￼
🤔 ok cool, go find in devtools…
￼
![image](https://user-images.githubusercontent.com/5633265/41195766-c1d8c7be-6c01-11e8-8181-a93184719c68.png)

🤔 ok, go to bundle…

found this CSS import in https://github.com/material-components/material-components-web-react/tree/master/packages/material-icon: `@import url(https://fonts.googleapis.com/icon?family=Material+Icons);`

🤔 ok, go get that resource in chrome…

```css
@font-face {
  font-family: 'Material Icons';
  font-style: normal;
  font-weight: 400;
  src: url(https://fonts.gstatic.com/s/materialicons/v38/flUhRq6tzZclQEJ-Vdg-IuiaDsNa.woff) format('woff');
}

.material-icons {
  font-family: 'Material Icons';
  font-weight: normal;
  font-style: normal;
  font-size: 24px;
  line-height: 1;
  letter-spacing: normal;
  text-transform: none;
  display: inline-block;
  white-space: nowrap;
  word-wrap: normal;
  direction: ltr;
}
```
💪 ok cool, `MaterialIcon` has sass with this fonts.gstatic.com URL.

Basically any of these fonts in css files that are in HEAD will block rendering on the network if the user puts a link to the css file in head or in a style tag.

So I pre-loaded https://fonts.googleapis.com/icon?family=Material+Icons in the head of my HTML, and could preload the woff file also, but that is not really a stable way to go about it since any of the packages in the chain could break without warning.

## try to fix issue
I was able to configure webpack to consume the @ material sass directly, but because of this URL I don't know how to make this code lighthouse-friendly without putting it after the body. I may be doing something wrong at a higher level.

## solution?
**TL;DR:** Could we come up with a way to bundle in the font resources? Or SVGs? Maybe that takes the form of documentation or an example.

[...]

PR #155, opened 2018-07-19
fix(material-icon): removed blocking line and updated documentation

moog16:
fixes https://github.com/material-components/material-components-web-react/issues/98


packages/material-icon/README.md
```diff
@@ -22,6 +22,16 @@ with CSS:
 import '@material/react-material-icon/dist/material-icon.css';
 ```
 
+#### Importing Icons
+
+The React material icon package does not come packaged with [Google Font's Material Icons](https://google.github.io/material-design-icons/). If it was included in the Sass package this would block rendering of the page until the icons download.
+We recommend following the two methods of adding the Material Icons to your app documented [here](https://google.github.io/material-design-icons/#getting-icons). We have personally taken the `<link>` tag approach as it doesn't block page render. Code is pasted here for your convenience.
```
lynnjepsen: "following the two methods" ... this is unclear. Do you mean the developer has to do two things? Or the developer should pick one way or another. And which one is the `<link>` approach?

test/screenshot/webpack.config.js
```diff
@@ -1,14 +1,14 @@
 module.exports = [
-  ...require('./button/webpack.config.js'),
-  ...require('./card/webpack.config.js'),
-  ...require('./chips/webpack.config.js'),
-  ...require('./fab/webpack.config.js'),
-  ...require('./floating-label/webpack.config.js'),
-  ...require('./line-ripple/webpack.config.js'),
+  // ...require('./button/webpack.config.js'),
```
lynnjepsen: this should not be in this PR?

packages/material-icon/README.md
moog16: There are 2 ways that are documented within on material icons. How about...
```
We recommend following either of the 2 ways of adding Material Icons to your app, which are both documented ....

The 1st documented solution is the route our team has taken, since its not bundled with our CSS. This allows for quicker rendering, but a possibility of a Flash Of Unstyled Content (**FOUC**).
```

packages/material-icon/README.md
```diff
@@ -22,6 +22,22 @@ with CSS:
 import '@material/react-material-icon/dist/material-icon.css';
 ```
 
+#### Importing Icons
+
+The React material icon package does not come packaged with [Google Font's Material Icons](https://google.github.io/material-design-icons/). If it was included in the Sass package this would block rendering of the page until the icons download.
+
+
+We recommend following either of the 2 ways of adding Material Icons to your app, which are both documented [here](https://google.github.io/material-design-icons/#icon-font-for-the-web). The 1st documented solution is the route our team has taken, since its not bundled with our CSS. This allows for quicker rendering, but a possibility of a Flash Of Unstyled Content (**FOUC**). Code is pasted here for your convenience.
```
lynnjepsen: There are two different ways to add Material Icons to your website, both of which are documented in [Material Icons Documentation](https://google.github.io/material-design-icons/#icon-font-for-the-web). We recommend "Method 1: via Google Web Fonts", because it renders faster, but there can be Flash Of Unstyled Content (**FOUC**). The code to load Material Icons via Google Web Fonts is pasted here for your convenience.

Issue #127, opened 2018-07-06
Add Extended FAB


moog16:
Add extended fab as a variant in [Fab](https://github.com/material-components/material-components-web-react/tree/master/packages/fab).

https://github.com/material-components/material-components-web/blob/v0.37.0/packages/mdc-fab/mdc-fab.scss#L35

moog16:
fixed with https://github.com/material-components/material-components-web-react/pull/166

PR #159, opened 2018-07-20
chore(fab): Pass icon as a prop, not a child element

lynnjepsen:
This is to prepare us for the extended FAB (#127) where we need to accept both an icon and a label as input. Both icon and label will be passed as props to the `Fab` component


packages/fab/index.js
```diff
@@ -22,13 +22,15 @@ export class Fab extends React.Component {
...
-    });
-  };
+  renderIcon() {
+    const {icon} = this.props;
+
+    if (!icon) {
+      return;
+    }
+
+    return this.addClassesToElement('mdc-fab__icon', icon);
```
moog16: since you only use this `addClassesToElement` once, I think you should inline the logic:

```
const updatedProps = {
      className: classnames('mdc-fab__icon', icon.props.className),
    };

    return React.cloneElement(icon, updatedProps);
```

packages/fab/index.js
```diff
@@ -47,14 +49,22 @@ export class Fab extends React.Component {
           className={this.classes}
           ref={initRipple}
           {...otherProps}>
-          {this.addIconClassToAllChildren()}
+          {this.renderIcon()}
         </button>
     );
   }
+
+  addClassesToElement(classes, element) {
```
moog16: remove this method

packages/fab/index.js
```diff
@@ -47,14 +52,15 @@ export class Fab extends React.Component {
...
         </button>
     );
   }
 }
 
 Fab.propTypes = {
   mini: PropTypes.bool,
+  icon: PropTypes.element,
   className: PropTypes.string,
   children: PropTypes.oneOfType([
```
moog16: remove children from proptypes. I think you're planning on using `label` or `text` props instead.

packages/fab/index.js
```diff
@@ -47,14 +52,15 @@ export class Fab extends React.Component {
           className={this.classes}
           ref={initRipple}
           {...otherProps}>
-          {this.addIconClassToAllChildren()}
+          {this.renderIcon()}
```
moog16: comment for line above ^ remove children from list, since you don't use it

packages/fab/index.js
```diff
@@ -22,13 +22,18 @@ export class Fab extends React.Component {
...
   }
 
-  addIconClassToAllChildren = () => {
-    return React.Children.map(this.props.children, (item) => {
-      const className = `${item.props.className} mdc-fab__icon`;
-      const props = Object.assign({}, item.props, {className});
-      return React.cloneElement(item, props);
-    });
-  };
+  renderIcon() {
```
moog16: comment for lines above^

You are not using this.state.classList. Please remove that from state.

PR #167, opened 2018-07-23
fix(top-bar-bar): add missing material icon imports

moog16:
A few top app bar screenshot tests were missing material icon imports in the html files. This caused failures in the screenshot tests.


moog16:
branched from select :(

Issue #164, opened 2018-07-22
Add additional prop for button and fab


simsim0709:
For now, `button` and `fab` render fixed `<button>` element.
However, button like components is usually used with `<a>` element or another react components as well.

For example, when I want to use react-router's `Link` component with @material/react-button,
it should be possible:
```js
<Button component={Link} to="/home">button</Button>

// or

<Button tag="a" href="/somewhere">button</Button>
```

It's not possible, now.

If it's an acceptable feature request, I can make a pr.

Thanks.

moog16:
This seems like a valid request, since we're doing that in icon-button (see https://github.com/material-components/material-components-web-react/pull/154). However, I will have to check with MDC Web guidance to make sure we're consistent before moving ahead with these changes. Thanks for checking with us first. 

As a work around could you instead just wrap the <Button> or <Fab> with an anchor element?

simsim0709:
Yes, sure. I can do it for now.

I just want it to be implemented later, or I can help you with.

Thanks  👍

moog16:
@simsim0709 chatted with a teammate - and yes we agree Button should be able to return an anchor tag instead of a button. 

I had a second look at your 1st example of accepting a Link tag. I don't think that's really a use case for MDC React. I think you would instead want to do

```
<Link className='mdc-button' to='/home' />
```

wesleyabbey:
What about having an `href` prop that when passed in turned the button into an `<a/>` tag?
```
    const SemanticButton = href ? 'a' : 'button';
    return (
      <SemanticButton
        className={classes}
        ref={initRipple}
        {...otherProps}
      >
        {icon ? this.renderIcon() : null}
        {children}
      </SemanticButton>
```

wesleyabbey:
Yeah, I'll open one up

PR #174, opened 2018-07-25
feat(button): Add an href prop

wesleyabbey:
Adding a new href prop. This href prop will automatically convert the button into an anchor tag.

See issue #164


moog16:
@wesleyabbey There should be a screenshot test case for this, but I'm pretty sure you cannot do that. I'll try adding it to this PR

moog16:
Running tests here: https://github.com/material-components/material-components-web-react/pull/176

@wesleyabbey since I don't have access to your branch, could you please add the code in this commit: 
https://github.com/material-components/material-components-web-react/pull/176/commits/e66d886e4e690f74538bd7b0d6a2c4df99222b05

wesleyabbey:
I'm not sure why that test is failing. Is this an issue on my end?

PR #193, opened 2018-07-31
feat(bidirection): add package

moog16:
created package to share dir=rtl logic/code between text field and select

### DIFF
--- a/packages/bidirection/package.json
+++ b/packages/bidirection/package.json
@@ -1,27 +1,21 @@
 {
-  "name": "@material/react-text-field",
-  "version": "0.4.0",
-  "description": "Material Components React Text Field",
+  "name": "@material/bidirection",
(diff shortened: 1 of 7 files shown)


packages/bidirection/package.json
```diff
@@ -0,0 +1,25 @@
+{
+  "name": "@material/bidirection",
```
kfranqueiro: If this is for MDC React only, it should be `react-` prefixed.

If this is applicable to JS in general, we should consider adding it to `mdc-rtl` instead in MDC Web... but we should also look at current implementations in MDC Web (e.g. I'm pretty sure several adapters have `isRtl` functions)

moog16:
Moving into MDC Web

moog16:
created: https://github.com/material-components/material-components-web/issues/3274

Issue #194, opened 2018-07-31
Update Text Field to accept isRtl as a prop


moog16:
Moving rtl logic away from `getComputedStyle` to a prop aligns more with React methodology. It would look something like this:

```
<body>
  <div dir={this.state.rtl}>
    <TextField isRtl={this.state.rtl}/>
  </div>
</body>
```

The end developer will be responsible for having access to `this.state.rtl`. Also using getComputedStyle requires a style recalc per [this paulirish gist](https://gist.github.com/paulirish/5d52fb081b3570c81e3a).

PR #195, opened 2018-07-31
fix(text-field): update isRtl to a prop

moog16:
fixes https://github.com/material-components/material-components-web-react/issues/194

BREAKING CHANGE: In order for rtl to work on the Notched Outline component, isRtl will need to be passed in as boolean (default is false). Updating isRtl will result in a re-render.


test/unit/text-field/index.test.js
```diff
@@ -80,7 +80,7 @@ test('#componentDidUpdate calls setValue if state.value updates', () => {
 test('#componentDidUpdate does not call setValue if another property updates', () => {
   const wrapper = shallow(<TextField label='my label'><Input /></TextField>);
   wrapper.instance().foundation_.setValue = td.func();
-  wrapper.setState({dir: 'rtl'});
+  wrapper.setState({dir: 'rtl'}); // could be anything
```
bonniezhou: What does "could be anything" mean?

test/screenshot/webpack.config.js
```diff
@@ -1,14 +1,14 @@
 module.exports = [
```
bonniezhou: Uncomment these

test/unit/text-field/index.test.js
```diff
@@ -107,24 +107,22 @@ test('#adapter.removeClass removes class from state.classList', () => {
...
-test('#adapter.isRtl returns true if the direction is true', () => {
-  const div = document.createElement('div');
-  // needs to be attached to real DOM to get width
-  // https://github.com/airbnb/enzyme/issues/1525
-  document.body.style.direction = 'rtl';
-  document.body.append(div);
-  const options = {attachTo: div};
-  const wrapper = mount(
-    <TextField label='my label'><Input /></TextField>, options);
+test('#adapter.isRtl returns true props.isRtl is true', () => {
```
bonniezhou: forgot a word: "if"

test/unit/text-field/index.test.js
```diff
@@ -107,24 +107,22 @@ test('#adapter.removeClass removes class from state.classList', () => {
...
+  const wrapper = shallow(
+    <TextField isRtl label='my label'><Input /></TextField>);
   assert.isTrue(wrapper.instance().foundation_.adapter_.isRtl());
-  document.body.style.direction = 'initial';
-  div.remove();
+});
+
+test('#adapter.isRtl returns false props.isRtl is false', () => {
```
bonniezhou: forgot a word: "if"

packages/text-field/index.js
```diff
@@ -184,13 +184,6 @@ class TextField extends React.Component {
     };
   }
 
-  getIsRtl = () => {
-    if (this.textFieldElement.current) {
```
bonniezhou: Can we get rid of the ref to `textFieldElement` now?

packages/text-field/README.md
```diff
@@ -62,6 +62,7 @@ dense | Boolean | Enables dense variant.
 floatingLabelClassName | String | An optional class added to the floating label element fo the text field.
 fullWidth | Boolean | Enables fullWidth variant.
 helperText | Element | Helper text that appears below the text field. Use the `<HelperText>` component that comes with this package.
+isRtl | Boolean | Updates the `<NotchedOutline>` to the correct orientation (left or right).
```
bonniezhou: This description is an implementation detail, maybe say instead "Whether the direction of the text field element is set to RTL."

test/unit/text-field/index.test.js
moog16: I'll rewrite as  // could be any property/value pair. Or should i just remove the comment altogether?

PR #200, opened 2018-08-01
fix(ripple): add test to avoid computeBoundingRect error

### LINKED ISSUE #199: ripple - TypeError: Cannot read property 'getBoundingClientRect' of null
Sorry to open this, but it looks like issue that seemed mentioned in
https://github.com/material-components/material-components-web-react/issues/111#issuecomment-401171254 still exists.

Removing `TopAppBar` which includes `MaterialIcon` element produces TypeError:

```
index.js:124 Uncaught TypeError: Cannot read property 'getBoundingClientRect' of null
    at Object.computeBoundingRect (VM512 index.js:124)
    at MDCRippleFoundation.layoutInternal_ (VM512 index.js:482)
    at eval (VM512 index.js:441)
```

Here is a minimal test case - pressing *menu* will trigger TypeError:

https://codesandbox.io/s/721pv52661

Let me know if anything else is needed.

moog16:
fixes #199

### DIFF
--- a/packages/ripple/index.js
+++ b/packages/ripple/index.js
@@ -66,9 +66,16 @@ const withRipple = (WrappedComponent) => {
         registerResizeHandler: (handler) => window.addEventListener('resize', handler),
         deregisterResizeHandler: (handler) => window.removeEventListener('resize', handler),
         updateCssVariable: this.updateCssVariable,
-        computeBoundingRect: this.props.computeBoundingRect ?
-          (() => this.props.computeBoundingRect(instance)) :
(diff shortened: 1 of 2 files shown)


moog16:
It does fail without the fix. Removing the `if()` in the computeBoundingRect adapter method fails I didn't try assert.doesNotThrow since it is the `mouseDown` event that actually triggers the sequence of events. The error wouldn't throw until the raf() callback in the test. So the assert.doesNotThrow wouldn't work in this case. I really don't need an assert in the test, since what I'm testing for is no error throwing.

PR #208, opened 2018-08-06
feat(switch): Add component

### LINKED ISSUE #152: Add Switch

bonniezhou:
Fixes #152.

Ripple is modified because the element with the ripple surface (thumb underlay) is different from the ripple activating element (native input).


packages/switch/index.js
```diff
@@ -0,0 +1,92 @@
...
+    return (
+      <div
+        className={this.classes}
+        {...otherProps}
+      >
+        <div className='mdc-switch__track' />
+        <ThumbUnderlay
+          checked={this.state.checked}
+          disabled={disabled}
+          id={id}
```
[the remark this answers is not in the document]
bonniezhou: Yup! I renamed it `nativeControlId` to make it clearer

packages/switch/ThumbUnderlay.js
```diff
@@ -0,0 +1,77 @@
...
+
+export class ThumbUnderlay extends React.Component {
+  rippleActivatorEl = null;
+
+  init = (el) => {
+    this.props.initRipple(el, this.rippleActivatorEl);
+  }
+
+  get classes() {
+    return classnames('mdc-switch__thumb-underlay', this.props.className);
```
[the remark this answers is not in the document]
bonniezhou: I don't think so? The classes from index.js aren't passed down to ThumbUnderlay

packages/switch/index.js
```diff
@@ -0,0 +1,91 @@
...
+import classnames from 'classnames';
+import PropTypes from 'prop-types';
+import {MDCSwitchFoundation} from '@material/switch/dist/mdc.switch';
+
+import ThumbUnderlay from './ThumbUnderlay';
+
+export default class Switch extends Component {
+  foundation_ = null;
+  state = {
+    checked: this.props.checked,
```
moog16: oh cool - i didn't know this worked.

packages/switch/README.md
```diff
@@ -0,0 +1,55 @@
...
+
+```js
+import React, {Component} from 'react';
+import Switch from '@material/react-switch';
+
+class MyApp extends Component {
+  render() {
+    return (
+      <Switch nativeControlId='my-switch' />
+      <label for='my-switch'>My Switch</label>
```
moog16: for -> htmlFor

packages/switch/README.md
```diff
@@ -0,0 +1,55 @@
...
+```js
+import '@material/react-switch/index.scss';
+```
+
+with CSS:
+```js
+import "@material/react-switch/dist/switch.css";
+```
+
+### Javascript Instantiation
```
moog16: how does one get a hold of the value? if its checked/not checked

packages/switch/ThumbUnderlay.js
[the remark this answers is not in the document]
moog16: Gotcha - you're just not passing any classes from index.js to the ThumbUnderlay component so it seemed like you didn't need this. But I guess it doesn't hurt to have this.

packages/switch/ThumbUnderlay.js
```diff
@@ -0,0 +1,78 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import classnames from 'classnames';
+import withRipple from '../ripple';
```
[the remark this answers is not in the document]
bonniezhou: But ripple is modified in this PR so it won't work if it uses @material/ripple :(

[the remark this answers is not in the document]
moog16: even if you do `lerna add @material/ripple --scope=@material/react-switch`?

packages/switch/ThumbUnderlay.js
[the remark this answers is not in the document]
bonniezhou: Oh I see what you're saying. `ThumbUnderlay` is actually receiving `props.className` from `withRipple`, so it's necessary. On the other hand `NativeControl` isn't getting classes from anywhere, so I removed it there!

packages/switch/README.md
bonniezhou: Added a note to the README, basically same way as in text field

packages/switch/ThumbUnderlay.js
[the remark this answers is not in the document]
moog16: ah! i didn't even notice :)

packages/switch/ThumbUnderlay.js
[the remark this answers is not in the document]
bonniezhou: it worked!

packages/ripple/index.js
```diff
@@ -30,19 +30,22 @@ const withRipple = (WrappedComponent) => {
...
 
-    createAdapter_ = (instance) => {
+    createAdapter_ = (surface, activator) => {
       const MATCHES = util.getMatchesProperty(HTMLElement.prototype);
 
       return {
         browserSupportsCssVars: () => util.supportsCssVariables(window),
         isUnbounded: () => this.props.unbounded,
-        isSurfaceActive: () => instance[MATCHES](':active'),
+        isSurfaceActive: () => activator ? activator[MATCHES](':active') : surface[MATCHES](':active'),
```
lynnjepsen: this is so simple and lovely. I wonder if we could simplify the Vanilla version of Checkbox and Switch in MDCW to look more like this. 
https://github.com/material-components/material-components-web/blob/master/packages/mdc-switch/index.js#L79
https://github.com/material-components/material-components-web/blob/master/packages/mdc-checkbox/index.js#L63

packages/ripple/index.js
```diff
@@ -148,7 +145,13 @@ const withRipple = (WrappedComponent) => {
       this.setState({style: updatedStyle});
     }
 
-    getMergedStyles = () => {
+    get classes() {
+      const {className: wrappedCompClasses} = this.props;
```
lynnjepsen: wrappedComponentClasses. Don't shorten it

packages/switch/NativeControl.js
```diff
@@ -0,0 +1,40 @@
...
+    );
+  }
+}
+
+NativeControl.propTypes = {
+  checked: PropTypes.bool,
+  disabled: PropTypes.bool,
+  id: PropTypes.string,
+  onChange: PropTypes.func,
+  setRippleActivator: PropTypes.oneOfType([
```
lynnjepsen: can this just be `rippleActivator`, which I assume is an Element, since it gets passed to a ref

packages/switch/README.md
```diff
@@ -0,0 +1,62 @@
...
+}
+```
+
+> _NOTE_: In order to get access to the checked value, you must add an `onChange` handler that accepts an `Event` and updates a checked state as shown above.
+
+## Props
+
+Prop Name | Type | Description
+--- | --- | ---
+className | String | Classes to be applied to the chip element
```
lynnjepsen: hmmm, not a chip

packages/switch/index.js
```diff
@@ -0,0 +1,91 @@
...
+        const {classList} = this.state;
+        classList.add(className);
+        this.setState({classList});
+      },
+      removeClass: (className) => {
+        const {classList} = this.state;
+        classList.delete(className);
+        this.setState({classList});
+      },
+      isNativeControlChecked: () => this.state.checked,
```
lynnjepsen: is it weird that we implement `isNativeControlChecked` by look at the root component's state? I guess we are basically guaranteeing that if the native control's state changes, then ThumbUnderlay will respond in an `onChange` callback. And then if ThumbUnderlay's state changes we guarantee that Switch's state will change also with and `handleChange` callback. 

But for some reason that feels a bit fragile. Did we do the same design for the text field?

packages/switch/ThumbUnderlay.js
```diff
@@ -0,0 +1,79 @@
...
+    );
+  }
+}
+
+ThumbUnderlay.propTypes = {
+  checked: PropTypes.bool,
+  className: PropTypes.string,
+  disabled: PropTypes.bool,
+  nativeControlId: PropTypes.string,
+  handleChange: PropTypes.func,
```
lynnjepsen: why not `onChange`? Why `handleChange`?

packages/switch/index.js
bonniezhou: Yep, but text field only has one layer of nesting (is nesting the right word?) `TextField` -> `Input` which makes it seem less fragile. And similarly `Select`->`NativeControl`. Switch is the only component for now with 3 layers because it needs one for the native input, one for the ripple surface, and one for the top-level component.

packages/switch/ThumbUnderlay.js
bonniezhou: I think `on*` conventionally expects an event parameter. Confirm, @moog16?

packages/switch/NativeControl.js
bonniezhou: It's a `ref` which gets this PropType definition according to [StackOverflow](https://stackoverflow.com/questions/48007326/what-is-the-correct-proptype-for-a-ref). Changed the name to `rippleActivatorRef` to make it clearer.

packages/ripple/index.js
bonniezhou: We could probably modify `createAdapter` to take in an extra param as well

packages/switch/ThumbUnderlay.js
moog16: I've been using `on*` as an event handler. I think we invented `handle*` so it  wouldn't collide with an end developer. But in this case the end dev won't ever pass an onChange method so we can change it to `onChange`

bonniezhou: 👍 Updated!

lynnjepsen:
I pushed some changes to this PR. But We it now relies on the `MDCSwitchFoundation.handleChange` method taking an event as an argument. That is implemented in https://github.com/material-components/material-components-web/pull/3342, but it is not merged or released yet. So I think we might have to hold off merging this PR until that PR 3342 is merged.

Or we could do some work around for now? I'd rather wait than think about hacks, but maybe someone has a clever idea

moog16:
Ya I agree - no sense in making a work around and rush getting this PR in. We can wait until MDCW releases 0.39.0. Then we can update MDCR repo, and get this PR working. Then release 0.5.0

packages/switch/NativeControl.js
```diff
@@ -0,0 +1,35 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export default class NativeControl extends React.Component {
```
moog16: change this to stateless component:
```
export default (props) => {
  return (...)
}
```

packages/switch/NativeControl.js
```diff
@@ -0,0 +1,35 @@
...
+    const {
+      rippleActivatorRef,
+      ...otherProps
+    } = this.props;
+
+    return (
+      <input
+        type='checkbox'
+        role='switch'
+        className='mdc-switch__native-control'
```
moog16: Do you think people will use this component on its own? If so we should merge this.props.className and 'mdc-switch__native-control'

packages/switch/index.js
```diff
@@ -0,0 +1,114 @@
...
+import PropTypes from 'prop-types';
+import {MDCSwitchFoundation} from '@material/switch/dist/mdc.switch';
+
+import ThumbUnderlay from './ThumbUnderlay';
+import NativeControl from './NativeControl';
+
+export default class Switch extends Component {
+  rippleActivator = React.createRef();
+  foundation_ = null;
+  state = {
```
moog16: I think we should use constructor to set initial state:

```
constructor(props) {
  super(props);
   rippleActivator = React.createRef();
    foundation_ = null;
    this.state = {
     checked: props.checked,
      classList: new Set(),
      disabled: props.disabled,
      nativeControlChecked: props.checked,
      nativeControlDisabled: props.disabled,
    };
}

File: packages/switch/package.json [hunk 16]
```diff
@@ -0,0 +1,28 @@
+{
+  "name": "@material/react-switch",
+  "version": "0.0.0",
+  "description": "Material Components React Switch",
+  "license": "Apache-2.0",
```
moog16: MIT!

packages/switch/package.json
```diff
@@ -0,0 +1,28 @@
...
+    "material design",
+    "switch"
+  ],
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/material-components/material-components-web-react.git"
+  },
+  "dependencies": {
+    "@material/react-ripple": "^0.4.0",
+    "@material/ripple": "^0.38.0",
```
moog16: ripple needs to be 39

moog16: actually remove this and only rely on react-ripply

PR #215, opened 2018-08-09
fix(tab-indicator): update component to work with mdc tab

moog16:
- removed component management of `--active` class
- using foundation.active() method to manage `--active` class
- using ref to directly set content style instead of state. Animation did not always show
- put computeContentClientRect on component to be called by parent component (mdc-tab)
- moved tab indicator content child to render within content element. This allows the content element to accept a ref, instead of trying to pass it to the child (code is a lot more complex). 
  - this change requires us to have a override style for height/width on the `--icon .material-icons` element. Since .material-icons has a 24px font-size


packages/tab-indicator/index.scss
```diff
@@ -1 +1,6 @@
 @import "@material/tab-indicator/mdc-tab-indicator";
+
+.mdc-tab-indicator__content--icon .material-icons {
```
lynnjepsen: dont do this. Instead make the material-icons stylesheet load before the mdc-tab-indicator stylesheet

moog16: Actually the `.material-icons` class has the font-size: 24px. So rearranging the stylesheets will not help here.

I moved the logic around so that we now have the ref on the TabIndicatorElement. We then use .getElementsByClassName, which I think is fine in this case. The call is scoped to just the tabIndicator.

PR #216, opened 2018-08-09
feat(tab): add new component

moog16:
fixes https://github.com/material-components/material-components-web-react/issues/148


packages/tab-indicator/index.js
```diff
@@ -5,15 +5,14 @@ import classnames from 'classnames';
 import {
   MDCFadingTabIndicatorFoundation,
   MDCSlidingTabIndicatorFoundation,
-} from '@material/tab-indicator/dist/mdc.tabIndicator';
+} from '@material/tab-indicator';///dist/mdc.tabIndicator';
```
moog16: ignore this file - its included in #215

packages/tab/README.md
```diff
@@ -0,0 +1,90 @@
+# React Tab Indicator
```
moog16: This is for me TODO

packages/tab/index.js
```diff
@@ -0,0 +1,192 @@
...
+
+      registerEventHandler: () => this.allowTransitionEnd_ = true,
+      deregisterEventHandler: () => this.allowTransitionEnd_ = false,
+      getOffsetLeft: () => this.tabElement_.current && this.tabElement_.current.offsetLeft,
+      getOffsetWidth: () => this.tabElement_.current && this.tabElement_.current.offsetWidth,
+      getContentOffsetLeft: () => this.tabContentElement_.current && this.tabContentElement_.current.offsetLeft,
+      getContentOffsetWidth: () => this.tabContentElement_.current && this.tabContentElement_.current.offsetWidth,
+      focus: () => this.tabElement_.current && this.tabElement_.current.focus(),
+
+      // activateIndicator, deactivateIndicator, and computeIndicatorClientRect
```
moog16: I don't think activate/deactivate are necessary, but we can chat about the `computeIndicatorClientRect` method in person

packages/tab/index.js
```diff
@@ -0,0 +1,192 @@
...
+
+    if (indicator) {
+      const indicatorProps = {
+        active,
+        previousIndicatorClientRect: previousActiveClientRect,
+        ref: this.tabIndicatorElement_,
+      };
+      const Indicator = indicator(indicatorProps);
+      if (Indicator.type !== TabIndicator) {
+        new Error('this.props.indicator must be a function that returns an instance of TabIndicator');
```
moog16: This will be documented in more detail in the readme

packages/tab/index.js
```diff
@@ -0,0 +1,192 @@
...
+      },
+      removeClass: (className) => {
+        const classList = new Set(this.state.classList);
+        classList.delete(className);
+        this.setState({classList});
+      },
+      hasClass: (className) => this.classes.split(' ').includes(className),
+      setAttr: (attr, value) => this.setState({[attr]: value}),
+
+      registerEventHandler: () => this.allowTransitionEnd_ = true,
```
lynnjepsen: hmmmm, should we wait for MDCW to remove the need for registerEventHandler from the adapter?

packages/tab/index.js
lynnjepsen: Hmmm, I am out tomorrow but I do think we need to talk in person. I say we talk Monday. Maybe you and Pat can try and do a quick audit of the MDCW TabAdapter on Friday, so that you have some detailed technical information. But lets talk Monday and wait to merge until that talk.

packages/tab/index.js
```diff
@@ -0,0 +1,192 @@
...
+      ...otherProps
+    } = this.props;
+
+    const {
+      tabIndex,
+      ['aria-selected']: ariaSelected,
+    } = this.state;
+
+    return (
+      <button
```
lynnjepsen: the tab is a button??

packages/tab/index.js
lynnjepsen: HMMMMM

packages/tab/index.js
```diff
@@ -0,0 +1,192 @@
...
+        <span className='mdc-tab__ripple'></span>
+      </button>
+    );
+  }
+
+  renderIndicator() {
+    const {
+      active,
+      fadeIndicator,
+      indicator,
```
lynnjepsen: indicator is a prop? and its a function?

packages/tab/index.js
moog16: ya it is documented in MCW as a `<button>`. It makes sense to me, since the user's main action is to click it.

packages/tab/index.js
moog16: Ya this pattern is `renderProps`, something I brought up in chat. We could use our existing pattern that attaches classes, refs, other properties to the indicator, but this seems to give the end developer more control over what is rendered. 

This is definitely just a handle for advanced users that want something other than the default underline indicator. If they don't supply a `props.indicator` it will be default to underline.

lynnjepsen:
Remember to fix the package.json to install @material/tab

PR #220, opened 2018-08-14
Button iconPosition

oste:


moog16:
@oste let me check with the team to see if we want to include this. I think we'd want MDCW to support this first, since MDCR shouldn't be adding new functionality. 

Also we would want to add screenshot tests and unit tests.

agree with @kfranqueiro about `==`

oste:
Sounds great. Yeah, this was more of a rough draft and what I am using in my project for now. Will need to get up to speed on the proper PR procedures and hopefully a better solution.

PR #221, opened 2018-08-15
docs(text-field): Add HelperText Validity Usage Docs

### LINKED ISSUE #177: Add HelperText Usage Docs for use with validation prop
Hello,

I'm trying to get HelperText component to display as an error message like in material.io specs but I really don't know what props to set and when.

Also, if I set both `isValid` and `validation` to true, the helper text doesn't appear.
It's opacity is set to 0 (expected) but the transition isn't being run when I press the text field.

### Step to reproduce
Do the usual `create-react-app` setup and `npm install @material/helper-text`.
Import the sass file:
``` css
@import '@material/react-text-field/index.scss';
```
And use this App.js
``` js
import React, { Component } from 'react';
import TextField, {Input, HelperText} from '@material/react-text-field';
import './App.css';

class App extends Component {
  constructor(props) {
    super(props);
    this.state = {value:"hello"};
  }

  render() {
    return (
      <div className="App">
        <TextField
          box
          label='Dog'
          helperText={<HelperText isValid validation>Help Me!</HelperText>}
        >
          <Input
            value={this.state.value}
            onChange={(e) => this.setState({value: e.target.value})}/>
        </TextField>
      </div>
    );
  }
}

export default App;
```

### Expected behaviour
The helper text to appear, ideally in a different color.

### What actually happens
Nothing, the text helper doesn't appear.

### How I'll handle it for now

I'll simply use it without setting any props and do the color logic on my side but it'd be helpful if the component did this by itself (trembling animations and all), since it has the proper props.

But maybe I'm I doing something wrong ?
Maybe I forgot to include some scss files or I'm missing a character somewhere ?

Ranguna:
Fixes #177

### DIFF
--- a/packages/text-field/helper-text/README.md
+++ b/packages/text-field/helper-text/README.md
@@ -30,6 +30,39 @@ role | String | Sets role on HTML element.
 showToScreenReader | Boolean | toggles the aria-hidden property to allow element to be visible to the screen reader.
 validation | Boolean | If true, alters the helper text to an error message.
 
+## Input Validation
+
+HelperText provides validity styling by setting the `validation` prop in HelperText. Validation can be checked through the appropriate `Input` validation properties (pattern, min, max, require, step, minlength, maxlength).
+
+The following snipped is an example of how to use regex and check for minimum length with HelperText:
+``` js
+import React from 'react';
+import TextField, {HelperText, Input} from '@material/react-text-field';
+class MyApp extends React.Component {
+  state = {username: ''};
+  render() {
+    return (
+      <div>
+        <TextField
+          box
+          label='Username'
+          helperText={
+            <HelperText validation>Invalid username</HelperText>
+          }
+        >
+          <Input
+            pattern='^([a-zA-Z_]+[0-9]*)$'
+            minLength={8}
+            value={this.state.username}
+            onChange={(e)=>this.setState({username:e.target.value})}
+          />
+        </TextField>
+      </div>
+    );
+  }
+}
+```
+
 ## Sass Mixins
 
 Sass mixins may be available to customize various aspects of the Components. Please refer to the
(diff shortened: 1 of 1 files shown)


packages/text-field/helper-text/README.md
```diff
@@ -30,6 +30,39 @@ role | String | Sets role on HTML element.
 showToScreenReader | Boolean | toggles the aria-hidden property to allow element to be visible to the screen reader.
 validation | Boolean | If true, alters the helper text to an error message.
 
+## Input Validation
+
+HelperText provides validity styling by setting the `validation` prop in HelperText. Validation can be checked through the appropriate `Input` validation properties (pattern, min, max, require, step, minlength, maxlength).
```
moog16: require --> required

moog16: capitalize minlength/maxlength --> minLength/maxLength

packages/text-field/helper-text/README.md
```diff
@@ -30,6 +30,39 @@ role | String | Sets role on HTML element.
...
+
+The following snipped is an example of how to use regex and check for minimum length with HelperText:
```
moog16: - snipped --> snippet
- regex --> pattern regex

packages/text-field/helper-text/README.md
Ranguna: Should have reviewed those props, will change.

packages/text-field/helper-text/README.md
Ranguna: How did the `d` get in there, it's so far apart from the `t`.
I'll change that too.

Ranguna:
I signed it!

moog16:
Just a docs change, so ignoring tests

Ranguna:
No problem, glad I could help :+1:

PR #226, opened 2018-08-17
fix(tab): update readme and ref name in tab component

moog16:


package.json
```diff
@@ -34,6 +34,7 @@
           "line-ripple",
           "ripple",
           "top-app-bar",
+          "tab",
           "infrastructure",
```
bonniezhou: Also add tab-scroller here and to `devDependencies`

PR #229, opened 2018-08-20
feat(tab-bar): Add component

### LINKED ISSUE #151: Add Tab Bar

bonniezhou:
Fixes #151 

[MDC Web #3386](https://github.com/material-components/material-components-web/pull/3386) fixes the issue where activating an already-activated tab will deactivate it.

Changes in `Tab`:
- update `previousIndicatorClientRect` from state (passed into the adapter) instead of from props
- add `activate`/`deactivate` methods for `TabBar` to call
- add custom event details to click event 
- remove `isIconIndicator` prop because of redundancy

Changes in `TabScroller`:
- set the style of scroll content and scroll area from state


bonniezhou:
Still writing tests for tab bar but I figured the rest of it can get reviewed in the meantime

packages/tab-bar/index.js
```diff
@@ -0,0 +1,145 @@
...
+      ...otherProps
+    } = tab.props;
+
+    const handleTabInteraction = (e) => {
+      this.foundation_.handleTabInteraction(e);
+      onClick(e);
+    };
+
+    return (
+      <Tab
```
moog16: I think we can use React.cloneElement here instead of using <Tab>. I know we want our Tab to be used, but they might have some custom Tab that they want to use.

packages/tab-bar/package.json
```diff
@@ -0,0 +1,30 @@
+{
+  "name": "@material/react-tab-bar",
+  "version": "0.4.1",
```
moog16: change to 0.0.0

packages/tab-bar/package.json
```diff
@@ -0,0 +1,30 @@
...
+    "tab bar",
+    "tabbar",
+    "tabs"
+  ],
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/material-components/material-components-web-react.git"
+  },
+  "dependencies": {
+    "@material/react-tab": "^0.0.0",
```
moog16: if we use cloneElement we can remove this dep

packages/tab/index.js
```diff
@@ -113,6 +130,12 @@ export default class Tab extends Component {
         role='tab'
         aria-selected={ariaSelected}
         tabIndex={tabIndex}
+        onClick={(e) => {
+          if (onClick) {
+            const tabEvent = new e.constructor(e.type, e, {detail: {tab: this}});
```
moog16: what is happening here?, you cannot pass the evt? i dont think MDCW is doing this so why do we?

test/screenshot/tab-bar/index.js
```diff
@@ -0,0 +1,70 @@
+import React from 'react';
+import ReactDOM from 'react-dom';
+import Tab from '../../../packages/tab-bar';
```
moog16: tab-bar --> tab

test/screenshot/webpack.config.js
```diff
@@ -1,18 +1,19 @@
...
-  ...require('./button/webpack.config.js'),
-  ...require('./card/webpack.config.js'),
-  ...require('./chips/webpack.config.js'),
-  ...require('./fab/webpack.config.js'),
-  ...require('./floating-label/webpack.config.js'),
-  ...require('./line-ripple/webpack.config.js'),
-  ...require('./material-icon/webpack.config.js'),
-  ...require('./notched-outline/webpack.config.js'),
-  ...require('./select/webpack.config.js'),
+  // ...require('./button/webpack.config.js'),
```
moog16: uncomment :)

packages/tab/index.js
bonniezhou: `MDCTabBarFoundation.handleTabInteraction` requires that the tab be passed into the `event.detail`, MDCW does this in the custom `MDCTab:interacted` event

packages/tab-bar/README.md
```diff
@@ -0,0 +1,69 @@
...
+> _NOTE_: You can also use a custom tab component with the `TabBar`, but it must implement the methods `activate`, `deactivate`, `focus`, `computeIndicatorClientRect`, and `computeDimensions`. See [`MDCTab` documentation](https://github.com/material-components/material-components-web/blob/master/packages/mdc-tab/README.md#mdctab-properties-and-methods) for more details.
+
+## Props
+
+Prop Name | Type | Description
+--- | --- | ---
+activeIndex | number | Index of the active tab.
+handleActiveIndexUpdate | Function(activeIndex: number) => void | Callback after the active index is updated.
+className | string | Classes to appear on className attribute of root element.
+isRTL | Boolean |  Whether the direction of the tab bar is RTL.
```
moog16: lets change this to `isRtl` to be consistent with the rest of the repo

packages/tab-bar/index.js
```diff
@@ -0,0 +1,146 @@
...
+      getTabIndicatorClientRectAtIndex: (index) => this.tabList_[index].computeIndicatorClientRect(),
+      getTabDimensionsAtIndex: (index) => this.tabList_[index].computeDimensions(),
+      getActiveTabIndex: () => this.props.activeIndex,
+      getFocusedTabIndex: () => {
+        const activeElement = document.activeElement;
+        return this.tabList_.indexOf((tab) => tab.tabElement_.current === activeElement);
+      },
+      getIndexOfTab: (tabToFind) => this.tabList_.indexOf(tabToFind),
+      getTabListLength: () => this.tabList_.length,
+      notifyTabActivated: (index) => this.props.handleActiveIndexUpdate(index),
```
moog16: usually we omit these notify* methods. I think  you should have a componentDidUpdate() that checks for props.activeIndex. Then calls this.activateTab(index)

packages/tab-bar/index.js
```diff
@@ -0,0 +1,146 @@
...
+    };
+  }
+
+  activateTab(index) {
+    this.foundation_.activateTab(index);
+  }
+
+  scrollIntoView(index) {
```
moog16: Also what if this was in the componentDidUpdate() method as well? And we created a new prop, like `props.tabToScroll`?

I just say that since its the user will need a `ref` to the tabbar in order to scrollIntoView(). I think we should minimize the amount we force the user to use a ref.

packages/tab/index.js
moog16: I think MDCW shouldn't be using the tab in the foundation code. Its strange that we need to construct an event in this Tab component for MDC Tab Bar to consume. 

I like your idea of using id's like with chips

packages/tab-bar/index.js
```diff
@@ -0,0 +1,146 @@
...
+        <TabScroller ref={this.tabScroller_}>
+          {React.Children.map(children, this.renderTab)}
+        </TabScroller>
+      </div>
+    );
+  }
+
+  renderTab = (tab) => {
+    const {
+      /* eslint-disable no-unused-vars */
```
moog16: remove these, since they're not on tab.props.

packages/tab/index.js
bonniezhou: Talked offline, sounds good. We can call `foundation.activateTab` directly with the tab's index (instead of assigning an id) and bypass `foundation.handleTabInteraction` entirely

bonniezhou:
Commit f03f897 depends on [MDCW #3394](https://github.com/material-components/material-components-web/pull/3394)

packages/tab-bar/index.js
```diff
@@ -0,0 +1,165 @@
...
+      setActiveTab: (index) => this.props.handleActiveIndexUpdate(index),
+      activateTabAtIndex: (index, clientRect) => this.tabList_[index].activate(clientRect),
+      deactivateTabAtIndex: (index) => this.tabList_[index].deactivate(),
+      focusTabAtIndex: (index) => this.tabList_[index].focus(),
+      getTabIndicatorClientRectAtIndex: (index) => this.tabList_[index].computeIndicatorClientRect(),
+      getTabDimensionsAtIndex: (index) => this.tabList_[index].computeDimensions(),
+      getPreviousActiveTabIndex: () => this.state.previousActiveIndex,
+      getFocusedTabIndex: () => {
+        const activeElement = document.activeElement;
+        for (let i = 0; i < this.tabList_.length; i++) {
```
moog16: Use a forEach here instead of a for.

Issue #173, opened 2018-07-24
Update Chip with public foundation handlers


bonniezhou:
Update Chip to call handlers which were made public in [#3146](https://github.com/material-components/material-components-web/pull/3146).

Issue #153, opened 2018-07-18
Support icons in chips


bonniezhou:

lynnjepsen:
This is adding a new prop, similar to how we did Button and TopAppBar
https://github.com/material-components/material-components-web-react/blob/master/packages/button/index.js#L59
https://github.com/material-components/material-components-web-react/blob/master/packages/top-app-bar/index.js#L183

PR #246, opened 2018-08-24
feat(chips): Update chips to use ids and add input chips

### LINKED ISSUE #131: Add input chips

bonniezhou:
Fixes #131, #153,  #173.

### DIFF
--- a/packages/chips/Chip.js
+++ b/packages/chips/Chip.js
@@ -99,6 +99,10 @@ export class Chip extends Component {
         'mdc-chip__icon',
         'mdc-chip__icon--trailing',
       ),
+      onClick: this.handleRemoveIconClick,
+      onKeyDown: this.handleRemoveIconClick,
+      tabIndex: 0,
+      role: 'button',
       ...otherProps,
     };
 
@@ -127,6 +131,7 @@ export class Chip extends Component {
 
     return (
       <div
+        tabIndex='0'
         className={this.classes}
         onClick={this.handleClick}
         onTransitionEnd={this.handleTransitionEnd}
--- a/packages/chips/Chip.js
+++ b/packages/chips/Chip.js
@@ -13,6 +13,13 @@ export class Chip extends Component {
   componentDidMount() {
     this.foundation_ = new MDCChipFoundation(this.adapter);
     this.foundation_.init();
+    this.foundation_.setSelected(this.props.selected);
+  }
+
+  componentDidUpdate(prevProps) {
+    if (this.props.selected !== prevProps.selected) {
+      this.foundation_.setSelected(this.props.selected);
+    }
   }
 
   componentWillUnmount() {
@@ -21,10 +28,8 @@ export class Chip extends Component {
 
   get classes() {
     const {classList} = this.state;
-    const {className, selected} = this.props;
-    return classnames('mdc-chip', Array.from(classList), className, {
-      'mdc-chip--selected': selected,
-    });
+    const {className} = this.props;
+    return classnames('mdc-chip', Array.from(classList), className);
   }
 
   get adapter() {
@@ -52,14 +57,17 @@ export class Chip extends Component {
 
   render() {
     const {
-      className, // eslint-disable-line no-unused-vars
-      label,
-      handleSelect, // eslint-disable-line no-unused-vars
-      onClick, // eslint-disable-line no-unused-vars
-      chipCheckmark,
-      computeBoundingRect, // eslint-disable-line no-unused-vars
+      /* eslint-disable no-unused-vars */
+      className,
+      selected,
+      handleSelect,
+      onClick,
+      computeBoundingRect,
+      unbounded,
+      /* eslint-enable no-unused-vars */
       initRipple,
-      unbounded, // eslint-disable-line no-unused-vars
+      chipCheckmark,
+      label,
       ...otherProps
     } = this.props;
 
@@ -78,7 +86,7 @@ export class Chip extends Component {
 }
 
 Chip.propTypes = {
-  id: PropTypes.number,
+  id: PropTypes.string,
   label: PropTypes.string,
   className: PropTypes.string,
   selected: PropTypes.bool,
@@ -92,7 +100,7 @@ Chip.propTypes = {
 };
 
 Chip.defaultProps = {
-  id: -1,
+  id: '',
   label: '',
   className: '',
   selected: false,
--- a/packages/chips/Chip.js
+++ b/packages/chips/Chip.js
@@ -5,6 +5,7 @@ import withRipple from '../ripple';
 import {MDCChipFoundation} from '@material/chips/dist/mdc.chips';
 
 export class Chip extends Component {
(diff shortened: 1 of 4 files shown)


packages/chips/ChipSet.js
```diff
@@ -41,6 +54,20 @@ export default class ChipSet extends Component {
     };
   }
 
+  handleSelect = (chipId) => {
+    if (this.props.filter || this.props.choice) {
+      this.foundation_.toggleSelect(chipId);
+    }
+    this.props.handleSelect(this.foundation_.getSelectedChipIds());
```
bonniezhou: `MDCChipSetFoundation` already keeps a list of the select chip ids internally, so we can use that instead of keeping one in the React ChipSet. This means that `adapter.setSelected` won't be used.

Talked offline with @moog16: an alternative we considered is instead of having the user pass a `handleSelect(allSelectedChipIds)` callback, have the user pass a `setSelected(chipId, selected)` callback which will be called from `adapter.setSelected` (within `foundation.toggleSelect`).

packages/chips/ChipSet.js
```diff
@@ -41,6 +54,20 @@ export default class ChipSet extends Component {
     };
   }
 
+  handleSelect = (chipId) => {
```
bonniezhou: Bypassing `foundation.handleChipInteraction` so we don't need to pass the id in `event.details`

packages/chips/ChipSet.js
```diff
@@ -41,6 +54,20 @@ export default class ChipSet extends Component {
...
+    if (this.props.filter || this.props.choice) {
+      this.foundation_.toggleSelect(chipId);
+    }
+    this.props.handleSelect(this.foundation_.getSelectedChipIds());
+  }
+
+  handleRemove = (chipId) => {
```
bonniezhou: Bypassing `foundation.handleChipRemoval` so we don't need to pass the id in `event.details`

packages/chips/Chip.js
```diff
@@ -35,18 +36,28 @@ export class Chip extends Component {
   componentDidMount() {
     this.foundation_ = new MDCChipFoundation(this.adapter);
     this.foundation_.init();
+    this.foundation_.setSelected(this.props.selected);
```
bonniezhou: Move adding/removing of the `mdc-chip--selected` class to here and `componentDidUpdate`

packages/chips/Chip.js
```diff
@@ -35,18 +36,28 @@ export class Chip extends Component {
...
+      this.foundation_.setSelected(this.props.selected);
+    }
   }
 
   componentWillUnmount() {
     this.foundation_.destroy();
   }
 
+  init = (el) => {
+    this.chipElement_ = el;
```
bonniezhou: Required for implementing adapter methods

packages/chips/Chip.js
```diff
@@ -72,53 +87,112 @@ export class Chip extends Component {
     this.props.handleSelect(this.props.id);
   }
 
+  handleRemoveIconClick = (e) => this.foundation_.handleTrailingIconInteraction(e);
+
+  handleTransitionEnd = (e) => this.foundation_.handleTransitionEnd(e);
```
bonniezhou: This calls the adapter method `notifyRemoval`. If we're aiming to not implement any "notify" adapter methods, we could move `props.handleRemove` to be called directly here instead.

packages/chips/ChipSet.js
```diff
@@ -23,24 +23,84 @@
...
+      },
     };
   }
 
+  updateChipSelection() {
+    React.Children.forEach(this.props.children, (child) => {
+      const {id} = child.props;
+      if (this.props.selectedChipIds.indexOf(id) > -1) {
+        this.foundation_.select(id);
+      } else {
```
moog16: what if we had a `child.props.selected` check here? So we don't always call deselected if the chip is already deselected.

packages/chips/README.md
```diff
@@ -131,8 +173,11 @@ Prop Name | Type | Description
 className | String | Classes to be applied to the chip element
 id | Number | Unique identifier for the chip
```
moog16: id is require, right?

packages/chips/README.md
```diff
@@ -85,30 +77,75 @@ Filter chips include a leading checkmark to indicate selection. To define a set
...
+      // See: https://reactjs.org/docs/state-and-lifecycle.html#do-not-modify-state-directly
+      const chips = [...this.state.chips]; 
+      chips.push({label, id});
+      this.setState({chips});
+      e.target.value = '';
     }
-    this.setState({selectedChipIds});
+  }
+
+  handleRemove = (id) => {
```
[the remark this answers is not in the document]
moog16: I commented out the code in the screenshot tests and it works fine. I don't think we need this in the docs

test/screenshot/chips/index.js
```diff
@@ -2,73 +2,139 @@ import React from 'react';
...
-        <Chip selected={this.isSelected(2)} id={2} label='Shoes' handleSelect={this.handleSelect}/>
-      </ChipSet>
+      <div>
+        <input
+          type="text"
+          onKeyDown={this.handleKeyDown}
+        />
+        <ChipSet
+          input
+          handleRemove={(chipId) => {
```
moog16: This can be removed

packages/chips/ChipSet.js
```diff
@@ -23,24 +23,84 @@
...
+    selectedChipIds: new Set(this.props.selectedChipIds),
+  };
+
+  componentDidMount() {
+    this.foundation_ = new MDCChipSetFoundation(this.adapter);
+    this.foundation_.init();
+    this.updateChipSelection();
+  }
+
+  componentDidUpdate() {
```
moog16: i actually think we should have this check again

```
    if (this.props.selectedChipIds !== prevProps.selectedChipIds) {	
      this.updateChipSelection();	
    }	
  }```

File: packages/chips/README.md (hunk 9 above)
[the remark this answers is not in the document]
Comment (bonniezhou) [reply packages/chips/README.md]: `handleRemove` removes the chip DOM element. If you comment out the code and open the debug tool, you'll see that the DOM element is still on the page even though it's hidden.

File: test/screenshot/chips/index.js (hunk 10 above)
Comment (bonniezhou) [reply test/screenshot/chips/index.js]: See other comment on handleRemove

File: test/screenshot/chips/index.js [hunk 12]
```diff
@@ -2,73 +2,140 @@ import React from 'react';
...
+        <input
+          type="text"
+          onKeyDown={this.handleKeyDown}
+        />
+        <ChipSet
+          input
+          // handleRemove removes the chip element from the page
+          handleRemove={(chipId) => {
+            const {chips} = this.state;
+            const chip = chips.find((chip) => chip.id === chipId);
```
moog16: I think .find is not suppported by IE
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find

Issue #172, opened 2018-07-24
react-ripple does not contain styles


eric-zeng:
Getting this error when attempting to follow the example for react-ripple:

`Uncaught Error: Cannot find module '@material/react-ripple/index.scss'`

Looks like neither index.scss nor dist/ripple.css exists in the react-ripple npm package.

<img width="239" alt="screen shot 2018-07-24 at 12 59 12 am" src="https://user-images.githubusercontent.com/2206755/43124832-1d0af906-8edd-11e8-9f26-acba2e1689fb.png">

moog16:
@eric-zeng you're right - we missed adding the index.scss file for ripple. In our screenshot tests we actually  import the sass from material-web. 

As a work around until we release again, you should be able to import the sass like so:

```scss
@import "@material/ripple/common";
@import "@material/ripple/mixins";
```

moog16:
closing: https://github.com/material-components/material-components-web-react/pull/251

PR #251, opened 2018-08-28
fix(ripple): react-ripple does not contain styles (#172)

bmihelac:


bmihelac:
Not sure why travis screesnshots tests are failing but, I notice that they are also failing with documentation update from https://github.com/material-components/material-components-web-react/pull/250

moog16:
@bmihelac screenshot tests from external developers are flaky. I'm not sure why either, and it requires some investigation on our part (we have an issue open).

Thanks for fixing this issue as well!

fixes  https://github.com/material-components/material-components-web-react/issues/172

moog16:
Tests passing https://github.com/material-components/material-components-web-react/pull/252

Issue #248, opened 2018-08-27
Missing ripple on card


jcburnette:
Hello,

I'm experiencing a problem and I'm not sure if it's something I'm doing wrong or not. I'm using the card component (0.4.2) in a project generated by create react app. When I click on the card, I see the background fade it but the animation for the ripple itself does not appear to fire.

Here is a gif of the behavior I'm seeing:

https://imgur.com/a/3qQkhcW

As you can see, the ripple does not behave in the expected way. As a side note, I have created a clean react application and installed both 0.4.2 and 0.4.1 and both experience this behavior. I have also tried wrapping an element using ripple and it does not work either. The card in rmwc works, however.

Am I doing something wrong or is this a bug in the package?

moog16:
@jcburnette We're missing ripple on the card, and this is a bug. We have a React ripple HOC. You should be able to use that in the meantime until we fix this. We'll get on this bug, but if you figure it out before we fix it, please feel free to open a PR.

Thanks for reporting!

bmihelac:
If I understand correctly, [CardPrimaryContent](https://github.com/material-components/material-components-web-react/blob/master/packages/card/PrimaryContent.js) should be decorated with `withRipple`. For consistency with `MaterialIcon` component, new prop `hasRipple` should be added to `CardPrimaryContent` to be able to have both variant with and without ripple.

moog16:
@bmihelac that's right, the CardPrimaryContent is the only component we provide in Card that should have the ripple on it. Buttons and icons should too, but the end developer is responsible for using the MDCButton or MaterialIcon to attach the ripple to each individual button or icon.

moog16:
are you gonna fix this one too? :)

bmihelac:
sure, I'll try

moog16:
sweet thanks!

moog16:
sweet thanks! If you need help please reach out - let me know if you wanna chat too. I can screenshare or hop onto discord as well. If you think the withRipple HOC can be improved, I'm definitely open to hearing that too.

bmihelac:
sure, will let you know here or on discourse

moog16:
fixed in https://github.com/material-components/material-components-web-react/pull/255

PR #255, opened 2018-08-30
fix(card): add missing hasRipple (#248)

bmihelac:


packages/card/PrimaryContent.js
```diff
@@ -24,33 +24,67 @@ import React from 'react';
...
-      >
-        {children}
-      </div>
+      <PrimaryContentDefault {...otherProps} />
     );
   }
 }
 
+
+const PrimaryContentDefault = (props) => {
```
moog16: Instead of PrimaryContentDefault, can we name it PrimaryContentBase?

bmihelac: No problem. That should probably be renamed as well in material-icon:

https://github.com/material-components/material-components-web-react/blob/master/packages/material-icon/index.js#L57-L80

bmihelac:
Good question. Looking mdc-card it looks like there is no use-case  for PrimaryContent without ripple:

> Only applicable to cards that have a primary action that the main surface should trigger.

I can refactor this PR to remove `hasRipple` prop and export decorated with `withRipple` component instead.

moog16:
@bmihelac that sounds good to me!

bmihelac:
@moog16 PR has been refactored (would rebase when add tests)

One thing that I am not sure how to do is testing that PrimaryContent has ripple. Here is my not working code:

https://github.com/material-components/material-components-web-react/blob/728a0ee0c3f01a948c74ded4398c1ff44ac12d40/test/unit/card/Content.test.js#L15-L21

Any idea how to test this?

moog16:
@bmihelac figured it out...the upgraded class is added in a requestAnimationFrame. So just wrap the assert in a requestAnimationFrame.

moog16:
@bmihelac - instead of wrapping it in a rAF, please use the rAF helper we have https://github.com/material-components/material-components-web-react/blob/master/test/unit/helpers/raf.js. Example seen in (https://github.com/material-components/material-components-web-react/blob/master/test/unit/ripple/index.test.js#L27)[ripple unit test].

bmihelac:
Thanks @moog16, that worked perfectly.
I have added missing test and rebased PR.

bmihelac:
Let me know if anything else is needed regarding this.

moog16:
looks good to me. I'm going to open another pr for tests

moog16:
Once CI passes on this PR: https://github.com/material-components/material-components-web-react/pull/263 we can merge this PR.

@bmihelac can you please sign this PR?

bmihelac:
> @bmihelac can you please sign this PR?

signed.

moog16:
CI Passing: https://travis-ci.org/material-components/material-components-web-react/builds/430245677?utm_source=github_status&utm_medium=notification

Merging!

bmihelac:
thank you @moog16 !

PR #261, opened 2018-09-17
feat(tab): Add ripple

bonniezhou:
Create `TabRipple` component used in `Tab`


test/unit/tab/index.test.js
```diff
@@ -212,12 +212,6 @@ test('should render content with icon and text children', () => {
   assert.equal(textLabel.type(), 'span');
 });
 
-test('should render mdc tab ripple', () => {
```
moog16: I think you can change this to find the <TabRipple /> element, instead of doing `.hasClass`

PR #278, opened 2018-09-24
Update button css paths with correct filenames

kevinsimper:
This confused me a little bit, looked into the `node_modules` and found the correct paths.


kevinsimper:
I looked into the button component and not into the react-button folder, my mistake

Issue #276, opened 2018-09-24
TabBar arrow key throws error


bonniezhou:
In `getFocusedTabIndex()`, the `return` statement doesn't break a `forEach`, so it goes through the entire tab list and returns undefined.

moog16:
fixed in #280

PR #280, opened 2018-09-24
fix(tab-bar): Arrow keys should focus adjacent tab

### LINKED ISSUE #279: Focus is not visible on tabs
When you click on a tab, the ripple activates but the ripple's focus background color is not present.

bonniezhou:
Fixes #279 and #276

### DIFF
--- a/packages/tab-bar/index.js
+++ b/packages/tab-bar/index.js
@@ -63,6 +63,8 @@ export default class TabBar extends Component {
       getPreviousActiveTabIndex: () => this.state.previousActiveIndex,
       getFocusedTabIndex: () => {
         const activeElement = document.activeElement;
+        // cannot use findIndex, not supported in IE11
+        // cannot use forEach, return statement inside the forEach loop will not return
         for (let i = 0; i < this.tabList_.length; i++) {
           if (this.tabList_[i].tabElement_.current === activeElement) {
             return i;
--- a/packages/tab-bar/index.js
+++ b/packages/tab-bar/index.js
@@ -63,11 +63,12 @@ export default class TabBar extends Component {
       getPreviousActiveTabIndex: () => this.state.previousActiveIndex,
       getFocusedTabIndex: () => {
         const activeElement = document.activeElement;
-        this.tabList_.forEach((tabList, index) => {
-          if (tabList.tabElement_.current === activeElement) {
-            return index;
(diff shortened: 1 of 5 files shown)


packages/tab-bar/index.js
```diff
@@ -63,11 +63,12 @@ export default class TabBar extends Component {
       getPreviousActiveTabIndex: () => this.state.previousActiveIndex,
       getFocusedTabIndex: () => {
         const activeElement = document.activeElement;
-        this.tabList_.forEach((tabList, index) => {
-          if (tabList.tabElement_.current === activeElement) {
-            return index;
+        for (let i = 0; i < this.tabList_.length; i++) {
```
moog16: I don't think you need this change.

test/unit/tab/index.test.js
```diff
@@ -305,3 +305,25 @@ test('#componentWillUnmount destroys foundation', () => {
   wrapper.unmount();
   td.verify(foundation.destroy(), {times: 1});
 });
+
+test('on focus event calls handleFocus on TabRipple', () => {
+  const wrapper = shallow(<Tab />);
```
moog16: It looks like you're trying to get the ref here...you can just use `mount` instead of shallow. Then these 3 lines can just be:

```
const wrapper = mount(<Tab />);
const ripple = wrapper.instance(). tabRipple_.current;
```

test/unit/tab/index.test.js
```diff
@@ -305,3 +305,25 @@ test('#componentWillUnmount destroys foundation', () => {
...
+  wrapper.instance().tabRipple_ = {current: ripple};
+  ripple.handleFocus = td.func();
+
+  const evt = {test: 'test'};
+  wrapper.simulate('focus', evt);
+  td.verify(ripple.handleFocus(evt), {times: 1});
+});
+
+test('on blur event calls handleBlur on TabRipple', () => {
+  const wrapper = shallow(<Tab />);
```
moog16: same here

packages/tab-bar/index.js
bonniezhou: This is what's causing #276, the `forEach` is returning undefined

moog16: ahh gotcha - lets just leave a comment so no one changes it back to a forEach or something

bonniezhou:
Good call 👍

PR #282, opened 2018-09-24
fix(chips): remove handleRemove method for chipsUpdate method

moog16:
Removes the responsibility of removing the chip from the user into the component. Instead of having a handleRemove prop method, I renamed to chipsUpdate which is passed an array of chip props.

### DIFF
--- a/packages/chips/ChipSet.js
+++ b/packages/chips/ChipSet.js
@@ -94,13 +94,32 @@ export default class ChipSet extends Component {
   }
 
   handleRemove = (chipId) => {
-    const {input, handleRemove} = this.props;
+    const {input} = this.props;
     if (input) {
       // this should be calling foundation_.handleChipRemoval, but we would
       // need to pass evt.detail.chipId
       this.foundation_.deselect(chipId);
     }
-    handleRemove(chipId);
+    this.removeChip(chipId);
+  }
+
+  // this should be adapter_.removeChip, but cannot be complete until
+  // https://github.com/material-components/material-components-web/issues/3613
+  // is fixed
+  removeChip = (chipId) => {
+    const {chipsUpdate, children} = this.props;
+    if (!children) return;
+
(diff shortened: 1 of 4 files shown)


packages/chips/ChipSet.js
```diff
@@ -94,13 +94,32 @@ export default class ChipSet extends Component {
...
+
+    const chips = React.Children.toArray(children).slice();
+    for (let i = 0; i < chips.length; i ++) {
+      const chip = chips[i];
+      if (chip.props.id === chipId) {
+        chips.splice(i, 1);
+        break;
+      }
+    }
+    const chipsArray = chips.length ? chips.map((chip) => chip.props) : [];
```
bonniezhou: I don't think the length check is necessary, and I don't know if we want to expose all the chip props to the user. How about 
```
const chipsArray = chips.map((chip) => {
  const {label, id} = chip.props;
  return {label, id};
});
```

packages/chips/ChipSet.js
```diff
@@ -144,7 +163,7 @@ ChipSet.propTypes = {
   className: PropTypes.string,
   selectedChipIds: PropTypes.array,
   handleSelect: PropTypes.func,
-  handleRemove: PropTypes.func,
+  chipsUpdate: PropTypes.func,
```
bonniezhou: Maybe `updateChips`? Makes it sound more obviously a function

packages/chips/ChipSet.js
bonniezhou: oh true, is it possible to just get the props that the user added?
```
  const chipsArray = chips.map((chip) => {
      const {
        handleSelect,
        handleRemove,
        // other props we wanna hide,
        ...otherProps,
      } = props;
      return {...otherProps};
    });
```

bonniezhou: Actually nvm, what if they override one of them. I'm fine with as is

PR #292, opened 2018-09-26
chore(select): update to 40 and add line ripple center support

moog16:
fixes https://github.com/material-components/material-components-web-react/issues/290


moog16:
Correct! I wont merge this until #289

PR #295, opened 2018-09-27
feat(menu-surface): add component

### LINKED ISSUE #188: Add menu-surface

moog16:
fixes #188


packages/menu-surface/README.md
```diff
@@ -0,0 +1,149 @@
+# React Menu Surface
```
moog16: add a note about always being hoisted

packages/menu-surface/index.js
```diff
@@ -0,0 +1,294 @@
...
+    }
+  }
+
+  componentWillUnmount() {
+    this.deregisterWindowClickListener_();
+    this.foundation_.destroy();
+  }
+
+  hoistToBody() {
+    // this deviatees from the mdc web version.
```
bonniezhou: "deviates"

packages/menu-surface/index.js
```diff
@@ -0,0 +1,294 @@
...
+    // here we force the menu to hoist, and require either
+    // this.props.(x,y) or this.props.anchorElement.
+    if (!(this.menuSurfaceElement_ && this.menuSurfaceElement_.current)) return;
```
bonniezhou: I think we can remove the null check for `this.menuSurfaceElement_`

test/screenshot/menu-surface/index.js
```diff
@@ -0,0 +1,110 @@
...
+    if (this.props.contextmenu) { // eslint-disable-line react/prop-types
+      window.addEventListener('contextmenu', this.rightClickCallback_);
+    }
+  }
+
+  componentWillUnmount() {
+    window.removeEventListener('contextmenu', this.rightClickCallback_);
+  }
+
+  setAnchorElement = (element) => {
```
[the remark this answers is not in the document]
moog16: I want the component to re-render when the ref is set. I put it on state, so that it renders again once it sets. Otherwise I did a forceRender(), which isn't as obvious.

packages/menu-surface/index.js
moog16: I think its safer to keep it in. `adapter.restoreFocus` is called within `foundation.close()`. `foundation.close` is called in `handleBodyClick` and `handleKeydown`, which are async events. There is the possibility that the MenuSurface is unmounted before a keydown event or a click event occurs, which would cause an error. 

What do you think?

moog16:
I think its the same I attached screenshots of web and react both set to BOTTOM_END

screen shot 2018-10-01 at 16 45 51
screen shot 2018-10-01 at 16 45 55

test/unit/menu-surface/index.test.js
```diff
@@ -0,0 +1,407 @@
...
+});
+
+test('update to props.open sets firstFocusableElement_', () => {
+  const wrapper = mount(<MenuSurface><button>click</button></MenuSurface>);
+  wrapper.setProps({open: true});
+  assert.equal(wrapper.instance().firstFocusableElement_, wrapper.find('button').getDOMNode());
+  removeMenuFromBody(wrapper);
+});
+
+test('update to props.open sets firstFocusableElement_', () => {
```
bonniezhou: "lastFocusableElement_"

bonniezhou:
Oh one thing I noticed was that the menu surface doesn't trap focus like the MDC Web one. We can either fix that in this PR or open a new bug for it

moog16:
@bonniezhou I don't think menu-surface traps focus. Menu does - is that what you meant?

bonniezhou:
Oh it might have been menu that I was looking at. Disregard! :)

Issue #333, opened 2018-10-11
Text field should call foundation.setValue only on blur


moog16:
Validation should only occur onBlur, which is what foundation.setValue does. The current issue is once the user starts to type, we start validating.

PR #343, opened 2018-10-17
chore(text-field): add screenshot tests and split into variants

moog16:
partial fix for https://github.com/material-components/material-components-web-react/issues/333

Split the screenshots for text field into different variant pages. Resemble was having a difficult time comparing the large image. Now there is the standard, textArea, fullWidth, and outlined variant pages. Added more text fields to screenshot tests to better support validation.

New fields added:

screen shot 2018-10-15 at 14 14 00


test/screenshot/text-field/fullWidth.js
```diff
@@ -0,0 +1,58 @@
...
+  if (!hasHelperText) {
+    return '';
+  } else if (isValidationMsg) {
+    return 'validation';
+  } else {
+    return 'persistent';
+  }
+};
+
+const textFields = iconsMap.map((icon) => {
```
bonniezhou: Hmm this is a lot of redundant code for every variant. Can we instead create a new component called `TextFieldPermutations` to dump all this code in? It can go in the TestTextField file, so it'd look like
```
class TestField extends React.Component {
  ...
}

class TextFieldPermutations extends React.Component {
  getHelperKeyText = (helperText) => {
    ...
  }

  render() {
    return (iconsMap.map((icon) => {
      return denseMap.map((dense) => {
        ...
        const props = Object.assign({}, this.props, icon, ...);
        ...
        return <TestField {...props} key={key} id={key} />;
      });
    );
  }
}
```

And then in this file we'd just have
```
const TextFieldScreenshotTest = () => (
  <div className='text-field-container'>
    <TextFieldPermutations fullWidth /  >
  </div>
);
```

test/screenshot/text-field/TextFieldPermutations.js
```diff
@@ -0,0 +1,53 @@
...
+  return hasHelperText && helperTextMap.helperText.props.validation;
+};
+
+const textFields = (variant) => {
+  return iconsMap.map((icon) => {
+    return denseMap.map((dense) => {
+      return rtlMap.map((isRtl) => {
+        return requiredMap.map((isRequired) => {
+          return disabledMap.map((disabled) => {
+            return helperTextMap.map((helperText, helperTextIndex) => {
```
bonniezhou: `helperTextIndex` is unused

moog16:
I think required captures minLength. I think `pattern` would be a better one to do.

bonniezhou:
Ok, just capturing it with `required` sounds fine to me

Issue #348, opened 2018-10-17
Top App Bar should supply a top-app-bar--fixed-adjust component


moog16:
By adding the fixedAdjust component, we can remove the responsibility of these details from the user and just add this component into the documentation.

PR #349, opened 2018-10-17
fix(top-app-bar): add fixed adjust component

moog16:
fixes https://github.com/material-components/material-components-web-react/issues/348


moog16:
definitely!

Issue #303, opened 2018-10-02
text field input's autoFocus causes error


moog16:
<img width="1020" alt="screen shot 2018-10-02 at 13 27 31" src="https://user-images.githubusercontent.com/579873/46374986-0a256280-c647-11e8-8f1f-177182943dfb.png">

steps to reproduce:

1. create <TextField>
1. add `autoFocus` prop to <Input autoFocus ... />

result: error shown in above screenshot

moog16:
Issue https://github.com/material-components/material-components-web-react/issues/310 will be fixing that. There is currently no clean way to focus.

PR #350, opened 2018-10-17
fix(text-field): remove foundation.setValue call during onChange

moog16:
fixes https://github.com/material-components/material-components-web-react/issues/333

By removing the componentDidUpdate from the index.js file, we no longer call foundation.setValue every input change. MDC Web was not doing this, so this is the correct fix.


packages/text-field/index.js
```diff
@@ -262,7 +263,7 @@ class TextField extends React.Component {
         key='text-field-container'
       >
         {leadingIcon ? this.renderIcon(leadingIcon) : null}
-        {this.renderInput()}
+        {this.foundation_ ? this.renderInput() : null}
```
moog16: This actually fixes https://github.com/material-components/material-components-web-react/issues/303

packages/text-field/Input.js
```diff
@@ -43,20 +40,44 @@ export default class Input extends React.Component {
...
 
-    if (this.props.id !== prevProps.id) {
-      this.props.setInputId(this.props.id);
+    if (id !== prevProps.id) {
+      setInputId(id);
+    }
+
+    // this should be in the componentDidMount, but foundation is not created
+    // at that time.
+    if (value && foundation !== prevProps.foundation) {
```
moog16: Once this https://github.com/material-components/material-components-web-react/pull/353 is fixed, I will move this back into componentDidMount

packages/text-field/Input.js
```diff
@@ -27,13 +27,10 @@ import classnames from 'classnames';
 import {VALIDATION_ATTR_WHITELIST} from '@material/textfield/constants';
 
 export default class Input extends React.Component {
-  constructor(props) {
-    super(props);
-    this.inputElement = React.createRef();
-  }
+  inputElement = React.createRef();
+  state = {disableProgramaticSetValue: false};
```
bonniezhou: Change this to `enableProgrammaticSetValue`, it's easier to reason about logic if the variable is positive

packages/text-field/Input.js
```diff
@@ -96,7 +119,11 @@ export default class Input extends React.Component {
   // value of the input is.
   handleChange = (e) => {
     const {foundation, onChange} = this.props;
+    // autoCompleteFocus runs on `input` event in MDC Web. In React, onChange and
+    // onInput are the same event
+    // https://stackoverflow.com/questions/38256332/in-react-whats-the-difference-between-onchange-and-oninput
     foundation.autoCompleteFocus();
+    this.setState({disableProgramaticSetValue: true});
```
bonniezhou: So the `onChange` event is only emitted when it's from a user? Do programmatic changes emit any event?

packages/text-field/Input.js
```diff
@@ -43,20 +40,46 @@ export default class Input extends React.Component {
...
-    if (this.props.value !== prevProps.value) {
-      this.props.handleValueChange(this.props.value);
+    if (value !== prevProps.value) {
+      handleValueChange(value, () => {
+        // only call #foundation.setValue on programatic changes;
+        // not changes by the user.
+        if (!this.state.disableProgramaticSetValue) {
+          foundation.setValue(value);
+        } else {
+          this.setState({disableProgramaticSetValue: false});
```
bonniezhou: at what point does the Text Field's `state.value` get updated if the change is from a user? should we be calling `handleValueChange` in `onBlur`?

packages/text-field/Input.js
moog16: Right the onChange event is only triggered when a user types. BOTH events (user and programatic) trigger updates props, which calls `componentDidUpdate`.

packages/text-field/Input.js
moog16: we call setValue in `handleBlur`, but its called within MDC's `foundation.deactivateFocus`

moog16:
update - ready for re-review

packages/text-field/Input.js
```diff
@@ -43,20 +40,46 @@ export default class Input extends React.Component {
...
-    if (this.props.id !== prevProps.id) {
-      this.props.setInputId(this.props.id);
+    if (id !== prevProps.id) {
+      setInputId(id);
+    }
+
+    // this should be in the componentDidMount, but foundation is not created
+    // at that time.
+    // Will fix this in
+    // https://github.com/material-components/material-components-web-react/pull/350/files
```
bonniezhou: should be 353, not 350

PR #353, opened 2018-10-18
fix(text-field): put foundation on state and do render input unless foundation is present

moog16:
fixes https://github.com/material-components/material-components-web-react/issues/303


moog16:
Well the inner components of a component will be initialized first. In this case, text field initializes <Input> before it considers itself "mounted". So the <Input>'s componentDidMount is called before text field's componentDidMount. <Input> relies on the foundation existing, which only happens after the text field's componentDidMount. 

So by adding foundation to state, we can guarantee that the <Input> will always have a foundation.

moog16:
Test passing, but CLA is not passing because I accidentally merged @TroyTae's commits into this branch. I reverted, so will just merge into master.

PR #357, opened 2018-10-20
Corrected imports of "MaterialIcon" component

urkopineda:
The package is `@material/react-material-icon` not `@material/material-icon`


moog16:
Thanks for fixing! Merging since its just a doc change

PR #361, opened 2018-10-23
package(top-app-bar): permit otherProps

TroyTae:
I can see most of components permit otherProps.
If there is any reason to permit otherProps.
I want to add and use it.
Thank you for read :)


TroyTae:
@googlebot 
I signed it!

PR #362, opened 2018-10-23
package(top-app-bar): permit otherProps

TroyTae:
I can see most of components permit otherProps.
If there is any special reason I want permit otherProps.
Thank you for read :)


TroyTae:
@bonniezhou 
https://travis-ci.com/material-components/material-components-web-react/jobs/153567120
I don't know about this error.
Can you help me?

PR #369, opened 2018-10-24
fix(infrastructure): update screenshot tests

bonniezhou:


bonniezhou:
nvm it was a node issue....updated node to v10.12.0 to match Travis CI

Issue #356, opened 2018-10-19
Ability to override text field validation


oste:
The text field input validation is currently set to `this.inputElement.current.validity.valid` as seen in the link below. It would be useful to support a property to override this for cases when you don't want to show the red invalidation state right away. For example, not showing validation messages until the user submits a form.

https://github.com/material-components/material-components-web-react/blob/377f2501d947f03fddfe7430ff9b6778fd2d3eb2/packages/text-field/Input.js#L120

PR #374, opened 2018-10-25
fix(text-field): allow for input isValid override

moog16:
fixes https://github.com/material-components/material-components-web-react/issues/356


packages/text-field/Input.js
```diff
@@ -117,7 +117,12 @@ export default class Input extends React.Component {
   }
 
   isBadInput = () => this.inputElement.current.validity.badInput;
-  isValid = () => this.inputElement.current.validity.valid;
+  isValid = () => {
+    if (typeof this.props.isValid === 'boolean') {
```
bonniezhou: We're just checking to see if this prop exists, right?
```
if (this.props.isValid !== undefined)
```

bonniezhou: On second thought I'm not sure which way of putting it is less awkward, up to you

moog16: I did `typeof this.props.isValid === 'boolean'`, because we want it to be `true` or `false`. If its anything else (null, undefined, 12, 'abc', etc.) then it shouldn't be considered.

moog16: actually...I guess they would get a warning via PropTypes. I will change to your suggestion.

moog16:
Created https://github.com/material-components/material-components-web/issues/4054 which will include another follow up PR once completed.

Issue #253, opened 2018-08-29
Switch PropType.String --> PropType.node


moog16:
PropType.String is a little restrictive. Some examples where we would want to allow for PropType.node:

```
<Foo text={<Theme {...}>Hello</Theme>} />
<Foo text={<I18n key='something' />} />
<Foo text={<FormatedNumber value={123} /> />
```

This specific issue is concerning Top App Bar. The feature request is being able to put an image into the Top App Bar’s title. The developer specifically referenced the Fortnightly Material study.

moog16:
@bmihelac is there a font-family for the Fortnightly study?

bmihelac:
I cannot find one, but from description I guess it is to expect that logo is image:

> Fortnightly’s top app bar contains a navigation icon, logo, search capability, and chips. The chips present trending hashtags.

moog16:
I'm going to ask the designers, just wondering if you had found anything.

bmihelac:
Even material.io has an image :)

![screen shot 2018-08-30 at 20 10 39](https://user-images.githubusercontent.com/13813/44870615-f75aef00-ac90-11e8-8fc9-138d1b578616.png)

sashless:
The Input Label for MDCTextField is also requiring `text`. Would be nice to allow `PropType.node` as well there.

PR #376, opened 2018-10-25
fix(top-app-bar): allow react elements in title

moog16:
fixes https://github.com/material-components/material-components-web-react/issues/253


moog16:
created https://github.com/material-components/material-components-web-react/issues/388

PR #387, opened 2018-10-30
fix(list): selectedIndex 0 isn't working

TroyTae:
As you know, 0 is replaced with false in javascript.
When selectedIndex is 0, setSelectedIndex is not executed.
So I add check code "Number.isInteger".

### DIFF
--- a/packages/list/index.js
+++ b/packages/list/index.js
@@ -50,7 +50,7 @@ export default class List extends Component {
     this.foundation_.setSingleSelection(singleSelection);
     this.foundation_.setWrapFocus(wrapFocus);
     this.foundation_.setVerticalOrientation(this.props[ARIA_ORIENTATION] === VERTICAL);
-    if (selectedIndex) {
+    if (Number.isInteger(selectedIndex)) {
       this.foundation_.setSelectedIndex(selectedIndex);
     } else {
       const id = this.getListItemIdFromIndex_(0);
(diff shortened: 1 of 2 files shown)


TroyTae:
I signed it!

packages/list/index.js
```diff
@@ -50,7 +50,7 @@ export default class List extends Component {
     this.foundation_.setSingleSelection(singleSelection);
     this.foundation_.setWrapFocus(wrapFocus);
     this.foundation_.setVerticalOrientation(this.props[ARIA_ORIENTATION] === VERTICAL);
-    if (selectedIndex) {
+    if (Number.isInteger(selectedIndex)) {
```
moog16: `isInteger` is not supported by IE11, which is something we need to support. Try using the `typeof selectedNumber === 'number' && isNaN(selectedNumber)` check. I believe `typeof NaN` returns 'number', which is why we need the last part of that statement. 

Also please add a few unit tests to cover the case you mentioned in your first comment of this PR. 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isInteger

TroyTae: Oh I mistook it for typescript.
I will change it and write some test code!

TroyTae:
@moog16 
Thank you for your verification.

moog16:
Opened PR https://github.com/material-components/material-components-web-react/pull/406 for tests

moog16:
tests passing...merging!

Issue #326, opened 2018-10-08
Investigate CRA 2.0


moog16:
- potentially need to update docs
- try test app using CRA 2.0 beta

PR #393, opened 2018-10-31
docs: update main README with CRA2 guidelines

moog16:
fixes https://github.com/material-components/material-components-web-react/issues/326 and https://github.com/material-components/material-components-web/issues/3985

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,7 +113,7 @@ If you permanently want to add this to your environment, read [adding environmen
 ```sass
 // ./react-button-overrides.scss
 
-@import "@material/button/mixins.scss";
+@import "@material/button/mixins";
 
 .button-alternate {
   @include mdc-button-container-fill-color(lightblue);
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ Component | Spec | MDC Web
 
 `create-react-app` is a popular CLI tool to getting started with React. If you're new to React or Webpack, you might be starting out here. This section will walk you through configuring `create-react-app` to install and use our components.
 
-If you're using an older version (v1) of `create-react-app`, please refer to our [create-react-app-v1 doc](../docs/create-react-app-v1.md).
+If you're using an older version (v1) of `create-react-app`, please refer to our [create-react-app-v1 doc](docs/create-react-app-v1.md).
 
 >  Recommended things to know
 
@@ -75,11 +75,11 @@ npm install --save @material/react-button
 
 If you want to use the compiled CSS and not customize any colors, text, etc. you can skip to [Step 3a](#step-3a-use-compiled-css).
 
-Most likely you'll want to start using the [Sass mixins](https://github.com/material-components/material-components-web/blob/master/docs/code/architecture.md#sass) to customize your app. MDC Sass files are not supported out of the box, since we do not append `~` to our module imports. See this [Github issue](https://github.com/facebook/create-react-app/issues/4494#issuecomment-428531848) explaining the issue in detail. There is a workaround, but requires some work on your end (we promise it is not too difficult).
+Most likely you'll want to start using the [Sass mixins](https://github.com/material-components/material-components-web/blob/master/docs/code/architecture.md#sass) to customize your app. MDC Sass files are not supported out of the box, since we do not prepend `~` to our module imports. See this [Github issue](https://github.com/facebook/create-react-app/issues/4494#issuecomment-428531848) explaining the issue in detail. There is a workaround, but requires some work on your end (we promise it is not too difficult).
 
 ##### Add environment variable
 
-To get MDC React Components to work with `create-react-app` you need to set a `SASS_PATH` environment variable that points to your `node_modules` directory. To quickly do this on OSx or Linux enter the following in your command line:
+To get MDC React Components to work with `create-react-app` you need to set a `SASS_PATH` environment variable that points to your `node_modules` directory. To quickly do this on OS X or Linux enter the following in your command line:
 
 ```sh
 export SASS_PATH=./node_modules
@@ -88,7 +88,7 @@ export SASS_PATH=./node_modules
 If you're on Windows use the following:
 
 ```bat
-SET SASS_PATH=./node_modules
+SET SASS_PATH=.\node_modules
 ```
 
 If you permanently want to add this to your environment, read [adding environment variables](./docs/adding-env-variables.md). You're now ready to start using MDC React Sass modules in your `create-react-app`.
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -40,6 +40,8 @@ Component | Spec | MDC Web
 
 `create-react-app` is a popular CLI tool to getting started with React. If you're new to React or Webpack, you might be starting out here. This section will walk you through configuring `create-react-app` to install and use our components.
 
+If you're using an older version (v1) of `create-react-app`, please refer to our [create-react-app-v1 doc](../docs/create-react-app-v1.md).
```
kfranqueiro: Remove `../` from url

README.md
```diff
@@ -49,18 +51,17 @@ Component | Spec | MDC Web
...
 
-
 #### Step 1: Install create-react-app
 
 Install `create-react-app`:
 
 ```
-npm i -g create-react-app
-create-react-app my-app
+npx create-react-app my-app
```
kfranqueiro: This removed the step that actually installs create-react-app. I assume based on this it's intended to be installed locally instead of globally now?

README.md
```diff
@@ -49,18 +51,17 @@ Component | Spec | MDC Web
...
 cd my-app
+npm start
 ```
 
-> NOTE: all npm commands can be replaced with yarn
+> _NOTE:_ all npm commands can be replaced with yarn
```
kfranqueiro: Does yarn have an equivalent to `npx`?

README.md
```diff
@@ -74,41 +75,25 @@ npm install --save @material/react-button
...
 
-Most likely you'll want to start using the [Sass mixins](https://github.com/material-components/material-components-web/blob/master/docs/code/architecture.md#sass) to customize your app. There are a few ways to achieve this. `create-react-app` does have a [recommended approach](https://github.com/facebook/create-react-app/blob/master/packages/react-scripts/template/README.md#adding-a-css-preprocessor-sass-less-etc), which we also recommend.
-
-The following is an alternate version of the `create-react-app` approach. The difference being all the `node_modules` imports will go into `./src/App.scss`. First install `node-sass-chokidar`:
+Most likely you'll want to start using the [Sass mixins](https://github.com/material-components/material-components-web/blob/master/docs/code/architecture.md#sass) to customize your app. MDC Sass files are not supported out of the box, since we do not append `~` to our module imports. See this [Github issue](https://github.com/facebook/create-react-app/issues/4494#issuecomment-428531848) explaining the issue in detail. There is a workaround, but requires some work on your end (we promise it is not too difficult).
```
kfranqueiro: append -> prepend

README.md
```diff
@@ -74,41 +75,25 @@ npm install --save @material/react-button
...
-  "watch-css": "npm run build-css && node-sass-chokidar --include-path ./src --include-path ./node_modules --watch ./src/App.scss ./src/App.css",
-  "start-js": "react-scripts start",
-  "start": "npm-run-all -p watch-css start-js",
-  "build-js": "react-scripts build",
-  "build": "npm-run-all build-css build-js",
-  "test": "react-scripts test --env=jsdom",
-  "eject": "react-scripts eject"
-}
+```bat
+SET SASS_PATH=./node_modules
```
kfranqueiro: `/` is not the file separator on windows. have we tested on windows?

README.md
```diff
@@ -74,41 +75,25 @@ npm install --save @material/react-button
...
-The following is an alternate version of the `create-react-app` approach. The difference being all the `node_modules` imports will go into `./src/App.scss`. First install `node-sass-chokidar`:
+Most likely you'll want to start using the [Sass mixins](https://github.com/material-components/material-components-web/blob/master/docs/code/architecture.md#sass) to customize your app. MDC Sass files are not supported out of the box, since we do not append `~` to our module imports. See this [Github issue](https://github.com/facebook/create-react-app/issues/4494#issuecomment-428531848) explaining the issue in detail. There is a workaround, but requires some work on your end (we promise it is not too difficult).
 
-```
-npm install -D node-sass-chokidar npm-run-all
-```
+##### Add environment variable
 
-In `package.json` replace the following:
+To get MDC React Components to work with `create-react-app` you need to set a `SASS_PATH` environment variable that points to your `node_modules` directory. To quickly do this on OSx or Linux enter the following in your command line:
```
kfranqueiro: This is going to sound like a severe nit, but "OSx" -> "OS X" here and other places it's spelled this way

docs/adding-env-variables.md
```diff
@@ -0,0 +1,73 @@
...
+
+> _NOTE:_ Be sure to close that terminal and open a new one for the new settings to take effect.
+
+## Mac OSx
+
+#### Open your `bash_profile`
+You can add an environment variable to your system by updating your `~/.bash_profile`. Open a terminal window and follow these steps:
+
+```sh
+touch ~/.bash_profile # run this line only if the file doesn't exist
```
kfranqueiro: There's not really any reason to do this since the next command will create it if it doesn't exist anyway

docs/adding-env-variables.md
```diff
@@ -0,0 +1,73 @@
...
+Save and exit from the file.
+
+#### Source your updated `bash_profile`
+
+Open new terminal window for these settings to take effect.
+
+If you want to keep the same terminal, source your `bash_profile` with the following:
+
+```sh
+. ~/.bash_profile
```
kfranqueiro: If we want to be super-safe, replace `.` with `source` for compatibility with shells other than bash

docs/adding-env-variables.md
```diff
@@ -0,0 +1,73 @@
...
+```
+
+## Linux
+
+Linux is pretty similar to Mac OSx if you're familiar with that operating system. The steps are identical, with the exception that your environment configuration lives in the `~/.bashrc` file.
+
+
+
+You can add an environment variable to your system by updating your `~/.bash_profile` (on OSx) or `~/.bash_rc` (on Linux). If you're on Windows please read
```
kfranqueiro: There is no underscore in `.bashrc`

README.md
moog16: from what i read, `npx` installs a package globally. It is also what is on the create-react-app github page.


https://www.npmjs.com/package/npx

```
Executes <command> either from a local node_modules/.bin, or from a central cache, installing any packages needed in order for <command> to run.```

README.md
moog16: not completely equivalent, but they do have `yarn create`

https://yarnpkg.com/lang/en/docs/cli/create/

Want to document this too?

README.md
moog16: yup tested on windows...good catch thanks :)

README.md
moog16: I always thought it was OSx...like the opposite of iPhone.

docs/adding-env-variables.md
moog16: true...is that also true of `nano`?

PR #404, opened 2018-11-01
fix(text-field): uncomment test fixed from 0.35.0

moog16:


moog16:
I think pre-v0.35.0 the logic wasn't there to support it. I can't remember :|

PR #405, opened 2018-11-02
feat(list): allow semantic tags (mainly, nav on List and a on ListItem)

### LINKED ISSUE #382: Modal variant of drawer crashes React over uncaught exception in FocusTrap
Hi! 👋

# TL;DR
Using `<Drawer modal ... />` causes uncaught exception in `<FocusTrap />` upon clicking any button that activates the drawer, which in turn crashes React.

## Reproduction repo
* https://github.com/4cm4k1/mdc-react-modal-drawer-troubleshooting

## Environment info (just in case)
* macOS v10.14.1
* Node v11
* `yarn` v1.10.1
* `react`/`react-dom` v16.6.0
* `next` v7.0.2
  * `webpack` v4.20.2
* `@material/react-drawer` v0.6.1

## Overview
If I am not mistaken, there seems to be a bug in the implementation of `focus-trap-react` in`@material/react-drawer`. Activating the modal variant of the drawer causes `react` to crash and attempt to recreate the component tree unsuccessfully, with the below console output. I have reproduced this both in this example repository and my [personal website's repo](https://github.com/4cm4k1/website) in development and production modes.

I have done my best in following the documentation, which currently isn't the greatest and has several typographical errors in variable/function naming. I've also tried troubleshooting this myself, but without much success. Notably, by setting manually `focusTrapOptions` values all to `true`, I prevented the crash, but it then results in Chrome killing a call stack that maxes out, which hints at a looping issue. Then, I looked at `focus-trap`'s docs and found that `initialFocus` should not be set as a Boolean [as it is here](https://github.com/material-components/material-components-web-react/blob/master/packages/drawer/index.js#L155).

However, I'm finding the layers of abstraction a little bit hard to follow, so that's why I'm raising the issue here. I don't know how to solve this. :P

I greatly appreciate the team's assistance here, and I've been very excited to see MDC making such big strides on multiple platforms! Thanks!

### Dev mode

```shell
Uncaught Error: You can't have a focus-trap without at least one focusable element
    at getInitialFocusNode (VM1105 index.js:214)
    at updateTabbableNodes (VM1105 index.js:222)
    at Object.activate (VM1105 index.js:211)
    at FocusTrap.componentDidMount (VM1105 index.js:906)
    at commitLifeCycles (VM1117 react-dom.development.js:15255)
    at commitAllLifeCycles (VM1117 react-dom.development.js:16523)
    at HTMLUnknownElement.callCallback (VM1117 react-dom.development.js:149)
    at Object.invokeGuardedCallbackDev (VM1117 react-dom.development.js:199)
    at invokeGuardedCallback (VM1117 react-dom.development.js:256)
    at commitRoot (VM1117 react-dom.development.js:16677)
    at completeRoot (VM1117 react-dom.development.js:18069)
    at performWorkOnRoot (VM1117 react-dom.development.js:17997)
    at performWork (VM1117 react-dom.development.js:17901)
    at performSyncWork (VM1117 react-dom.development.js:17873)
    at batchedUpdates$1 (VM1117 react-dom.development.js:18108)
    at batchedUpdates (VM1117 react-dom.development.js:2198)
    at dispatchEvent (VM1117 react-dom.development.js:4939)
```

```shell
The above error occurred in the <FocusTrap> component:
    in FocusTrap (created by Drawer)
    in aside (created by Drawer)
    in Drawer (at pages/index.js:20)
    in Home (created by App)
    in Container (created by App)
    in App

React will try to recreate this component tree from scratch using the error boundary you provided, App.
```

### Prod mode

```shell
Error: You can't have a focus-trap without at least one focusable element
    at h (index.js:58)
    at E (index.js:58)
    at Object.activate (index.js:58)
    at t.value (index.js:308)
    at Ia (commons.2a89c77150ec82f381ee.js:22)
    at ja (commons.2a89c77150ec82f381ee.js:22)
    at Oa (commons.2a89c77150ec82f381ee.js:22)
    at Aa (commons.2a89c77150ec82f381ee.js:22)
    at Ue (commons.2a89c77150ec82f381ee.js:22)
    at Nn (commons.2a89c77150ec82f381ee.js:22)
```

4cm4k1:
### Backstory
This PR began as an attempt to help with the focus-trap issues that using `Drawer` and `List` together were causing (reported here: https://github.com/material-components/material-components-web-react/issues/382; tracked here: https://github.com/material-components/material-components-web-react/issues/433).

### Purpose
What this PR now is simply adding the ability to override the default tags (`nav` [default: `ul`] on `List` and `a` [default: `li`] on `ListItem`) by using the `tag` prop.

### Context
Changing the tags would be recommended when using these components to render a page's navigation, as it is semantic HTML and helps with a11y. Additionally, this may help with compatibility with routing libraries such as [React Router](https://github.com/ReactTraining/react-router) and [Next.js' Link](https://github.com/zeit/next.js#with-link).

#### Notes
I've also added two unit tests corresponding to the functionality added by this PR and added text regarding this feature in `List`'s `README.md`.


packages/list/ListItem.js
```diff
@@ -72,14 +72,16 @@ export default class ListItem extends Component {
       ...otherProps
     } = this.props;
 
+    const SemanticListItem = this.props.href ? 'a' : 'li';
```
moog16: I do agree with this change, but I think there should be a way to programatically change routes. Things like React Router and SPA architecture require this. Semantically we will want an anchor tag, but no href, correct?

4cm4k1:
@moog16 Thank you for pointing that out. `ListItemText`, `ListItemGraphic`, and `ListItemMeta` have `tabIndex` attributes currently, while `ListItem` does not. What would be the best way, then, to implement it for `ListItem`? I guess I'm a bit confused as to why the children of `ListItem` have `tabIndex` but not `ListItem`.

I updated the readme with the `tag` prop name addition to `List`.

packages/list/ListItem.js
4cm4k1: I think so? I'm having a hard time with the layers of abstraction here. Below is an example of how I would imagine it working while using [`next`](https://github.com/zeit/next.js) and its routing logic via `next/link`.

```jsx
import Link from 'next/link';
import List, { 
  ListItem,
  ListItemGraphic,
  ListItemText,
} from '@material/react-list';

const NavList = [
  {
    href: '/',
    name: 'Home',
    icon: 'home',
  }, {
    href: '/info',
    name: 'Info',
    icon: 'info',
  }
];

export default () => (
  <List wrapFocus tag="nav">
    {NavList.map(({ href, name, icon }, index) => (
      <Link key={index} href={href} prefetch passHref>
        <ListItem key={index} tag="a">
          <ListItemGraphic graphic={<MaterialIcon icon={icon} />} />
          <ListItemText primaryText={name} />
        </ListItem>
      </Link>
      // With react-router-dom it would be pretty similar...
      // <Link to={href}>
      //   <ListItem ...>
      //     <ListItemGraphic ... />
      //     <ListItemText ... />
      //   </ListItem>
      // </Link>
    ))}
  </List>
);
```

moog16: What if people wanted to do 
```
<ListItem key={index} tag="a" onClick={this.goToAnotherRoute}>
          <ListItemGraphic graphic={<MaterialIcon icon={icon} />} />
          <ListItemText primaryText={name} />
</ListItem>
```

All I'm saying is, we should remove `const SemanticListItem = this.props.href ? 'a' : 'li';`, in favor of the `Tag` property.

moog16:
@4cm4k1 tabIndex is rather complicated, but necessary for a11y. WCAG states we should be able to tab to each separate <List>, but arrow through each <ListItem>. We also must add tabindex=-1 to all child elements, since some of them might have tabIndex.

I haven't had time this week to check, but you're saying using <List> with tabIndex does not solve modal issue of crashing?

4cm4k1:
> @4cm4k1 tabIndex is rather complicated, but necessary for a11y. WCAG states we should be able to tab to each separate , but arrow through each . We also must add tabindex=-1 to all child elements, since some of them might have tabIndex.
> 
> I haven't had time this week to check, but you're saying using with tabIndex does not solve modal issue of crashing?

Correct. The crash issue only seemed to go away after using `<a>` tags instead of `<li>`. I totally understand and support the need for a11y; I’m just admittedly not aware of the technical details.

packages/list/ListItem.js
4cm4k1: > What if people wanted to do
> 
> ```
> <ListItem key={index} tag="a" onClick={this.goToAnotherRoute}>
>           <ListItemGraphic graphic={<MaterialIcon icon={icon} />} />
>           <ListItemText primaryText={name} />
> </ListItem>
> ```
> All I'm saying is, we should remove `const SemanticListItem = this.props.href ? 'a' : 'li';`, in favor of the `Tag` property.

Got it. I've updated `ListItem` accordingly and updated the README.

Another thing to consider: When using router frameworks that nest elements within `<Link></Link>`, I encountered problems where the error output is something to the effect of `Cannot set property 'aria-selected' of undefined`, and when using the `selectedIndex` prop, it’s unable to set it on the index specified. Would this be fixed if users extended the `Link` component to pass the necessary props to its children (i.e., `ListItem`)? Or how else can we make it safe to nest `ListItem`?

I have the beginnings of that in [my project](https://github.com/4cm4k1/website/blob/master/components/Link.jsx), though currently using RMWC because of the crashing issues here, shown here:

```jsx
<Link {...props}>
  <ListItem tag="a" activated={activated}>
    {children}
  </ListItem>
</Link>
```

moog16:
@4cm4k1 could you please fix conflicts? I'll do another review after that is done. Thanks!

moog16:
@4cm4k1 opened this issue to solve the modal crashing issue https://github.com/material-components/material-components-web-react/issues/433

4cm4k1:
@4cm4k1 I resolved the merge conflicts, added some text to the `README.md`, renamed this PR, and rewrote the PR's description to better reflect what this is. Thanks! 😄

test/unit/list/ListItem.test.js
```diff
@@ -65,3 +65,8 @@ test('passes props.childrenTabIndex to children as props.tabIndex', () => {
   );
   assert.equal(wrapper.find('.list-item-child').props().tabIndex, 2);
 });
+
```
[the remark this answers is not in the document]
4cm4k1: Can you clarify how to get set up locally? I've been unable to successfully `npm run build` or `npm run test:unit`, which has made it difficult to validate these tests myself, and there doesn't seem to be any documentation on 'developing' the repo. For now, I've added the default cases and corrected a mistake in the list item `a` test. Thanks!

moog16:
created a tests pr https://github.com/material-components/material-components-web-react/pull/439

test/unit/list/ListItem.test.js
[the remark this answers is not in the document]
moog16: Sounds like you've got it figured out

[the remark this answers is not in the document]
moog16: Added a task to add a doc for contributions. Thanks! https://github.com/material-components/material-components-web-react/issues/440

moog16:
@4cm4k1 tests pass! Please sign it by commenting "I signed it" in this pr

4cm4k1:
I signed it

moog16:
Thanks @4cm4k1 !!

PR #412, opened 2018-11-05
docs: menu surface open state & ref errors

dan-kwiat:
The `open` prop doesn't control the open state of the menu surface (since clicking outside will always close the menu surface regardless of the `open` value).  Rather a change in the `open` prop will trigger an opening / closing of the menu surface.

### DIFF
--- a/packages/menu-surface/README.md
+++ b/packages/menu-surface/README.md
@@ -60,7 +60,7 @@ class MyApp extends React.Component {
           open={this.state.open}
           anchorCorner={Corner.BOTTOM_LEFT}
           onClose={() => this.setState({open: false})}
-          anchorElement={anchorElement}
+          anchorElement={this.state.anchorElement}
         >
           <img
             style={{maxWidth: '20vw', maxHeight: '20vh'}}
@@ -83,6 +83,7 @@ import MenuSurface from '@material/react-menu-surface';
 class MyApp extends React.Component {
   state = {
     open: false,
+    coordinates: null,
   };
 
   componentDidMount() {
@@ -109,7 +110,7 @@ class MyApp extends React.Component {
         <MenuSurface
           open={this.state.open}
           onClose={() => this.setState({open: false, coordinates: null})}
-          coordinates={coordinates}
+          coordinates={this.state.coordinates}
         >
           <img
             style={{maxWidth: '20vw', maxHeight: '20vh'}}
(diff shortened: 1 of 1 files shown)


packages/menu-surface/README.md
```diff
@@ -132,7 +133,7 @@ anchorCorner | Corner | Sets the corner that the menu surface will be anchored t
 anchorElement | Element | Sets the anchor element used as an anchor for `menu-surface` positioning logic. Should contain class `.mdc-menu-surface--anchor`.
 anchorMargin | Object* | Sets the distance from the anchor point that the menu surface should be shown.
 coordinates | {x: Number, y: Number} | Sets the anchor coordinates when menu surface does not use anchorElement.
-open | Boolean | Opens menu surface when true.
+open | Boolean | Changing value will trigger open/close of the menu surface.
```
moog16: revert this change.

dan-kwiat:
@moog16 thanks for the quick response.  I agree that changing the `open` prop from true to false will close the menu (and vice versa for opening).  However it's only this _change_ in prop value that can be used to control the state of the menu.  The value of the `open` prop doesn't always determine whether the menu is open or not since the foundation's `handleBodyClick` can also close it.  This is ok if you use `onClose` to keep track of those events but could be confusing for people who want to prevent the menu from closing e.g.

```jsx
const MyComponent = () => (
  <MenuSurface open={true}>
    I'll still close if you click outside
  </MenuSurface>
)
```

moog16:
@dan-kwiat so sorry - I read that change backwards for some reason. Your verbiage makes sense. 

RE your last note: Do you want to change the API? Are you saying we should handle it differently if you  don't want to implement the onClose method?

dan-kwiat:
@moog16 no worries!  I think it could make sense to have the menu fully controlled by avoiding the foundation's `handleBodyClick` & `handleKeydown` methods.  If that logic was in the `MenuSurface` component then `onClose` could just be called when clicking outside or hitting escape - leaving the user the choice of whether to update the `open` prop or not.  What do you think?

moog16:
@dan-kwiat - so the only way to open/close the surface is via the `prop.open`? By removing the keydown methods we decrease our a11y support. 

This React repo's main goals are to implement MDC Web's features. So the only way we could remove the handleBodyClick feature is to first go through MDC Web as an issue. 

I propose we accept this PR, and you can open an issue on MDC Web. How does that sound?

dan-kwiat:
@moog16 Another option would be to keep the keydown methods & just make the window click listener optional.

I think you're probably right though.  The purpose of this PR was just to clarify existing functionality.

Issue #366, opened 2018-10-23
Move list event handlers from the list on to the list item


moog16:
move `onKeyDown`, `onClick`, `onFocus`, and `onBlur` into the `renderListItem`, and pass the index. Then we can remove `getListItemIndexOfTarget_` within each event.

bonniezhou:
Closed by #413

Issue #365, opened 2018-10-23
Add support for adding/removing list items


moog16:
- add test for removing list item
- add test for adding list item
- add note in README
- add to screenshot tests

bonniezhou:
Closed by #413

PR #413, opened 2018-11-05
refactor(list): Refactor list item management

### LINKED ISSUE #367: [exploratory task] Reimplement list item management
Remove:

```
  listItems_ = {};
  listItemIndices_ = {};
  state = {
    listItemAttributes: {}, // maps id to {attr: val} map object
    listItemClassList: {}, // maps id to classList set
    listItemChildrenTabIndex: {}, // maps id to children's tabIndex
  };
```

Instead we can have methods to manage classes/attributes, focus, followHref, and togglingCheckbox on the listItem component. Having 2 "sources of truth" (the above and the rendered dom elements), is a bad coding pattern.

bonniezhou:
Move the logic having to do with list item classes, attributes, etc inside `ListItem` instead of in `List`. Move event handlers to `ListItem`. Remove the needs for IDs on list items, instead recalibrate the list item index on every re-render. 

Fixes #367, #366, #365.


packages/list/index.js
```diff
@@ -82,219 +85,188 @@ export default class List extends Component {
...
-
   get adapter() {
     return {
-      getListItemCount: () => Object.keys(this.listItems_).length,
-      getFocusedElementIndex: () => this.getListItemIndexOfTarget_(document.activeElement),
+      getListItemCount: () => this.listItemCount,
+      getFocusedElementIndex: () => {
+        let count = -1;
+        let index = -1;
+        React.Children.forEach(this.props.children, (child) => {
```
moog16: this is cool - didn't know you could inspect child.type.name. Aren't you give the index here? Is it because you're accounting for dividers and other elements that aren't ListItems?

packages/list/index.js
```diff
@@ -82,219 +85,188 @@ export default class List extends Component {
...
-      className;
-  }
-
   get adapter() {
     return {
-      getListItemCount: () => Object.keys(this.listItems_).length,
-      getFocusedElementIndex: () => this.getListItemIndexOfTarget_(document.activeElement),
+      getListItemCount: () => this.listItemCount,
+      getFocusedElementIndex: () => {
+        let count = -1;
```
moog16: I think count should start at 0

packages/list/index.js
```diff
@@ -82,219 +85,188 @@ export default class List extends Component {
...
-        const id = this.getListItemIdFromIndex_(index);
-        const listItem = this.listItems_[id];
-        if (listItem) {
-          listItem.focus();
+        const {listItemClassNames} = this.state;
+        if (!listItemClassNames[index]) {
+          return;
         }
+        const i = listItemClassNames[index].indexOf(className);
+        listItemClassNames[index].splice(i, 1);
```
moog16: I think you need to check if `i<= -1`. If you do `listItemClassNames[index].splice(-1, 1)`, its going to delete something, which I believe is unintended.

packages/list/index.js
```diff
@@ -82,219 +85,188 @@ export default class List extends Component {
...
-    let target = eventTarget;
-
-    // Find the first ancestor that is a list item.
-    while (this.getIndexOfListItemElement_(target) === undefined) {
-      if (!target || target === document) {
-        return -1;
-      }
-      target = target.parentElement;
+  renderChild = (child) => {
+    if (child.type.name === 'ListItem') {
```
moog16: I think in the Readme, we should say that the children need to be type ListItem's (if we don't already). One of the issues this past week gave an example of this:

```
<Link>
  <ListItem ... />
</Link>
```

packages/list/index.js
moog16: Also is there a way to break early from this? If you've found the element

packages/list/index.js
```diff
@@ -82,219 +85,188 @@ export default class List extends Component {
...
-      ref: (listItem) => {
-        this.listItems_[idOrIndex] = listItem;
-        this.listItemIndices_[idOrIndex] = index;
-      },
+    const props = {
       ...otherProps,
-    },
-    this.state.listItemAttributes[idOrIndex]);
+      onKeyDown: (e) => {
+        onKeyDown();
```
moog16: pass the events in onKeyDown, onClick, onFocus, and onBlur. The user's may want that

packages/list/index.js
```diff
@@ -82,219 +85,188 @@ export default class List extends Component {
...
+        this.handleFocus(e, index);
+      },
+      onBlur: (e) => {
+        onBlur();
+        this.handleBlur(e, index);
+      },
+      shouldFocus: focusListItemAtIndex === index,
+      shouldFollowHref: followHrefAtIndex === index,
+      shouldToggleCheckbox: toggleCheckboxAtIndex === index,
+      attributesToSet: listItemAttributes[index],
```
moog16: How about renaming `attributesToSet` and `classNamesToAdd` to `attributesFromList` and `classNamesFromList`? I think it is inferred that we want to add them.

packages/list/index.js
bonniezhou: yep, the index thing is to get around dividers and non-ListItems. and you can't break early in `React.Children.forEach` so I changed it to use `React.Children.toArray` instead. https://stackoverflow.com/questions/51334473/react-how-to-break-out-of-react-children-foreach-loop

packages/list/index.js
bonniezhou: omg yes, I struggled hard with this naming 😅

packages/list/index.js
```diff
@@ -84,219 +87,199 @@ export default class List extends Component {
...
   }
 
-  onKeyDown = (e) => {
-    this.props.onKeyDown(e);
+  handleKeyDown = (e, index) => {
     e.persist(); // Persist the synthetic event to access its `key`
-    const index = this.getListItemIndexOfTarget_(e.target);
     if (index >= 0) {
       this.foundation_.handleKeydown(e, true /* isRootListItem is true if index >= 0 */, index);
+      // Work around until MDC Web issue is resolved:
```
moog16: the `if(index >= 0)` can be inlined with `if (e.key === 'Enter' || e.keyCode === 13 || e.key === 'Space' || e.keyCode === 32)`. The `foundation.handleKeyDown` method already does that check.

packages/list/index.js
```diff
@@ -84,219 +87,199 @@ export default class List extends Component {
...
-  onClick = (e) => {
-    this.props.onClick(e);
-    const index = this.getListItemIndexOfTarget_(e.target);
+  handleClick = (e, index) => {
     // Toggle the checkbox only if it's not the target of the event, or the checkbox will have 2 change events.
     const toggleCheckbox = e.target.type === 'checkbox';
     this.foundation_.handleClick(index, toggleCheckbox);
+    // Work around until MDC Web issue is resolved:
+    // https://github.com/material-components/material-components-web/issues/4053
+    this.props.handleSelect(index);
```
moog16: this should have an `if (index >= 0)` check. `foundation.handleClick` does that check, so we should also do it here.

packages/list/index.js
```diff
@@ -84,219 +87,199 @@ export default class List extends Component {
...
-      ref: (listItem) => {
-        this.listItems_[idOrIndex] = listItem;
-        this.listItemIndices_[idOrIndex] = index;
-      },
+    const props = {
       ...otherProps,
-    },
-    this.state.listItemAttributes[idOrIndex]);
+      onKeyDown: (e) => {
+        onKeyDown(e);
```
moog16: Sorry - should have thought of this earlier...do you think the `index` is useful information to pass to the `onKeyDown` event handler?

bonniezhou: If we include `index` in `onKeyDown` then we should probably include it in the rest of the handlers too. I think we can leave it as is for now, and add it if someone has a specific use case for `index`.

bonniezhou:
Filed #415

Issue #337, opened 2018-10-16
Text field needs a way to access `inputElement`


moog16:
There might be a time when a developer needs to access `HTMLValidity` object. The easiest way to give that to the developer is access to the `inputElement` via a ref.

moog16:
related https://github.com/material-components/material-components-web-react/issues/310

PR #414, opened 2018-11-05
fix(text-field): added reference to input element

### LINKED ISSUE #310: Add inputRef for Input in TextField
Create a way to programmatically focus on input element.
refer to #308

moog16:
fixes #310 and #337

### DIFF
--- a/packages/text-field/README.md
+++ b/packages/text-field/README.md
@@ -94,6 +94,7 @@ setDisabled | Function | Callback function that is called when the `disabled` pr
 setInputId | Function | Callback function that is called when the `id` attribute updates.
 handleFocusChange | Function | Callback function that is called when `focus` or `blur` events occur
 value | Number/String | Value of the input.
+ref | Function(input: ReactElement) => void | On mount of component, will call passed function with the instance of the `<Input />`.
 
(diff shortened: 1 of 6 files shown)


packages/text-field/README.md
```diff
@@ -97,6 +97,46 @@ value | Number/String | Value of the input.
...
+
+  render() {
+    return (
+      <div>
+        <div>
+          <Button onClick={this.focusTextField}>Focus Text Field</Button>
+        </div>
+        <TextField
+          label='Dog'
+          ref={this.textFieldElement}
```
bonniezhou: Would it be less confusing to access the native input element through the `Input` instead of the `Textfield`?

```
class MyApp extends React.Component {
  input = React.createRef();
  state = {value: 'Woof'};

  focusTextField = () => {
    const inputElement = this.input.current.inputElement;
    if (inputElement) {
      inputElement.focus();
    }
  }

  render() {
    return (
      <div>
        <div>
          <Button onClick={this.focusTextField}>Focus Text Field</Button>
        </div>
        <TextField
          label='Dog'
        >
          <Input
            value={this.state.value}
            onChange={(e) => this.setState({value: e.target.value})}
            ref={this.input}/>
        </TextField>
      </div>
    );
  }
}
```

Issue #269, opened 2018-09-21
missing adapter method: notifyInteract, should call handleChipInteraction


moog16:
Chip should not have an onClick method handler, but instead should implement the notifyInteract. `notifyInteract` will trigger `handleChipInteraction` on the ChipSet.

Delete: 

```
handleSelect = (chipId) => {
    const {handleSelect, choice, filter} = this.props;
    if (filter || choice) {
      this.foundation_.toggleSelect(chipId);
    }
    handleSelect(this.foundation_.getSelectedChipIds());
  }
```

AND

```
onClick = (e) => {
    const {onClick, handleSelect, id} = this.props;
    onClick(e);
    handleSelect(id);
  }
```

- In `renderChip`, `handleSelect` should proxy `handleChipInteraction`.
- implement: `notifyInteract` to call `this.props.handleSelect`

Issue #268, opened 2018-09-21
Replace deselect call with handleChipRemoval in ChipSet `handleRemove`


moog16:
`handleRemove` should call `handleChipRemoval` instead of `deselect`. In the meantime, this component can mock the event.detail.id object. But another issue should be created after https://github.com/material-components/material-components-web/issues/3613 is fixed.

PR #424, opened 2018-11-09
fix(chips): update to v0.41.0

moog16:
fixes https://github.com/material-components/material-components-web-react/pull/400
fixes https://github.com/material-components/material-components-web-react/issues/269
fixes https://github.com/material-components/material-components-web-react/issues/268


packages/chips/Chip.js
```diff
@@ -79,7 +79,10 @@ export class Chip extends Component {
         (propertyName) => window.getComputedStyle(this.chipElement_).getPropertyValue(propertyName),
       setStyleProperty: (propertyName, value) => this.chipElement_.style.setProperty(propertyName, value),
       notifyRemoval: () => this.props.handleRemove(this.props.id),
-      notifyInteraction: () => this.props.handleSelect(this.props.id),
+      notifyInteraction: () => this.props.handleInteraction(this.props.id),
+      notifySelection: (selected) => {
+        this.props.handleSelect(this.props.id, selected);
```
bonniezhou: move to one line

packages/chips/ChipSet.js
```diff
@@ -29,27 +29,35 @@ import ChipCheckmark from './ChipCheckmark';
...
   checkmarkWidth_ = 0;
-  foundation_ = null;
-  state = {
-    selectedChipIds: new Set(this.props.selectedChipIds),
-  };
+
+  constructor(props) {
+    super(props);
+    this.state = {
+      selectedChipIds: props.selectedChipIds,
```
bonniezhou: I think we don't need to keep track of `selectedChipIds` in state because it's already being kept track of within the foundation, and we should use that as the source of truth. More comments to explain below

packages/chips/ChipSet.js
```diff
@@ -29,27 +29,35 @@ import ChipCheckmark from './ChipCheckmark';
...
-    if (this.props.selectedChipIds !== prevProps.selectedChipIds) {
-      const selectedChipIds = new Set(this.props.selectedChipIds);
+  componentDidUpdate(prevProps, prevState) {
+    const {selectedChipIds} = this.props;
+
+    if (this.state.foundation !== prevState.foundation) {
+      this.initChipSelection();
+    }
+    if (selectedChipIds !== prevProps.selectedChipIds) {
       this.setState({selectedChipIds});
```
bonniezhou:
```suggestion
      this.updateChipSelection(selectedChipIds);
```

packages/chips/ChipSet.js
```diff
@@ -29,27 +29,35 @@ import ChipCheckmark from './ChipCheckmark';
...
   }
 
-  componentDidUpdate(prevProps) {
-    if (this.props.selectedChipIds !== prevProps.selectedChipIds) {
-      const selectedChipIds = new Set(this.props.selectedChipIds);
+  componentDidUpdate(prevProps, prevState) {
+    const {selectedChipIds} = this.props;
+
+    if (this.state.foundation !== prevState.foundation) {
+      this.initChipSelection();
```
bonniezhou:
```suggestion
      this.initChipSelection(selectedChipIds);
```

packages/chips/ChipSet.js
```diff
@@ -65,54 +73,39 @@ export default class ChipSet extends Component {
...
+        this.setState({selectedChipIds}, () => {
+          this.props.handleSelect(selectedChipIds);
+        });
       },
+      removeChip: this.removeChip,
     };
   }
 
-  updateChipSelection(ids = this.state.selectedChipIds) {
+  initChipSelection() {
```
bonniezhou:
```suggestion
  initChipSelection(selectedChipIds) {
    this.updateChipSelection(selectedChipIds);
    this.setState({hasInitialized: true});
  }

  updateChipSelection(selectedChipIds) {
    React.Children.forEach(this.props.children, (child) => {
      const {id} = child.props;
      const selected = selectedChipIds.indexOf(id) > -1;
      if (selected) {
        this.state.foundation.select(id);
      } else {
        this.state.foundation.deselect_(id);
      }
    });
  }
```

packages/chips/ChipSet.js
bonniezhou: Here we need to update the foundation with the new `props.selectedChipIds` in order to handle single selection logic for choice chips.

and since we're going to update the foundation anyways, we don't need to hold the `selectedChipIds` in state anymore. We can just use `foundation.getSelectedChipIds()` to determine if a chip is selected or not.

packages/chips/ChipSet.js
```diff
@@ -65,54 +73,39 @@ export default class ChipSet extends Component {
...
       setSelected: (chipId, selected) => {
-        const {selectedChipIds} = this.state;
-        if (selected) {
-          selectedChipIds.add(chipId);
-        } else {
-          selectedChipIds.delete(chipId);
-        }
-        this.setState({selectedChipIds});
+        const selectedChipIds = this.state.foundation.getSelectedChipIds();
+        this.setState({selectedChipIds}, () => {
```
bonniezhou: remove `this.setState` since we're not keeping `selectedChipIds` in state anymore

packages/chips/ChipSet.js
bonniezhou: Since we're now updating the foundation every time props updates, we have to make `deselect_` public in MDC Web

packages/chips/ChipSet.js
```diff
@@ -145,10 +138,11 @@ export default class ChipSet extends Component {
   renderChip = (chip) => {
     const {filter} = this.props;
     const {selectedChipIds} = this.state;
```
bonniezhou:
```suggestion
    const selectedChipIds = this.state.foundation.getSelectedChipIds();
```

packages/chips/Chip.js
```diff
@@ -36,7 +36,7 @@ export class Chip extends Component {
 
   componentDidMount() {
     this.foundation_ = new MDCChipFoundation(this.adapter);
-    this.foundation_.init();
+    this.foundation_.init;
```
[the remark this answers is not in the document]
moog16: AH! weren't supposed to :(

packages/chips/ChipSet.js
moog16: We need it on state because we need to be able to initialize the chips via props. Also if the user wants to update the Chips via props we need to be able to update the respective chips.

test/unit/chips/Chip.test.js
```diff
@@ -91,10 +91,10 @@ test('#adapter.removeClassFromLeadingIcon removes from state.leadingIconClassLis
 });
 
 test('#adapter.notifyInteraction calls #props.handleSelect w/ chipId', () => {
```
bonniezhou: `calls #props.handleInteraction`

test/unit/chips/Chip.test.js
```diff
@@ -91,10 +91,10 @@ test('#adapter.removeClassFromLeadingIcon removes from state.leadingIconClassLis
...
-  const handleSelect = td.func();
-  const wrapper = shallow(<Chip id='123' handleSelect={handleSelect} />);
+  const handleInteraction = td.func();
+  const wrapper = shallow(<Chip id='123' handleInteraction={handleInteraction} />);
   wrapper.instance().foundation_.adapter_.notifyInteraction();
-  td.verify(handleSelect('123'), {times: 1});
+  td.verify(handleInteraction('123'), {times: 1});
 });
```
bonniezhou: Add test for `notifySelection`

test/unit/chips/ChipSet.test.js
```diff
@@ -46,115 +46,81 @@ test('#adapter.hasClass returns false if component does not contains class', ()
...
 
 test('#adapter.setSelected removes selectedChipId from state', () => {
   const wrapper = shallow(<ChipSet><div id='1' /></ChipSet>);
   wrapper.setState({selectedChipIds: new Set(['1'])});
   wrapper.instance().adapter.setSelected('1', false);
-  assert.isFalse(wrapper.state().selectedChipIds.has('1'));
+  assert.isFalse(wrapper.state().selectedChipIds.indexOf('1') > -1);
 });
 
 test('#adapter.setSelected removes selectedChipId from state', () => {
```
bonniezhou: I think this test an exact duplicate of the one above lol

moog16: ugh - too much copy paste :(

PR #425, opened 2018-11-09
feature(linear-progress): add component

### LINKED ISSUE #411: New component implementation: Linear progress
Hiya,

I'd like to add a linear progress component implementation. I am comfortable with the component and adapter creation.

The part I am unsure about though is how to add a _golden_ screenshot for this. The dots move at a rather rapid pace and I am scared they don't line up properly for subsequent screenshots because of timing issues.

Edit: this might also be an issue for the `indeterminate` variant.

Any advice for how to proceed with screenshots for this component?

hvolschenk:
Closes #411

### DIFF
--- a/packages/linear-progress/README.md
+++ b/packages/linear-progress/README.md
@@ -2,7 +2,7 @@
 
 # React Linear Progress
 
-A React version of an [MDC Linear Progress](https://github.com/material-components/material-components-web/tree/master/packages/mdc-linear-progress).
+A React version of an [MDC Linear Progress](https://github.com/material-components/material-components-web/tree/master/packages/mdc-linear-progress) component.
 
 ## Installation
 
@@ -33,8 +33,8 @@ class MyApp extends React.Component {
   render() {
     return (
       <LinearProgress
-        buffer={0.8}
-        progress={0.9}
+        buffer={0.9}
+        progress={0.8}
       />
     );
   }
@@ -50,7 +50,7 @@ bufferingDots | Boolean | Default `false`. Whether to show the buffer dots in th
 className | String | Optional. Additional class names to add to the root component
 closed | Boolean | Default `false`. When changed to `true`, closes the component
 indeterminate | Boolean | Default `false`. When set to `true`, renders the indeterminate variant
-progress | Number | Decimal value between 0 and 1, sets the progress amount
+progress | Number | Decimal value between 0 and 1, sets the progress bar width
 reversed | Boolean | Default `false`. When set to `true`, renders the reversed variant
 
 ## Sass Mixins
--- a/packages/linear-progress/README.md
+++ b/packages/linear-progress/README.md
@@ -1,5 +1,3 @@
-> ✨ Are you a part of the Material Design web community? Help us improve by filling out this <a href='https://bit.ly/materialwebsurvey'>**10 minute survey**</a>. ✨
-
 # React Linear Progress
 
 A React version of an [MDC Linear Progress](https://github.com/material-components/material-components-web/tree/master/packages/mdc-linear-progress) component.
--- a/packages/linear-progress/README.md
+++ b/packages/linear-progress/README.md
@@ -1,0 +1,50 @@
+> ✨ Are you a part of the Material Design web community? Help us improve by filling out this <a href='https://bit.ly/materialwebsurvey'>**10 minute survey**</a>. ✨
+
+# React Linear Progress
+
+A React version of an [MDC Linear Progress](https://github.com/material-components/material-components-web/tree/master/packages/mdc-linear-progress).
+
+## Installation
+
+```
+npm install @material/react-linear-progress
+```
+
+## Usage
+
+### Styles
+
+with Sass:
+```js
+import '@material/react-linear-progress/index.scss';
+```
+
+with CSS:
+```js
+import '@material/react-linear-progress/dist/linear-progress.css';
+```
+
+### Javascript Instantiation
+```js
+import React from 'react';
+```
+
+## Components
+
+### Grid
+
+#### Props
+
+Prop Name | Type | Description
+--- | --- | ---
+align | String (`left` or `right`) | An optional alignment of the grid contents
+children | Node | A React node to render within the Grid, usually a `Row` or `Row`s
+className | String | Classes to be applied to the root element
+tag | String | The tag type to render (default `'div'`)
+
+## Sass Mixins
+
+Sass mixins may be available to customize various aspects of the Components. Please refer to the
+MDC Web repository for more information on what mixins are available, and how to use them.
+
+[Advanced Sass Mixins](https://github.com/material-components/material-components-web/tree/master/packages/mdc-linear-progress#sass-mixins)
--- a/packages/linear-progress/README.md
+++ b/packages/linear-progress/README.md
@@ -52,6 +52,7 @@ closed | Boolean | Default `false`. When changed to `true`, closes the component
 indeterminate | Boolean | Default `false`. When set to `true`, renders the indeterminate variant
 progress | Number | Decimal value between 0 and 1, sets the progress bar width
 reversed | Boolean | Default `false`. When set to `true`, renders the reversed variant
+tag | String | Default `'div'`. The tag type to render
 
 ## Sass Mixins
 
--- a/packages/linear-progress/README.md
+++ b/packages/linear-progress/README.md
@@ -26,21 +26,32 @@ import '@material/react-linear-progress/dist/linear-progress.css';
 
 ### Javascript Instantiation
 ```js
+import LinearProgress from '@material/react-linear-progress';
 import React from 'react';
+
+class MyApp extends React.Component {
+  render() {
+    return (
+      <LinearProgress
+        buffer={0.8}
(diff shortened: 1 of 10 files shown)


moog16:
Nice! I'll take a look in a bit.

hvolschenk:
When you let me know which branch the actual target branch is (I am assuming it's not `master`) then I will rebase and sort out this `package-lock.json` conflict.

packages/linear-progress/README.md
```diff
@@ -0,0 +1,61 @@
...
+```js
+import LinearProgress from '@material/react-linear-progress';
+import React from 'react';
+
+class MyApp extends React.Component {
+  render() {
+    return (
+      <LinearProgress
+        buffer={0.8}
+        progress={0.9}
```
hvolschenk: I should rather swap these two around for the sample, otherwise copying this will not reveal the buffer.

packages/linear-progress/README.md
```diff
@@ -0,0 +1,61 @@
+> ✨ Are you a part of the Material Design web community? Help us improve by filling out this <a href='https://bit.ly/materialwebsurvey'>**10 minute survey**</a>. ✨
+
+# React Linear Progress
+
+A React version of an [MDC Linear Progress](https://github.com/material-components/material-components-web/tree/master/packages/mdc-linear-progress).
```
hvolschenk: The _article_ (an) doesn't feel right here.

Either make it "A React version of an MDC Linear Progress **component**." or "A React version of **~an~** MDC Linear Progress"

packages/linear-progress/README.md
```diff
@@ -0,0 +1,61 @@
...
+## Props
+
+Prop Name | Type | Description
+--- | --- | ---
+buffer | Number | Decimal value between 0 and 1, sets the buffer bar width
+bufferingDots | Boolean | Default `false`. Whether to show the buffer dots in the un-progressed section
+className | String | Optional. Additional class names to add to the root component
+closed | Boolean | Default `false`. When changed to `true`, closes the component
+indeterminate | Boolean | Default `false`. When set to `true`, renders the indeterminate variant
+progress | Number | Decimal value between 0 and 1, sets the progress amount
```
hvolschenk: progress ~amount~ bar width.

packages/linear-progress/index.js
```diff
@@ -0,0 +1,160 @@
+import {MDCLinearProgressFoundation} from '@material/linear-progress/dist/mdc.linearProgress';
+import classnames from 'classnames';
+import PropTypes from 'prop-types';
+import React from 'react';
+
+class LinearProgress extends React.Component {
+  constructor(props) {
+    super(props);
+    this.bufferElement_ = null;
+    this.foundation_ = new MDCLinearProgressFoundation(this.adapter);
```
hvolschenk: This is done here in the `constructor` instead of in `componentDidMount` as it is then possible to unit test the `componentDidMount` implementation.

If this happens inside `componentDidMount` (as in many other instances) the instantiation overwrites the `testdouble` mock.

I understand that unit tests should never influence the design of a class/component/module, I felt in this case it didn't make much difference, while gaining the confidence of being able to unit test this piece.

Let me know if this is frowned upon and I will gladly change it, or maybe there is another way I am missing.

test/screenshot/linear-progress/index.js
```diff
@@ -0,0 +1,77 @@
+import React from 'react';
+
+import './index.scss';
+import LinearProgress from '../../../packages/linear-progress/index';
+import '../../../packages/linear-progress/index.scss';
+
+class Closed extends React.Component {
```
hvolschenk: Name this `Closable` to better define what it is/does.

packages/linear-progress/README.md
moog16: I think `the` is more appropriate

packages/linear-progress/index.js
moog16: No you're right...we lack unit tests for componentDidMount. We've also been putting the foundation on state. The team hasn't had the time to do analysis on whether or not it should go on state, moved the constructor, or in the DidMount. I've been moving some to state, since rendering child components is dependent on foundation.

To be more consistent with what we've been doing, try putting it on state instead. That way we only have 2 patterns instead of 3.

moog16:
@hvolschenk it should be rc7.0. We're targeting November 20

packages/linear-progress/index.js
moog16: I think what you can do, is {disableLifeCycleMethods: true}, then call mount() in the tests. Does that work? Otherwise keep it in the constructor for now to unblock you. Then we can hash out the details when you're ready for a review.

hvolschenk: Unfortunately not. I did try that at first. The issue with that approach is that enzyme sees a `shallow()` mounted component as _mounted_, and a component that is considered _mounted_ cannot be re-mounted.

From [the docs](https://airbnb.io/enzyme/docs/api/ReactWrapper/mount.html):
> A method that re-mounts the component, if it is not currently mounted.

moog16:
@hvolschenk I think I may have figured out how to do the screenshot tests. There is a CSS property that can pause animations. Please see:

Pausing CSS animations for screenshot tests:

- https://github.com/material-components/material-components-web/blob/7b3942b350c31a303fe30f73a86afbe153b437f3/test/screenshot/spec/fixture.scss#L161-L169
- https://github.com/material-components/material-components-web/blob/7b3942b350c31a303fe30f73a86afbe153b437f3/test/screenshot/spec/mdc-linear-progress/classes/baseline.html#L43

I think you'll be able to copy what MDC Web has implemented, and get a similar screenshot test suite for linear progress.

test/unit/linear-progress/index.test.js
```diff
@@ -0,0 +1,149 @@
...
+import React from 'react';
+import td from 'testdouble';
+
+import LinearProgress from '../../../packages/linear-progress/index';
+
+suite('LinearProgress');
+
+test('Skips rendering the buffering dots when props.bufferingDots is false', () => {
+  const wrapper = shallow(<LinearProgress bufferingDots={false} />, {disableLifecycleMethods: true});
+  assert.isNotTrue(wrapper.find('.mdc-linear-progress__buffering-dots').exists());
```
moog16: just use assert.isFalse()

test/unit/linear-progress/index.test.js
```diff
@@ -0,0 +1,149 @@
...
+});
+
+test('#foundation_.setReverse gets called when props.reversed updates', () => {
+  const wrapper = mount(<LinearProgress reversed={false} />);
+  wrapper.instance().foundation_.setReverse = td.func();
+  wrapper.setProps({reversed: true});
+  td.verify(wrapper.instance().foundation_.setReverse(true), {times: 1});
+});
+
+test('#foundation_.destroy gets called when the component unmounts', () => {
```
moog16: please test adapter methods #adapter.addClass, #adapter.removeClass, etc

packages/linear-progress/index.js
```diff
@@ -0,0 +1,160 @@
...
+        return this.bufferElement_;
+      },
+      getPrimaryBar: () => {
+        return this.primaryBarElement_;
+      },
+      hasClass: (className) => {
+        return this.state.classList.has(className);
+      },
+      removeClass: (className) => {
+        const {classList} = this.state;
```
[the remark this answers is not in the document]
hvolschenk: I am quite sure you're right that this could happen. Have added it.

packages/linear-progress/index.js
```diff
@@ -0,0 +1,160 @@
...
+      // eslint-disable-next-line no-unused-vars
+      indeterminate,
+      // eslint-disable-next-line no-unused-vars
+      progress,
+      // eslint-disable-next-line no-unused-vars
+      reversed,
+      ...otherProps
+    } = this.props;
+    return (
+      <div className={this.classes} role="progressbar" {...otherProps}>
```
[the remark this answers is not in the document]
hvolschenk: I added the ability now for an optional `tag` prop. Also added it to unit tests and documentation.

I can't see _why_ someone would want to change it, but people do all kinds of things, so why not.

test/unit/linear-progress/index.test.js
hvolschenk: Cool. Added them.

hvolschenk:
@moog16 I have addressed the comments. Thanks for the feedback, always appreciated. Please have a look when you get a chance.

packages/linear-progress/README.md
```diff
@@ -0,0 +1,62 @@
+> ✨ Are you a part of the Material Design web community? Help us improve by filling out this <a href='https://bit.ly/materialwebsurvey'>**10 minute survey**</a>. ✨
```
moog16: please remove this...survey is over :)

moog16:
100% coverage! nice!

screen shot 2018-11-14 at 11 34 43

moog16:
Please sign the PR

hvolschenk:
I signed it

moog16:
#443 tests passing. Merging!

moog16:
Thanks @hvolschenk!!

Issue #408, opened 2018-11-02
[drawer] Change from modal and non-modal at runtime


lucasecdb:
hello guys! :-)

## Description

I'm using the `Drawer` component and I wanted it to change between modal and fixed version and vice-versa depending on the `window.innerWidth`, but the component crashes when it goes from `modal={false}` to `modal={true}` and I try to open it.

Bellow is the link to the codesandbox with the behaviour I've encountered.

## Reproduction sandbox

[![Edit 6wm7myn9yk](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/6wm7myn9yk)

moog16:
@lucasecdb thanks for reporting this issue. Totally a use case we need to support. The issue is permanent does not initialize a foundation. I'll fix the issue with 7.0

PR #427, opened 2018-11-09
fix(drawer): initialize foundation if props.open updates

moog16:
fixes https://github.com/material-components/material-components-web-react/issues/408


packages/drawer/index.js
```diff
@@ -51,16 +59,18 @@ class Drawer extends React.Component {
...
-    }
   }
 
   componentDidUpdate(prevProps) {
     const {dismissible, modal, open} = this.props;
+    const changedToModal = prevProps.modal !== this.props.modal;
+    const changedToDismissible = prevProps.dismissible !== this.props.dismissible;
     if (!(dismissible || modal)) return;
 
+    if ((changedToModal || changedToDismissible) && !this.foundation_) {
```
bonniezhou:
```suggestion
    if (!this.foundation_) {
```
If I'm understanding this right, it doesn't matter whether it was changed to modal/dismissible or if it was already modal/dismissible, but as long as it's modal/dismissible and doesn't have a foundation then we should initialize it, right?

moog16: Modal has a different foundation from dismissible. So actually this might be incorrect

Issue #409, opened 2018-11-03
TabBar / Tab issues


pladaria:
Hi, 

I'm happy to see material components in React. I've been playing with some components and I've detected several issues in tab components:

* `activeIndex` is used to set component initial state and after that it is ignored. This looks like a bad controlled component (props should be the source of truth for controlled components).
* Ripple animation is fired twice

I've been unable to reproduce [these examples](https://material-components.github.io/material-components-web-catalog/#/component/tabs) using react.  Perhaps I'm doing something wrong, some examples would be appreciated.

pladaria:
I've created a codesandox to reproduce those problems. The code is taken from `react-tab-bar` README

https://codesandbox.io/s/3v9xz998v1

To reproduce the double ripple issue, open codesandbox preview in new page, open dev tools and set a mobile viewport

moog16:
@pladaria thanks for opening the issue...RE `activeIndex`, you can programmatically update the activeIndex and the `componentDidUpdate` will catch that change. https://github.com/material-components/material-components-web-react/blame/master/packages/tab-bar/index.js#L32
So I'm not sure what you mean by that issue, but I did find a different issue, which I have a fix for.

RE: double ripple - I do see that, and will also be fixing. Strange that its happening, but maybe its because there is a touch event as well.

PR #428, opened 2018-11-09
fix(tab-bar): set previousActiveIndex to props.activeIndex initially to avoid tab of undefined error

moog16:
partial fix https://github.com/material-components/material-components-web-react/issues/409


test/screenshot/tab-bar/index.js
```diff
@@ -9,6 +9,12 @@ class TabBarTest extends React.Component {
     activeIndex: this.props.activeIndex || 0, // eslint-disable-line react/prop-types
   };
 
+  componentDidUpdate(prevProps) {
```
bonniezhou: This can be removed if we don't wrap the test components in too many layers. See comments below

test/screenshot/tab-bar/index.js
```diff
@@ -9,6 +9,12 @@ class TabBarTest extends React.Component {
...
+    if (this.props.activeIndex !== prevProps.activeIndex) {
+      this.setState({activeIndex: this.props.activeIndex});
+    }
+  }
+
   render() {
     const {
```
bonniezhou:
```suggestion
    const {
      /* eslint-disable react/prop-types */
      className,
      indicatorContent,
      isFadingIndicator,
      numTabs,
      isRtl,
      hasUpdateActiveIndexButton,
      /* eslint-enable react/prop-types */
    } = this.props;

    const renderTab = (num, index) => {
      return (
        <Tab
          key={index}
          isFadingIndicator={isFadingIndicator}
          indicatorContent={indicatorContent}
        >
          <span className='mdc-tab__text-label'>Tab {num + 1}</span>
        </Tab>
      );
    };

    return (
      <React.Fragment>
        <TabBar
          isRtl={isRtl}
          className={className}
          activeIndex={this.state.activeIndex}
          handleActiveIndexUpdate={(activeIndex) => this.setState({activeIndex})}
        >
          {[...Array(numTabs).keys()].map(renderTab)}
        </TabBar>
        {hasUpdateActiveIndexButton ? <button onClick={() => this.setState({activeIndex: 1})}>Click</button> : null}
      </React.Fragment>
    );
```

test/screenshot/tab-bar/index.js
```diff
@@ -45,9 +51,24 @@ class TabBarTest extends React.Component {
...
+      </React.Fragment>
+    );
+  }
+}
+
 const TabBarScreenshotTest = () => {
   return (
     <div>
+      <div>
+        <TabBarActiveIndexTest />
```
bonniezhou:
```suggestion
        <TabBarTest numTabs={3} hasUpdateActiveIndexButton/>
```

test/screenshot/tab-bar/index.js
bonniezhou: I think we're trying to wrap too many layers, which makes the screenshot test code very error prone. How about we add the button inside `TabBarTest` instead and hide it behind a boolean like `hasUpdateActiveIndexButton`?

test/screenshot/tab-bar/index.js
```diff
@@ -45,9 +51,24 @@ class TabBarTest extends React.Component {
   }
 }
 
+class TabBarActiveIndexTest extends React.Component {
```
bonniezhou: This can be removed too if we don't wrap the test components in too many layers

packages/tab-bar/index.js
```diff
@@ -11,8 +11,11 @@ export default class TabBar extends Component {
...
 
   foundation_ = null;
-  state = {
-    previousActiveIndex: -1,
+  constructor(props) {
+    super(props);
+    this.state = {
+      previousActiveIndex: props.activeIndex,
+    };
   }
```
bonniezhou: In the process of debugging the screenshot tests I found another bug: we forgot to update `state.previousActiveIndex` when `props.activeIndex` updates
```suggestion
  componentDidUpdate(prevProps) {
    if (this.props.activeIndex !== prevProps.activeIndex) {
      this.setState({previousActiveIndex: prevProps.activeIndex},
        () => this.foundation_.activateTab(this.props.activeIndex));
    }
    if (this.props.indexInView !== prevProps.indexInView) {
      this.foundation_.scrollIntoView(this.props.indexInView);
    }
  }
```

test/screenshot/tab-bar/index.js
moog16: Good call!

test/unit/tab-bar/index.test.js
```diff
@@ -25,9 +25,9 @@ test('#componentWillUnmount destroys foundation', () => {
   td.verify(foundation.destroy(), {times: 1});
 });
 
-test('initially sets state.previousActiveIndex to -1', () => {
+test('initially sets state.previousActiveIndex to 0', () => {
```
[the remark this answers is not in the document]
moog16: thanks for catching that!

PR #430, opened 2018-11-12
Fix/text field custom validation

### LINKED ISSUE #421: bug(text-field): Custom validation only works on second blur
Hiya,

Version: `@material/react-text-field`: `0.6.2`

I am trying to add custom validation to a `react-text-field` component but getting some strange behaviour.
The error message only becomes an error message the second time I blur the field, the first time it still shows as a standard `HelperText` message.

I am testing with this component (I just added it as a sample on `master` on my fork of this repo):

```
import React from 'react';

import TextField, {HelperText, Input} from '../../../packages/text-field/index';

class CustomValidation extends React.Component {
  constructor(props) {
    super(props);
    this.state = {error: undefined, value: ''};
    this.onChange = this.onChange.bind(this);
    this.renderHelperText = this.renderHelperText.bind(this);
    this.validate = this.validate.bind(this);
  }
  onChange(e) {
    this.setState({value: e.target.value});
    this.validate();
  }
  validate() {
    this.setState({error: 'BAD'});
  }
  renderHelperText() {
    if (!!this.state.error) {
      return (
        <HelperText
          isValid={false}
          isValidationMessage
          persistent
          validation
        >
          {this.state.error}
        </HelperText>
      );
    } else {
      return <HelperText persistent>Helper text</HelperText>;
    }
  }
  render() {
    return (
      <TextField
        helperText={this.renderHelperText()}
        label="Bad input"
      >
        <Input
          onBlur={this.validate}
          onChange={this.onChange}
          name="bad-field"
          isValid={!(!!this.state.error)}
          value={this.state.value}
        />
      </TextField>
    );
  }
}

export default CustomValidation;
```

From my observations the `HelperText` component gets the correct classes applied immediately, but the `TextField` only adds the `mdc-text-field--invalid` class the second time I blur the field.

It also feels rather strange that I have to add the `isValidationMessage` and `validation` props to the `HelperText` component. It's not clear to me what the difference is from the [README.md](https://github.com/material-components/material-components-web-react/blob/v0.6.2/packages/text-field/helper-text/README.md) file.

Maybe I am just misunderstanding something here, but I feel like I have tried all variants to no avail. If it is just my misunderstanding I'll gladly update the documentation with a sample of how it should work.

hvolschenk:
fixes #421 

This does fix the issue, I am just not sure if this is how you want it to be fixed :blush:


packages/text-field/Input.js
```diff
@@ -44,6 +51,10 @@ export default class Input extends React.Component {
     if (value) {
       handleValueChange(value, () => foundation.setValue(value));
     }
+    if (isValid !== undefined && foundation.setUseNativeValidation) {
```
moog16: when would the foundation.setUseNativeValidation not exist?

packages/text-field/Input.js
```diff
@@ -79,6 +91,15 @@ export default class Input extends React.Component {
         }
       });
     }
+
+    if (isValid !== prevProps.isValid) {
```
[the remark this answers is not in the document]
hvolschenk: The issue was/is that in the example there is some custom validation.

To make the validation message appear I needed to _focus_ the field, _blur_ it, _focus_ it again, and then _blur_ it a second time for the validation message to appear.

packages/text-field/Input.js
hvolschenk: This was me just getting the sample to you. Will remove soon.

The only place this doesn't exist is during the tests when we overwrite the foundation(s) manually.

hvolschenk:
@moog16 I have updated the helper-text `README.md` to show custom validation. Not quite sure what to add in the main `README.md`. Should I add a reference to the helper-text one? Or create a duplicate example of the custom validation?

packages/text-field/Input.js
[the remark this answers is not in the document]
moog16: gotcha...interesting that it doesn't show up on the screenshot tests. I added `isValid={false}` on the <Input />, and it worked as expected.

packages/text-field/helper-text/README.md
```diff
@@ -65,6 +74,62 @@ class MyApp extends React.Component {
...
+    this.setState({isValid: value.includes('@'), username: value});
+  }
+  renderHelperText() {
+    if (this.state.isValid) {
+      return (<HelperText>Please enter your Username</HelperText>);
+    } else {
+      return (
+        <HelperText
+          isValid={false}
+          isValidationMessage
```
moog16: I see what you're saying in #421. let's remove isValidationMessage or validation as well and make them the same. If you wanna make this a different task we can.

packages/text-field/helper-text/README.md
```diff
@@ -65,6 +74,62 @@ class MyApp extends React.Component {
...
+    const {value} = e.target;
+    this.setState({isValid: value.includes('@'), username: value});
+  }
+  renderHelperText() {
+    if (this.state.isValid) {
+      return (<HelperText>Please enter your Username</HelperText>);
+    } else {
+      return (
+        <HelperText
+          isValid={false}
```
moog16: I think this should be set to this.state.isValid.

Also I'll create a task to revisit helperText.props.isValid. Its confusing and is probably implemented incorrectly.

packages/text-field/helper-text/README.md
```diff
@@ -65,6 +74,62 @@ class MyApp extends React.Component {
...
+    return (
+      <div>
+        <TextField
+          box
+          label='Username'
+          helperText={this.renderHelperText()}
+        >
+          <Input
+            isValid={this.state.isValid}
+            name="username"
```
moog16: nit: change to single quotes. We're using singles in JSX in this file.

moog16:
@hvolschenk please also sign the PR

packages/text-field/helper-text/README.md
hvolschenk: Adding link to #435 here for reference.

packages/text-field/helper-text/README.md
hvolschenk: Let's please solve this through #435 if that's all right.

hvolschenk:
I signed it

moog16:
Thanks! I changed the base to rc7.0. And will wait for #436 to pass, then merge it!

PR #431, opened 2018-11-13
docs(ripple): add sass clarification to docs

moog16:


packages/ripple/README.md
```diff
@@ -64,6 +64,19 @@ const RippleIcon = withRipple(Icon);
 Wrap your Icon component with the HOC `withRipple`, which returns a component
 with a ripple capable surface.
 
+You also need to include the sass mixins in order for this to work.
```
bonniezhou: Move this to the "Styles" section (under Sass)

bonniezhou: And maybe add a note under the CSS section that says only hover/focus styles will show up

packages/ripple/README.md
```diff
@@ -24,6 +24,20 @@ with CSS:
 import '@material/react-ripple/dist/ripple.css';
 ```
 
+You'll also need to include these sass mixins on the element. Please also refer to [Advanced Sass Mixins](https://github.com/material-components/material-components-web/blob/master/packages/mdc-ripple/README.md#sass-apis) to customize further.
+
+```sass
+@import "@material/ripple/common";
```
bonniezhou: Aren't these mixins already imported in `@material/react-ripple/index.scss`?

packages/ripple/README.md
```diff
@@ -24,6 +24,20 @@ with CSS:
...
+@import "@material/ripple/mixins";
+
+// refer to element in Javascript portion below
+.ripple-icon-component {
```
bonniezhou: This code should be under "with Sass" because I think this doesn't apply at all to users only using the `dist/ripple.css`

PR #442, opened 2018-11-14
chore revert survey

moog16:


moog16:
Probably...I was just thinking about how it will appear in the Changelog, or if we even want it to show up there. I'll just merge it to master and deal with the changelog when we release.

PR #448, opened 2018-11-15
feat(radio): add component

moog16:
fixes https://github.com/material-components/material-components-web-react/issues/157


packages/radio/NativeControl.js
```diff
@@ -0,0 +1,54 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+import React from 'react';
+import classnames from 'classnames';
+import PropTypes from 'prop-types';
+
+const NativeControl = (props) => {
+  const {
+    setRippleActivator,
```
bonniezhou: can we replace this with `rippleActivatorRef` similar to how it is in Checkbox?

packages/radio/index.js
```diff
@@ -0,0 +1,161 @@
...
+        {label ? <label htmlFor={nativeControlId}>{label}</label> : null}
+      </div>
+    );
+  }
+
+  renderNativeControl() {
+    const children = React.Children.only(this.props.children);
+    const updatedProps = Object.assign({}, children.props, {
+      disabled: this.state.disabled,
+      setRippleActivator: this.setRippleActivator,
```
bonniezhou: we can pass the rippleActivator as a ref directly to `NativeControl` (see comment above)

Issue #48, opened 2018-05-13
Add TypeScript support


ebeloded:
This issue will inevitably come up at some point: the community will need type definitions to reliably utilize the components in TypeScript projects.

Since this project is still at an early stage - maybe instead of creating and maintaining a separate package with definitions, it would make more sense to transition the project to TypeScript? This will provide a guarantee that the code, including changes and new features, are always in sync.

patrick91:
Yeah, TypeScript by default will really be helpful, it will be much easier the maintain, especially since most of the components won't be super complex anyway

bonniezhou:
Thanks for the feedback! It's not currently on our roadmap but we hear your feedback. We will add it to our tracker.

kraenhansen:
I would love to help this transition happen.

pspeter3:
I'm also excited and happy for this to happen.

nickmccurdy:
Would type contributions be accepted for existing components, or should we wait?

pspeter3:
I think that's a valid concern. One way we can mitigate that is by generating the TypeScript definitions automatically from the `PropTypes` and `defaultProps`. For example, Button has:

```js
Button.propTypes = {
  raised: PropTypes.bool,
  unelevated: PropTypes.bool,
  outlined: PropTypes.bool,
  disabled: PropTypes.bool,
  unbounded: PropTypes.bool,
  initRipple: PropTypes.func,
  className: PropTypes.string,
  icon: PropTypes.element,
  children: PropTypes.string,
};

Button.defaultProps = {
  raised: false,
  unelevated: false,
  outlined: false,
  disabled: false,
  unbounded: false,
  initRipple: () => {},
  className: '',
  icon: null,
  children: '',
};
```

The TypeScript would be

```ts
export interface Props {
    raised?: boolean;
    unelevated?: boolean;
    outline?: boolean;
    disabled?: boolean;
    initRipple?: Function;
    className?: string;
    icon?: React.ReactElement<any>
    children?: string;
}
```

Would you be open to a script which added type definitions?

nickmccurdy:
Maybe we can base types off those from [RMWC](https://github.com/jamesmfriedman/rmwc), which are autogenerated from Flow type declarations based on MDC.

jamesmfriedman:
Yeah for RMWC I actually wrote the code using Flow (more forgiving than TS) and wrote a ridiculously simple conversion to Typescript that happens during build time that relies heavily on convention.

[...]

PR #471, opened 2018-11-27
feat(text-field): add typescript support

moog16:
related https://github.com/material-components/material-components-web-react/issues/48

This converts all of
- Text field, notched outline, floating label, and line ripple
- screenshot tests
- unit tests


packages/floating-label/index.tsx
```diff
@@ -0,0 +1,140 @@
...
+};
+
+export default class FloatingLabel extends React.Component<
+  Props<HTMLLabelElement>,
+  FloatingLabelState
+  > {
+  foundation_?: MDCFloatingLabelFoundation;
+  labelElement: React.RefObject<HTMLLabelElement> = React.createRef();
+
+  static defaultProps: Partial<Props<HTMLLabelElement>> = {
```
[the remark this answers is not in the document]
moog16: we don't need a default for handleWidthChange, since it is just undefined.


I'm going to change to Props<HTMLLabelElement> because that is the props going into the component. I'm renaming to maybe make more sense. LMK what you think.

packages/floating-label/index.tsx
```diff
@@ -0,0 +1,140 @@
...
+import {MDCFloatingLabelFoundation} from '@material/floating-label';
+
+
+export interface FloatingLabelProps {
+  className?: string,
+  handleWidthChange?: (width: number) => void,
+  float?: boolean,
+};
+
+type Props<T> = FloatingLabelProps & React.LabelHTMLAttributes<T>;
```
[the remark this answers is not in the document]
moog16: I don't think so... lol I"ll remove that generic

packages/line-ripple/index.tsx
```diff
@@ -0,0 +1,144 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+import * as React from 'react';
+import * as classnames from 'classnames';
+// @ts-ignore
+import {MDCLineRippleFoundation} from '@material/line-ripple';
+
+export interface LineRippleProps {
+  className?: string,
```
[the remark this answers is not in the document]
moog16: style is part of `React.LabelHTMLAttributes<HTMLLabelElement>;`

packages/notched-outline/index.tsx
```diff
@@ -0,0 +1,137 @@
...
+        if (!this.idleElement_.current) return;
+        return window
+          .getComputedStyle(this.idleElement_.current)
+          .getPropertyValue(propertyName);
+      },
+    };
+  }
+
+  render() {
+    return [
```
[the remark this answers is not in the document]
moog16: This is the old way. Now we do Fragment. I can change it, but thought we didn't want to change anything unnecessary. What do you think?

packages/notched-outline/index.tsx
```diff
@@ -0,0 +1,137 @@
...
+  notch: boolean,
+  notchWidth: number
+};
+
+interface NotchedOutlineState {
+  classList: Set<string>
+};
+
+export default class NotchedOutline extends React.Component<
+  NotchedOutlineProps,
```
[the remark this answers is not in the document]
moog16: i didn't add it because a user cannot add `...otherProps` to our component. So there is no need.

packages/text-field/index.tsx
```diff
@@ -19,58 +19,107 @@
...
+  notchedLabelWidth: number;
+  activeLineRipple: boolean;
+  lineRippleCenter: number;
+  outlineIsNotched: boolean;
+  showHelperTextToScreenReader: boolean;
+  isValid: boolean;
+};
+
+class TextField<T extends {}> extends React.Component<TextFieldProps<T>, TextFieldState> {
+  floatingLabelElement: React.RefObject<FloatingLabel> = React.createRef();
```
bonniezhou: `floatingLabelElement_`?

packages/floating-label/index.tsx
```diff
@@ -19,28 +19,41 @@
...
+};
 
-export default class FloatingLabel extends React.Component {
-  foundation_ = null;
+export default class FloatingLabel extends React.Component<
+  FloatingLabelProps,
+  FloatingLabelState
+  > {
+  foundation_?: MDCFloatingLabelFoundation;
+  labelElement: React.RefObject<HTMLLabelElement> = React.createRef();
```
bonniezhou: `labelElement_`? I don't know if we've been consistent with our use of underscores for refs

packages/text-field/index.tsx
```diff
@@ -272,17 +327,14 @@ class TextField extends React.Component {
         {trailingIcon ? this.renderIcon(trailingIcon) : null}
       </div>
     );
-
     if (helperText) {
-      return ([
-        textField, this.renderHelperText(),
-      ]);
+      return [textField, this.renderHelperText()];
```
bonniezhou: reminder for other PR: change this to Fragment too

packages/text-field/index.tsx
```diff
@@ -158,7 +214,6 @@ class TextField extends React.Component {
     //    valid: boolean,
     //  },
     // }
-
     return {
       getNativeInput: () => {
         let badInput;
```
bonniezhou: why don't we need to type any of the variables in here?

packages/text-field/index.tsx
```diff
@@ -294,8 +346,9 @@ class TextField extends React.Component {
       <FloatingLabel
         className={floatingLabelClassName}
         float={this.state.labelIsFloated}
-        handleWidthChange={
-          (initialLabelWidth) => this.setState({initialLabelWidth})}
+        handleWidthChange={(initialLabelWidth) =>
```
bonniezhou: do we need to type `initialLabelWidth`?

packages/text-field/index.tsx
```diff
@@ -19,58 +19,107 @@
...
-import PropTypes from 'prop-types';
-import classnames from 'classnames';
-import {MDCTextFieldFoundation} from '@material/textfield/dist/mdc.textfield';
-
-import Input from './Input';
-import Icon from './icon';
-import HelperText from './helper-text';
+import * as React from 'react';
+import * as classnames from 'classnames';
+// no mdc .d.ts
```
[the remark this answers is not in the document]
moog16: .d.ts is a typings file. And MDC Web doesn't have them yet. Just leaving this comment to come back to when MDC Web has them available.

packages/text-field/index.tsx
moog16: It is implied by the FloatingLabel definition.

packages/text-field/icon/index.tsx
```diff
@@ -61,9 +77,21 @@ export default class Icon extends React.Component {
 
   get adapter() {
     return {
-      getAttr: (attr) => this.state[attr],
-      setAttr: (attr, value) => this.setState({[attr]: value}),
-      removeAttr: (attr) => this.setState({[attr]: null}),
+      // need toString() call when tabindex === 0.
```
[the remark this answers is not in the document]
moog16: If state.tabindex is a number. If its passed as a `0`, but the foundation is expecting a string.

packages/text-field/index.tsx
moog16: changed!

packages/floating-label/index.tsx
moog16: ya we haven't, but since everything is already I will change this one. I think we should remove all of them eventually and use the `private` keyword if possible

packages/text-field/index.tsx
moog16: I changed it

PR #480, opened 2018-11-29
fix(infrastructure): memoize puppeteer browser instance during screenshot test suite

moog16:
This issue was popping up: https://github.com/GoogleChrome/puppeteer/issues/594, which describes too many `puppeteer.launch()` calls executing before one is killed. The limit seems to be at 10, so on the 11th it would error. I thought a fix would be to only call `puppeteer.launch()` once, which is what this PR is doing. However the issue actually turned out to be due to expired GCP account creds.

This PR actually doesn't fix anything except we now only use one puppeteer instance, which seems better than have kicking off a new one for every screenshot test. I haven't been able to find any measurable time savings in the screenshot tests.


docs/screenshot-tests.md
```diff
@@ -27,7 +27,7 @@ docker pull mdcreact/screenshots
 Ensure that you have a `MDC_GCLOUD_SERVICE_ACCOUNT_KEY` environment variable (ie. ~/.bashrc, ~/.bash_profile). Otherwise you will need to paste it inline in the command below. You can get this off of the Google Cloud Platform Admin Tool (3rd party contributors will not be able to download/upload screenshots until this is part of the process is changed).
 
 ```
-docker run -it --rm --cap-add=SYS_ADMIN -e MDC_GCLOUD_SERVICE_ACCOUNT_KEY="${MDC_GCLOUD_SERVICE_ACCOUNT_KEY}" mdcreact/screenshots /bin/sh
+docker run -it --rm --cap-add=SYS_ADMIN -e MDC_GCLOUD_SERVICE_ACCOUNT_KEY="${MDC_GCLOUD_SERVICE_ACCOUNT_KEY}" mdcreact/screenshots /bin/bash
```
moog16: This gives us auto complete when in the docker instance.

moog16:
Closing as this has gone stale

PR #487, opened 2018-12-04
fix(infrastructure): unit test code coverage for typescript

moog16:
related #48 

This took me a long time to find. But basically my issue was `karma-coverage` does not work with `istanbul-instrumenter-loader`. Solution/Issue is documented on https://github.com/webpack-contrib/istanbul-instrumenter-loader/issues/32. This github issue led me to using `karma-coverage-istanbul-reporter` instead, which is a fork of karma-coverage.


scripts/karma/config.js
```diff
@@ -24,11 +24,11 @@ module.exports = {
   // preprocess matching files before serving them to the browser
   // available preprocessors: https://npmjs.org/browse/keyword/karma-preprocessor
   preprocessors: {
-    'test/unit/index.tsx': ['webpack'],
+    'test/unit/index.tsx': ['webpack', 'sourcemap'],
```
bonniezhou: What does `sourcemap` do?

scripts/karma/config.js
```diff
@@ -45,12 +45,21 @@ module.exports = {
...
           },
         }, {
-          test: /\.tsx?$/,
+          test: /\.tsx$/,
           loader: 'ts-loader',
+        }, {
+          enforce: 'post',
+          test: /\.(js|ts)x?$/,
+          use: {
+            loader: 'istanbul-instrumenter-loader',
```
bonniezhou: My understanding is that we're using `istanbul-instrumenter-loader` now instead of the `istanbul` webpack plugin, is that correct? can we remove `istanbul` from the package.json or does the loader require it?

scripts/karma/config.js
moog16: When you get minified or transpiled code (like things generated by ts-loader and babel-loader), its difficult to read when debugging. Sourcemaps provide a way for the browser to show the developer the original code (or at least readable). :)

scripts/karma/config.js
moog16: `istanbul-instrumenter-loader` requires `istanbul`...like when create-react-app requires `node-sass` when you want to use sass.

We needed to switch it because we the istanbul webpack plugin does not work with typescript.

PR #494, opened 2018-12-06
feat(drawer): typescript support 

moog16:
related #48


packages/drawer/index.tsx
```diff
@@ -0,0 +1,253 @@
...
+        const hasPreviousFocus =
+          this.previousFocus_ && this.previousFocus_.focus;
+        const drawerElement =
+          this.drawerElement_ && this.drawerElement_.current;
+        if (
+          drawerElement &&
+          hasPreviousFocus &&
+          drawerElement.contains(document.activeElement)
+        ) {
+          this.previousFocus_!.focus();
```
rlfriedman: Is the ! here intentional? Formatting of this if statement also seems a bit weird.

packages/drawer/index.tsx
```diff
@@ -0,0 +1,253 @@
...
+        }
+      },
+      focusActiveNavigationItem: () => {
+        const drawerElement =
+          this.drawerElement_ && this.drawerElement_.current;
+        if (!drawerElement) return;
+        const activeNavItemEl = drawerElement.querySelector(
+          `.${listCssClasses.LIST_ITEM_ACTIVATED_CLASS}`
+        ) as HTMLElement | null;
```
rlfriedman: Formatting/lack of indentation here is a bit confusing.

packages/drawer/index.tsx
moog16: UGH! it was the formatter...
And yes is intentional. For some reason the tslint doesn't see the IF check I have for this.previousFocus

moog16:
@rlfriedman updated!

PR #495, opened 2018-12-06
feat(fab): typescript support

moog16:
related #48


packages/fab/index.tsx
```diff
@@ -0,0 +1,79 @@
...
+    return null;
+  }
+  const updatedProps = {
+    className: classnames('mdc-fab__icon', icon.props.className),
+  };
+  return React.cloneElement(icon, updatedProps);
+};
+
+const TextLabel: React.FunctionComponent<{textLabel: string}> = ({
+  textLabel, // eslint-disable-line react/prop-types
```
williamernest: Are the eslint comments a temporary measure until we completely convert everything? 

We may want to make another issue to install [google/ts-tyle](https://github.com/google/ts-style) into this repo.

packages/fab/index.tsx
```diff
@@ -0,0 +1,79 @@
...
+import withRipple from '@material/react-ripple';
+
+export interface FabProps {
+  mini: boolean,
+  icon?: React.ReactElement<HTMLElement>,
+  textLabel: string,
+  className: string,
+  initRipple: (surface: HTMLButtonElement) => void,
+  unbounded: boolean,
+};
```
williamernest: Nit: semicolon isn't needed at the end of interfaces.

packages/fab/index.tsx
```diff
@@ -0,0 +1,79 @@
...
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+import * as React from 'react';
```
williamernest: Nit: Add new line after license.

test/screenshot/fab/index.tsx
```diff
@@ -0,0 +1,250 @@
+import * as React from 'react';
+// @ts-ignore
```
williamernest: I'm not sure why the Fab import has to specify .tsx (which is the error being suppressed here). Can you document why this is required in a comment?

packages/fab/index.tsx
moog16: ya temporary. I didn't want to break all the other js files until all are converted to TS

moog16: https://github.com/material-components/material-components-web-react/issues/507

test/screenshot/fab/index.tsx
moog16: This is because we still have the index.js files. This is temporary until we get all TSX files in.

PR #502, opened 2018-12-08
feat(menu-surface): typescript

moog16:
related #48


packages/menu-surface/index.tsx
```diff
@@ -0,0 +1,357 @@
...
+  },
+  onClose: () => void,
+  onOpen: () => void,
+  onKeyDown: (event: React.KeyboardEvent) => void,
+  quickOpen: boolean,
+  open: boolean,
+  fixed: boolean
+};
+
+type MenuSurfaceState = {
```
tec27: I'd probably prefer using an `interface` over a `type` alias here, as they'll give you better TS compiler error messages (and allow for extension, but that's probably not particularly important here)

packages/menu-surface/index.tsx
```diff
@@ -0,0 +1,357 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+import * as React from 'react';
+import * as classnames from 'classnames';
+import {MDCMenuSurfaceFoundation, Corner} from '@material/menu-surface';
+
+export interface MenuSurfaceProps {
+  className: string,
+  anchorElement: HTMLElement,
```
tec27: Seems like you use this as an optional value, but the type implies it will always be present. You probably want `anchorElement?: HTMLElement`

packages/menu-surface/index.tsx
```diff
@@ -0,0 +1,357 @@
...
+  anchorCorner: number,
+  anchorMargin: object,
+  styles: object,
+  coordinates: {
```
tec27: It feels like it might be slightly more idiomatic to expect that `coordinates` is either:

- completely undefined, or
- fully defined (both `x` and `y` present)

rather than allowing for partial definition of the object

packages/menu-surface/index.tsx
```diff
@@ -0,0 +1,357 @@
...
+// THE SOFTWARE.
+import * as React from 'react';
+import * as classnames from 'classnames';
+import {MDCMenuSurfaceFoundation, Corner} from '@material/menu-surface';
+
+export interface MenuSurfaceProps {
+  className: string,
+  anchorElement: HTMLElement,
+  anchorCorner: number,
+  anchorMargin: object,
```
tec27: Could this have a more specific type?

packages/menu-surface/index.tsx
```diff
@@ -0,0 +1,357 @@
...
+  styles: object,
```
tec27: You *might* be able to type this as `CSSProperties` from the React typings, although I'm not 100% sure what the limitations of that are

packages/menu-surface/index.tsx
```diff
@@ -0,0 +1,357 @@
...
+    styleRight: undefined,
+    styleTop: undefined,
+    styleBottom: undefined,
+    classList: new Set(),
+  } as MenuSurfaceState;
+
+  static defaultProps = {
+    className: '',
+    styles: {},
+    anchorElement: undefined,
```
tec27: I'd probably avoid adding keys that have `undefined` values. I'm not sure if it affects how TS 3.0's defaultProps handling works, but it feels a bit odd to me. Just leaving the keys out is probably preferred

packages/menu-surface/index.tsx
```diff
@@ -0,0 +1,357 @@
...
+
+  state = {
+    transformOrigin: '',
+    maxHeight: undefined,
+    styleLeft: undefined,
+    styleRight: undefined,
+    styleTop: undefined,
+    styleBottom: undefined,
+    classList: new Set(),
+  } as MenuSurfaceState;
```
tec27: I don't *think* this cast is necessary, I believe the `React.Component` typing should already type this field as `MenuSurfaceState` and check against the value you're setting it to

moog16: What about 
``` state: MenuSurfaceState = {
    transformOrigin: '',
    maxHeight: undefined,
    styleLeft: undefined,
    styleRight: undefined,
    styleTop: undefined,
    styleBottom: undefined,
    classList: new Set(),
  };
```

packages/menu-surface/index.tsx
tec27: 1) You'd only need to export the state type if you expect people to extend it in some way, but generally I'd say it's an internal implementation detail.

2) I'd probably just export the base type (without the intersection). If you wanted to include the `div` props for anyone that was extending your component class, it might be better to use an `extends` rather than type intersection (e.g. `interface MenuSurfaceProps extends React.HTMLProps<HTMLDivElement> {}`)

moog16: gotcha - I will only export the base props.

packages/menu-surface/index.tsx
tec27: Should work fine, but still redundant to the underlying types I think (`React.Component` declares the state field with the right type already, provided your generic parameters are correct). Not a problem to re-state it explicitly though, if you think it makes the code clearer.

moog16: Ah I changed it to this as Amr said its better to be more strict. But from the lens of React it is redundant. I think I'm going to keep it since its in all of the components right now. If we decide to remove them in the future, I can do it in one PR

Issue #505, opened 2018-12-10
Active tabs don't have an underline


nworden:
Sorry in advance if the problem is me, I'm pretty new to this library (and React (and JavaScript)).

[The Tab docs](https://github.com/material-components/material-components-web-react/tree/master/packages/tab) say that active Tabs should have an underline indicator by default. However, that doesn't seem to be working.

Here's my code:
```javascript
<TabBar
  activeIndex={0}
  handleActiveIndexUpdate={(activeIndex) => this.setState({activeIndex})}
>
  <Tab>
    <span className='mdc-tab__text-label'>
      Tab A
    </span>
  </Tab>
  <Tab>
    <span className='mdc-tab__text-label'>
      Tab B
    </span>
  </Tab>
</TabBar>
```

I do get a span with classes mdc-tab-indicator and mdc-tab-indicator--active, but it has a height and width of 0 (I'm not sure if it's actually relevant to the expected underline or if it's just a wrapper for people who set an indicator icon).

nworden:
Ah, thanks Matt, yeah I was missing the tab-indicator CSS.

moog16:
Keeping it open to update Readme!

moog16:
merged #510.

PR #510, opened 2018-12-11
Include tab-indicator with Tab documentation.

nworden:
From discussion on issue #505.

### DIFF
--- a/packages/tab/README.md
+++ b/packages/tab/README.md
@@ -6,6 +6,7 @@ A React version of an [MDC Tab](https://github.com/material-components/material-
 
 ```
 npm install @material/react-tab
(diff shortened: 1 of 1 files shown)


nworden:
This is just a documentation change, so I don't think it could have caused a build failure.

moog16:
No, we documented on the open source contributions in this [doc](https://github.com/material-components/material-components-web-react/blob/master/docs/contributions/README.md). One of the members needs to open a new PR specifically for tests, since your branch is not in our repository yet. 

But you're right, this is just a doc change, so we can just  merge it. Typically you need to sign the CLA ( with a separate comment, "I signed it"), but it looks like its passing. I'm going to merge it.

Issue #492, opened 2018-12-06
Selected ListItem styles not working with SSR


dan-kwiat:
With client-side rendering the first item gets the `mdc-list-item--selected` class and `aria-selected="true"` as expected.  However with server-side rendering (tested with next.js) this is not the case (and all the items look the same).  I believe it's an issue with this library rather than material-components-web?

```jsx
<List
  singleSelection
  selectedIndex={0}
>
  <ListItem>
    <ListItemText primaryText='One' />
  </ListItem>
  <ListItem>
    <ListItemText primaryText='Two' />
  </ListItem>
  <ListItem>
    <ListItemText primaryText='Three' />
  </ListItem>
</List>
```

moog16:
@dan-kwiat yes from my understanding of how SSR works, I believe the problem is with MDC React and not Web. I'm not super familiar with SSR. 

Do you have a test project that I can look at/learn from with the List HTML shown above?

dan-kwiat:
@moog16 to be honest I'm pretty new to SSR as well.  This example has a List in the Drawer (see DrawerList.js): https://github.com/dan-kwiat/material-next-template

In development the first item is selected as expected (see screenshot below) and it responds to item clicks.  In next's static build however no item is selected (see demo) and `handleSelect` is never triggered.

![screenshot 2018-12-07 at 09 34 57](https://user-images.githubusercontent.com/5648607/49639648-6cae2c80-fa03-11e8-89ae-3b27d09d8ff1.png)

moog16:
I visited your demo https://dan-kwiat.github.io/material-next-template/ and downloaded your project. I noticed that the `setSelected` method is not being called either, which leads me to believe that your environment isn't properly setup. 

In your dev setup, clicking on a list item console.log(), but this does not happen in your live demo.

dan-kwiat:
@moog16 it isn't being called because `List`'s `handleSelect` method isn't triggered.  Setting the state in an explicit click event handler on `ListItem` works in both cases (see updated demo & code) but the active class still isn't applied.

dan-kwiat:
@moog16 `componentDidMount` is not executed server side but still gets called on first rendering by the client.

I think the problem is the use of `child.type.name` in @material/react-list.  The type name for `<ListItem>` children is `"ListItem"` with CSR but `"t"` for SSR (see updated demo logs).  Therefore with SSR `this.listItemCount` is always zero and so is `index` in all the event handlers.

[...]

PR #509, opened 2018-12-11
fix(list): minification-safe element type comparison

dan-kwiat:
#492 
I didn't have a chance to test this - please do so before merging

PR #511, opened 2018-12-11
DO NOT MERGE: TESTS fix(list): change ListItem check to use child.type

moog16:
fixes: https://github.com/material-components/material-components-web-react/issues/492


moog16:
Closed for #509

PR #512, opened 2018-12-11
Fix/unmounting select

### LINKED ISSUE #489: bug(text-field): destroying foundation fails
Hiya,

This is a bit of a tricky one to describe so please bear with me for a moment.

I have a `react-redux` container which connects a form to a store. On mount a fetch is fired off to get a list of countries to display in a `@material/react-select`. The form itself is part of a render-prop so it only gets rendered when the list of countries is fetched successfully.

When rendering the component for the first time the state goes from _Pending_ to _Success_ and thus swaps out a loader for the actual form. All good so far.

When I navigate to another route and hit the back button in my browser, the state is still _Success_ so the form starts rendering immediately. The mount fetch method fires off which then immediately swaps the status to _Pending_ again. It is at this point where `@material/react-text-field` breaks.

As the swapping between statuses happens so fast the `@material/react-text-field` never actually gets mounted before un-mounting again. When it un-mounts `this.foundation_.destroy()` gets called but at that point `this.foundation_` is still set to `null`, thus throwing the classic `Cannot read property 'destroy' of null` error.

**Possible solutions**:

I could change the status of the countries list to _Pending_ or something similar myself the moment the form un-mounts. This doesn't sound right though.

! could check if the countries list is already fetched on mount and then not fetch it again on the second mount. This works but not quite what I'd prefer as this same problem could be seen for lists that I actually _do_ want updated every time to prevent stale data.

I could add a check to the `@material/react-text-field`'s `componentWillUnmount` method to only destroy the foundation if it already exists. I would prefer this, but would have to do this for all form controls as the same could happen on selects, checkboxes etc.

Please let me know what you think.

hvolschenk:
Fixes #489

### DIFF
--- a/packages/checkbox/index.js
+++ b/packages/checkbox/index.js
@@ -69,7 +69,9 @@ export class Checkbox extends Component {
   }
 
   componentWillUnmount() {
-    this.foundation_.destroy();
+    if (this.foundation_) {
+      this.foundation_.destroy();
+    }
   }
 
   init = (el) => {
--- a/packages/radio/index.js
+++ b/packages/radio/index.js
@@ -59,7 +59,9 @@ class Radio extends React.Component {
   }
 
   componentWillUnmount() {
-    this.foundation_.destroy();
+    if (this.foundation_) {
+      this.foundation_.destroy();
+    }
   }
 
   componentDidUpdate(prevProps, prevState) {
--- a/packages/select/index.js
+++ b/packages/select/index.js
@@ -72,7 +72,9 @@ export default class Select extends React.Component {
   }
 
   componentWillUnmount() {
-    this.foundation_.destroy();
+    if (this.foundation_) {
+      this.foundation_.destroy();
+    }
   }
 
   onChange = (evt) => {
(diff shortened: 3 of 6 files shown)


packages/checkbox/index.js
```diff
@@ -69,7 +69,9 @@ export class Checkbox extends Component {
   }
 
   componentWillUnmount() {
-    this.foundation_.destroy();
+    if (this.foundation_) {
+      this.foundation_.destroy();
+    }
```
hvolschenk: This change (and all the others like it below) are not testable.
This means the PR here will bring the coverage amount down substantially.

It's a bit of a chicken/egg scenario as I need the foundation **to not exist** to check that `destroy()` is not called, but I need the foundation **to exist** to stub out `destroy()`. Sadface.

hvolschenk: > Merging #512 into master will increase coverage by 0.01%

Huh? If you say so _codecov_.

moog16:
@hvolschenk please sign :)

moog16:
opened test PR: https://github.com/material-components/material-components-web-react/pull/514

hvolschenk:
I signed it

PR #521, opened 2018-12-14
feat(typography): typescript

moog16:
related #48


packages/typography/index.tsx
```diff
@@ -0,0 +1,75 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+import typographyHOC from './typography';
+
+export const Body1 = typographyHOC<HTMLParagraphElement>({classModifier: 'body1', tag: 'p'});
+
+export const Body2 = typographyHOC<HTMLParagraphElement>({classModifier: 'body2', tag: 'p'});
+
+export const Button = typographyHOC<HTMLSpanElement>({classModifier: 'button', tag: 'span'});
```
patrickrodee: Should the tag be "button" instead of span? I guess that would make these all button elements which would be ugly?

packages/typography/typography.tsx
```diff
@@ -0,0 +1,44 @@
+import * as classnames from 'classnames';
+import * as React from 'react';
+
+export interface TypographyProps<T> extends React.HTMLProps<T> {
+  children?: React.ReactNode,
```
patrickrodee: Mostly stylistic but I'm fairly certain you can/should use semicolons instead of commas for terminating interface types.

https://www.typescriptlang.org/docs/handbook/interfaces.html#optional-properties

packages/typography/typography.tsx
```diff
@@ -0,0 +1,44 @@
...
+interface EnhancedProps {
+  tag: string,
+  classModifier: string,
+}
+
+const typographyHOC = <T extends {}>(options: EnhancedProps) => {
+  const {tag, classModifier} = options;
+
+  const Typography: React.FunctionComponent<TypographyProps<T>> = ({
+    /* eslint-disable */
```
patrickrodee: Why is es-lint disabled here?

packages/typography/index.tsx
moog16: We used span because the [catalog](https://material-components.github.io/material-components-web-catalog/#/component/typography) uses span.

packages/typography/typography.tsx
patrickrodee: Comments will be removed once using ts-lint instead of es-lint.

patrickrodee:
🆗

PR #525, opened 2018-12-14
feat(material-icon): typescript

moog16:
related #48


packages/material-icon/index.tsx
```diff
@@ -19,86 +19,61 @@
...
-  render() {
-    const {
-      icon,
-      hasRipple,
-      ...otherProps
-    } = this.props;
+export default class MaterialIcon extends React.Component<
+  MaterialIconProps,
+  {}
+  > {
```
rlfriedman: This seems like it is formatted a bit oddly to me, was confused about the '<' on this line.

packages/material-icon/index.tsx
```diff
@@ -19,86 +19,61 @@
...
+  static defaultProps: Partial<MaterialIconProps> = {
+    icon: '',
+    className: '',
+    hasRipple: false,
+    unbounded: false,
+  };
 
+  render() {
```
rlfriedman: Does this need a return type? I guess in general in these PRs so far render hasn't had one, but wanted to check.

packages/material-icon/index.tsx
rlfriedman: Yeah I guess I'm not totally sure what would be better, separate lines are probably fine. Up to you!

packages/material-icon/index.tsx
rlfriedman: Cool, thanks for the explanation!

PR #526, opened 2018-12-14
feat(linear-progress): typescript

moog16:
related #48


packages/linear-progress/index.tsx
```diff
@@ -1,21 +1,52 @@
...
+interface LinearProgressState {
+  classList: Set<string>;
+};
+
+class LinearProgress<T extends {} = HTMLDivElement> extends React.Component<
+  LinearProgressProps<T>,
+  LinearProgressState
+  > {
   isMounted_ = false;
+  bufferElement_: React.RefObject<HTMLDivElement> = React.createRef();
```
williamernest: Remove trailing underscore.

packages/linear-progress/index.tsx
```diff
@@ -124,43 +155,27 @@ class LinearProgress extends React.Component {
       ...otherProps
     } = this.props;
     return (
-      <Tag className={this.classes} role="progressbar" {...otherProps}>
-        {bufferingDots && <div className="mdc-linear-progress__buffering-dots"></div>}
-        <div className="mdc-linear-progress__buffer" ref={this.bufferElement_}></div>
+      // @ts-ignore
```
williamernest: Comment on why this ignore is needed. 

Also, why do we need to destructure all of the props here if most of them are unused ?

moog16: they need to be defined so that they don't appear in ...otherProps. It is like black listing

packages/linear-progress/index.tsx
```diff
@@ -1,21 +1,52 @@
...
+  classList: Set<string>;
+};
+
+class LinearProgress<T extends {} = HTMLDivElement> extends React.Component<
+  LinearProgressProps<T>,
+  LinearProgressState
+  > {
   isMounted_ = false;
+  bufferElement: React.RefObject<HTMLDivElement> = React.createRef();
+  primaryBarElement_: React.RefObject<HTMLDivElement> = React.createRef();
```
williamernest: remove trailing underscore.

packages/linear-progress/index.tsx
```diff
@@ -1,21 +1,52 @@
...
+  foundation_: MDCLinearProgressFoundation;
```
williamernest: remove trailing underscore.

moog16: actually can use the private keyword!

moog16:
`isMounted` is a reserved react class term. So I have to keep the underscore on the `isMounted_`.

PR #527, opened 2018-12-14
feat(ripple): typescript

moog16:
related: #48


packages/ripple/index.tsx
```diff
@@ -0,0 +1,319 @@
...
+    WrappedComponent: React.ComponentType<P>
+  ) => {
+  return class RippledComponent extends React.Component<
+  // Subtract removes any props "InjectedProps" if they are on "P"
+  // This allows the developer to override any props
+  // https://medium.com/@jrwebdev/react-higher-order-component-patterns-in-typescript-42278f7590fb
+    Subtract<P, InjectedProps<Surface, Activator>> & RippledComponentProps<Surface>,
+    RippledComponentState
+    > {
+    foundation_: MDCRippleFoundation | null = null;
```
patrickrodee: Are these private? If so, should we drop the trailing underscore and prefix them with `private`?

patrickrodee: Also, could this be an optional type instead of nullable?

packages/ripple/index.tsx
```diff
@@ -0,0 +1,319 @@
...
+  className: string,
+  onMouseDown: React.MouseEventHandler<T>,
+  onMouseUp: React.MouseEventHandler<T>,
+  onTouchStart: React.TouchEventHandler<T>,
+  onTouchEnd: React.TouchEventHandler<T>,
+  onKeyDown: React.KeyboardEventHandler<T>,
+  onKeyUp: React.KeyboardEventHandler<T>,
+  onFocus: React.FocusEventHandler<T>,
+  onBlur: React.FocusEventHandler<T>,
+  computeBoundingRect?: (surface: T) => {},
```
patrickrodee: Why is this optional?

packages/ripple/index.tsx
```diff
@@ -0,0 +1,319 @@
...
+
+    get style() {
+      const {style: wrappedStyle} = this.props;
+      const {style} = this.state;
+      return Object.assign({}, style, wrappedStyle);
+    }
+
+    render() {
+      const {
+        /* start black list of otherprops */
```
patrickrodee: Nit: use a different name (maybe exclusion list?)

packages/ripple/index.tsx
moog16: RE nullable v optional: ya!
RE private: But if I make it private I refer to it in the tests.

patrickrodee: You could declare it as private then add a public getter for both. Would that work or are you setting them in the tests, too?

moog16: in some instances I do set the foundation for stubbing/spying. I know these should be private, so I was thinking we'd do another round of system changes to converting things to be private. Although adding the getter/setter doesn't sound too bad. What are your thoughts?

patrickrodee: I see you also have `initializeFoundation_` and `createAdapter_` which are public but should probably be private, too. Some ideas:
1. Make the `foundation` property configurable via the constructor
2. Make the constructor take an optional factory function that returns an implementation of `MDCRippleFoundation`. You can then update the constructor to set the private foundation property to the return value of the foundation factory OR initialize it with the default behavior.

I can mock up what 2 looks like if you want.

patrickrodee: Ugh I forgot this is React which means modifying the constructor is silly... You could add a `MDCRippleFoundationFactory` field to the `props` that accepts an optional function with the signature `(surface: Surface, activator?: Activator): MDCRippleFoundation`. Then you could pass configure that via props while still having the `foundation` and `unbounded` fields private.

packages/ripple/index.tsx
```diff
@@ -0,0 +1,317 @@
...
+    // surface: This element receives the visual treatment (classes and style) of the ripple.
+    // activator: This element is used to detect whether to activate the ripple. If this is not
+    // provided, the ripple surface will be used to detect activation.
+    initializeFoundation_ = (surface: Surface, activator?: Activator) => {
+      const adapter = this.createAdapter_(surface, activator);
+      this.foundation_ = new MDCRippleFoundation(adapter);
+      this.foundation_.init();
+    };
+
+    createAdapter_ = (surface: Surface, activator?: Activator) => {
```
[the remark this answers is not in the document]
moog16: got a lot of errors when I tried making them private :(

moog16:
@amrtgaber & @patrickrodee Thanks for the initial reviews. I updated according to your requested changes. Please let me know what you think. Will be online in the afternoon.

PR #532, opened 2018-12-17
feat(tab): typescript

moog16:
related #48


packages/tab/index.tsx
```diff
@@ -19,35 +19,70 @@
...
+  tabIndex?: number;
+  activateIndicator: boolean;
+  previousIndicatorClientRect?: ClientRect;
+}
 
-import TabRipple from './TabRipple';
+// https://github.com/material-components/material-components-web-react/issues/528
+// TODO: switch when ripple is converted to TSX
+function isElementTabRipple(element: any): element is TabRippleBase {
+  return element !;
```
patrickrodee: What's this cute little exclamation mark doing?

packages/tab/index.tsx
```diff
@@ -19,35 +19,70 @@
...
+}
 
-export default class Tab extends Component {
-  foundation_ = null;
-  tabElement_ = React.createRef();
-  tabContentElement_ = React.createRef();
-  tabIndicator_ = React.createRef();
-  tabRipple_ = React.createRef();
+export default class Tab extends React.Component<TabProps, TabState> {
+  foundation_: MDCTabFoundation | null = null;
```
patrickrodee: Can we make these private?

packages/tab/index.tsx
```diff
@@ -79,34 +114,41 @@ export default class Tab extends Component {
...
+          previousIndicatorClientRect,
+        }),
       deactivateIndicator: () => this.setState({activateIndicator: false}),
-      // computeIndicatorClientRect is redundant in mdc-tab and is going to be
-      // removed in another release
     };
   }
 
-  activate(computeIndicatorClientRect) {
+  activate(computeIndicatorClientRect?: ClientRect) {
```
patrickrodee: Why is this function a method while the others are instance fields?

test/screenshot/tab/index.tsx
```diff
@@ -0,0 +1,130 @@
...
+  activeIndex: number;
+  previousActiveIndex: number;
+}
+
+class TabsController extends React.Component<TabsControllerProps, TabsControllerState> {
+  tabBoundingRects: {[n: number]: ClientRect} = {};
+
+  state = {activeIndex: 0, previousActiveIndex: 0};
+
+  tabInit = (tabEl: Tab | null, index: number) => {
```
patrickrodee: Can tabEl be optional instead of nullable?

packages/tab/index.tsx
patrickrodee: Also can foundation be optional instead of nullable?

packages/tab/index.tsx
moog16: You mean why do we have activate on the class? I believe this is because tabbar needs this to be on the class instance. 

In tab bar's `componentDidMount`

```
if (this.tabList_[activeIndex]) {
      this.tabList_[activeIndex].activate({} /* previousIndicatorClientRect */);
}
```

This is similar to the vanilla version

test/screenshot/tab/index.tsx
moog16: No it has to be nullable. But as @appsforartists pointed out it should be a React.Ref.

test/unit/tab/index.test.tsx
```diff
@@ -35,143 +37,153 @@ test('adds a class from state.classList', () => {
...
   wrapper.instance().adapter.setAttr('aria-selected', true);
   assert.isTrue(wrapper.state()['aria-selected']);
 });
 
 test('#adapter.getOffsetLeft returns tabElement_.offsetLeft', () => {
-  const wrapper = mount(<Tab />);
-  assert.equal(wrapper.instance().adapter.getOffsetLeft(), wrapper.instance().tabElement_.current.offsetLeft);
+  const wrapper = mount<Tab>(<Tab />);
+  assert.equal(
+    wrapper.instance().adapter.getOffsetLeft(),
```
patrickrodee: Can we revert this layout change?

test/unit/tab/index.test.tsx
```diff
@@ -35,143 +37,153 @@ test('adds a class from state.classList', () => {
...
+    wrapper.instance().tabElement_.current!.offsetLeft
+  );
 });
 
 test('#adapter.getOffsetWidth returns tabElement_.offsetWidth', () => {
-  const wrapper = mount(<Tab>Text</Tab>);
-  assert.equal(wrapper.instance().adapter.getOffsetWidth(), wrapper.instance().tabElement_.current.offsetWidth);
+  const wrapper = mount<Tab>(<Tab>Text</Tab>);
+  assert.equal(
+    wrapper.instance().adapter.getOffsetWidth(),
```
patrickrodee: Can we revert this layout change?

test/unit/tab/index.test.tsx
```diff
@@ -35,143 +37,153 @@ test('adds a class from state.classList', () => {
...
+  );
 });
 
 test('#adapter.getContentOffsetLeft returns tabContentElement_.offsetLeft', () => {
-  const wrapper = mount(<Tab>Text</Tab>);
+  const wrapper = mount<Tab>(<Tab>Text</Tab>);
   assert.equal(
     wrapper.instance().adapter.getContentOffsetLeft(),
-    wrapper.instance().tabContentElement_.current.offsetLeft);
+    wrapper.instance().tabContentElement_.current!.offsetLeft
```
patrickrodee: What is the exclamation mark doing here?

test/unit/tab/index.test.tsx
```diff
@@ -35,143 +37,153 @@ test('adds a class from state.classList', () => {
...
+  );
 });
 
 test('#adapter.getContentOffsetWidth returns tabContentElement_.offsetWidth', () => {
-  const wrapper = mount(<Tab>Text</Tab>);
+  const wrapper = mount<Tab>(<Tab>Text</Tab>);
   assert.equal(
     wrapper.instance().adapter.getContentOffsetWidth(),
-    wrapper.instance().tabContentElement_.current.offsetWidth);
+    wrapper.instance().tabContentElement_.current!.offsetWidth
```
patrickrodee: Same as above

test/unit/tab/index.test.tsx
```diff
@@ -35,143 +37,153 @@ test('adds a class from state.classList', () => {
...
 });
 
 test('#computeIndicatorClientRect returns the tabIndicator_ clientRect', () => {
-  const wrapper = mount(<Tab/>);
+  const wrapper = mount<Tab>(<Tab />);
   wrapper.instance().tabIndicator_.current.computeContentClientRect = td.func();
   wrapper.instance().computeIndicatorClientRect();
-  td.verify(wrapper.instance().tabIndicator_.current.computeContentClientRect(), {times: 1});
+  td.verify(
+    wrapper.instance().tabIndicator_.current.computeContentClientRect(),
```
patrickrodee: Can we revert this layout change?

patrickrodee:
BTW I'm happy to commit the above requested changes to the branch if you'd like.

test/unit/tab/index.test.tsx
moog16: It thinks it is possible that current might be null. So I just used the `!` to tell its wrong

test/screenshot/tab/index.tsx
```diff
@@ -0,0 +1,134 @@
...
+    return (
+      <Tabs>
+        {[1, 2, 3].map((num, index) => (
+          <Tab
+            active={index === activeIndex}
+            key={index}
+            previousIndicatorClientRect={this.tabBoundingRects[previousActiveIndex]}
+            // https://github.com/DefinitelyTyped/DefinitelyTyped/issues/28884
+            // @ts-ignore
+            ref={(el: React.Ref<Tab>) => this.tabInit(el, index)}
```
appsforartists: TS Style says you have to use `=> { /*code*/ }` if you don't use the result of a lambda.

appsforartists:
I'm disappointed that I can't block this PR with a sniper review…

test/screenshot/tab/index.tsx
appsforartists: I think your type is wrong here, and that this should not be `ts-ignore`d.  This does not use the `styled-components` library, which is what the linked issue is concerning.

If you look at the [typings for `Ref`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts#L65), you'll see that it's either a function that is passed a `T`, or the result of `createRef()`.  You've confused the two here.

`(el: Tab)`  is the correct signature.  You might not even need the `: Tab`, since TypeScript can often infer local types.

packages/tab/index.tsx
```diff
@@ -19,35 +19,70 @@
...
+import * as classnames from 'classnames';
+// TODO: Come back and use .d.ts file
+// @ts-ignore
 import TabIndicator from '@material/react-tab-indicator';
+// No mdc .d.ts files
+// @ts-ignore
 import {MDCTabFoundation} from '@material/tab/dist/mdc.tab';
+import TabRipple, {TabRipple as TabRippleBase} from './TabRipple';
+
+export interface TabProps extends React.HTMLProps<HTMLButtonElement> {
```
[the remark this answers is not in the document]
moog16: we named the main component `index` then if there were other auxiliary components we would create a new file/component.

But after doing these components I also prefer that proposal. Its more consistent.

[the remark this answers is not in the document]
moog16: https://github.com/material-components/material-components-web-react/issues/553

packages/tab/index.tsx
```diff
@@ -19,35 +19,70 @@
...
+    isFadingIndicator: false,
+    indicatorContent: null,
+    minWidth: false,
+    isMinWidthIndicator: false,
+    stacked: false,
+  };
 
-  state = {
+  state: TabState = {
     'classList': new Set(),
```
appsforartists: `classList` is a `DOMTokenList`, which is not constructable.  It looks like `Set` has some similar methods, but not all (e.g. `contains`, `item`).  What is `classList` used for?  Have you asserted that the overlapping set of functionality is all you need (e.g. nobody is going to try to call `contains` on your `Set`)?

This should all be commented.

Moreover, I don't see you using this state in your component.  That smells really badly to me.  Why are you storing something that isn't tied to rendering in `State`?

packages/tab/index.tsx
```diff
@@ -19,35 +19,70 @@
...
-
+import * as React from 'react';
+import * as classnames from 'classnames';
+// TODO: Come back and use .d.ts file
+// @ts-ignore
 import TabIndicator from '@material/react-tab-indicator';
+// No mdc .d.ts files
+// @ts-ignore
 import {MDCTabFoundation} from '@material/tab/dist/mdc.tab';
+import TabRipple, {TabRipple as TabRippleBase} from './TabRipple';
```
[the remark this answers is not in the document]
moog16: Used on line 222.

The TabRippleBase is without the ripple. Do you propose I rename the TabRippleBase in the TabRIpple.jsx file?

packages/tab/index.tsx
```diff
@@ -79,34 +114,41 @@ export default class Tab extends Component {
...
         classList.delete(className);
         this.setState({classList});
       },
-      hasClass: (className) => this.classes.split(' ').includes(className),
-      setAttr: (attr, value) => this.setState({[attr]: value}),
-      getOffsetLeft: () => this.tabElement_.current && this.tabElement_.current.offsetLeft,
-      getOffsetWidth: () => this.tabElement_.current && this.tabElement_.current.offsetWidth,
-      getContentOffsetLeft: () => this.tabContentElement_.current && this.tabContentElement_.current.offsetLeft,
-      getContentOffsetWidth: () => this.tabContentElement_.current && this.tabContentElement_.current.offsetWidth,
+      hasClass: (className: string) => this.classes.split(' ').includes(className),
```
appsforartists: It's unclear what your canonical data model is.  You have both `classList` and a getter `classes`.  `classList` contains some, but not all, of your classes.  You're delegating to an external library to generate a string, and then splitting it to see the results.

I don't know when/how often `hasClass` is called, but this is a surprising amount of logic for something that looks fairly simple.

packages/tab/index.tsx
```diff
@@ -79,34 +114,41 @@ export default class Tab extends Component {
...
+      setAttr: (attr: keyof TabState, value?: string | boolean) => (
```
appsforartists: Is this the right typing?  Do you expect this to pass?

```typescript
.setAttr(previousIndicatorClientRect)
```

Does it?

packages/tab/index.tsx
moog16: Used between lines 106 and 125.

Its a pattern we have in most components. Since its on state no one externally should call contains on the Set.

packages/tab/index.tsx
```diff
@@ -79,34 +114,41 @@ export default class Tab extends Component {
...
+        this.setState((prevState) => ({...prevState, [attr]: value}))
+      ),
+      getOffsetLeft: () =>
```
appsforartists: Should these be cast to `Number`?  Is it OK for `getOffsetLeft` to return `false`?

packages/tab/index.tsx
moog16: This is how we wrap the react components using the MDC Foundation/Adapter pattern. hasClass and any other adapter methods are used in MDC Web's foundation. 

We had to design for MDC's Foundation, the developer and static classes, hence the complexity.

packages/tab/index.tsx
appsforartists: FYI: using lambdas instead of methods is against TypeScript Style:

> Classes should usually not contain properties initialized to arrow functions. Arrow function properties require the calling function to understand that the callee's this is already bound, which increases confusion about what this is, and call sites and references using such handlers look broken (i.e. require non-local knowledge to determine that they are correct). Code should always use arrow functions to call instance methods (const handler = (x) => this.listener(x);), and should not obtain or pass references to instance methods (const handler = this.listener; handler(x);).
> 
> Note: in some specific situations, e.g. when binding functions in a template, arrow functions as properties are useful and create much more readable code. Use judgement with this rule.

appsforartists: Looks like you've got a lot of methods that are defined as lambdas and contain references to `this`.

packages/tab/index.tsx
```diff
@@ -142,27 +202,19 @@ export default class Tab extends Component {
       isMinWidthIndicator,
       ...otherProps
     } = this.props;
-
-    const {
-      tabIndex,
-      ['aria-selected']: ariaSelected,
-    } = this.state;
-
+    const {tabIndex, ['aria-selected']: ariaSelected} = this.state;
```
appsforartists: Why did you collapse these?

Personally, I find vertically-delimited lists much easier to maintain over time.

packages/tab/index.tsx
moog16: I mainly expect it to pass aria-selected or tabindex.

test/screenshot/tab/index.tsx
```diff
@@ -0,0 +1,134 @@
...
+import '../../../packages/tab-indicator/index.scss';
+import '../../../packages/tab/index.scss';
+import './index.scss';
+
+interface TabContentType {
+  num: number;
+}
+
+const TabContent: (args: TabContentType) => React.ReactElement<any> = ({
+  num, // eslint-disable-line react/prop-types
```
appsforartists: It's unclear to me why you are disabling this.  Isn't it just `number`?

test/screenshot/tab/index.tsx
```diff
@@ -0,0 +1,134 @@
...
+
+interface TabsProps {
+  children: React.ReactNode;
+}
+
+function isElementTab(element: any): element is Tab {
+  return element !== null;
+}
+
+const Tabs: (args: TabsProps) => JSX.Element = ({
```
appsforartists: This would be so much easier to read if you just used `function`.

test/screenshot/tab/index.tsx
```diff
@@ -0,0 +1,134 @@
...
+    <MaterialIcon className='mdc-tab__icon' icon='favorite' />
+    <span className='mdc-tab__text-label'>Tab {num}</span>
+  </React.Fragment>
+);
+
+interface TabsProps {
+  children: React.ReactNode;
+}
+
+function isElementTab(element: any): element is Tab {
```
appsforartists: This is a bad type guard.  I can call `isElementTab(5)` and `5` would be treated like a `Tab`.

You should make the check more specific, or rename the function to `isNotNull`.

test/screenshot/tab/index.tsx
```diff
@@ -0,0 +1,134 @@
...
+  activeIndex: number;
+  previousActiveIndex: number;
+}
+
+class TabsController extends React.Component<TabsControllerProps, TabsControllerState> {
+  tabBoundingRects: {[n: number]: ClientRect} = {};
+
+  state = {activeIndex: 0, previousActiveIndex: 0};
+
+  tabInit = (tabEl: React.Ref<Tab>, index: number) => {
```
appsforartists: As I've noted down below, I think this signature is wrong.

test/screenshot/tab/index.tsx
```diff
@@ -0,0 +1,134 @@
...
+    if (isElementTab(tabEl)) {
+      this.tabBoundingRects[index] = tabEl.computeIndicatorClientRect();
+    }
+  }
+
+  render() {
+    const {
+      tabContent, // eslint-disable-line
```
appsforartists: Why `disable-line`?

test/screenshot/tab/index.tsx
```diff
@@ -0,0 +1,134 @@
...
+      <h3>Tabs w/ Min Width Tab Indicator</h3>
+      <TabsController
+        isMinWidthIndicator
+        stacked
+        tabContent={(num) => <TabContent num={num} />}
+      />
+    </div>
+  );
+};
+export default TabScreenshotTest;
```
appsforartists: Default exports are discouraged by TS Style.

test/screenshot/tab/index.tsx
moog16: its eslint-disable...react/prop-types are no needed, and I will address this in a later issue

test/unit/tab/index.test.tsx
```diff
@@ -35,143 +37,140 @@ test('adds a class from state.classList', () => {
...
-  const wrapper = mount(<Tab active />);
+  const wrapper = mount<Tab>(<Tab active />);
   assert.exists(wrapper.instance().foundation_);
 });
 
 test('if props.active updates to true, activate is called with previousIndicatorClientRect prop', () => {
-  const clientRect = {test: 1};
-  const wrapper = shallow(<Tab previousIndicatorClientRect={clientRect}/>);
-  wrapper.instance().activate = td.func();
+  const clientRect = {test: 1} as unknown as ClientRect;
```
appsforartists:
```typescript
export function coerceForTesting<T>(value: {}): T {
  return value as T;
}
```

I wrote that function for our tests in the Sketch plug-in.  It does the same thing, but in a more obvious way:

```typescript
   const clientRect = coerceForTesting<ClientRect>({test: 1});
```

appsforartists: I recommend using it in all the places here you are using `as`.

test/screenshot/tab/index.tsx
moog16: so it would be like this:

```
function isNotNull(element: any): element is Tab {
```

That wouldn't look too good either?

packages/tab/index.tsx
moog16: https://github.com/material-components/material-components-web-react/pull/548

test/screenshot/tab/index.tsx
moog16: https://github.com/material-components/material-components-web-react/issues/555

test/screenshot/tab/index.tsx
moog16: https://github.com/material-components/material-components-web-react/issues/556

packages/tab/TabRipple.tsx
```diff
@@ -20,47 +20,41 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
-import React from 'react';
-import PropTypes from 'prop-types';
+import * as React from 'react';
 import classnames from 'classnames';
-import {withRipple} from '@material/react-ripple';
+import * as Ripple from '@material/react-ripple';
```
appsforartists: By convention, I usually leave a newline between packages I didn't author and ones I did.

```typescript
import * as React from 'react';
import * as classnames from 'classnames';

import {withRipple} from '@material/react-ripple';
import {somethingElse} from '@material/react-something-else';

import {something} from './a_local_file';
import {anotherThing} from './another_local_file';
```

packages/tab/TabRipple.tsx
```diff
@@ -20,47 +20,41 @@
...
-};
-
-TabRipple.defaultProps = {
-  className: '',
-  initRipple: () => {},
-  unbounded: false,
-};
-
-export default withRipple(TabRipple);
+export default Ripple.withRipple<TabRippleProps, HTMLDivElement>(TabRippleBase);
```
appsforartists: You should just import `{withRipple}`.  There's no reason to name the package.

packages/tab/TabRipple.tsx
```diff
@@ -20,47 +20,41 @@
...
 // THE SOFTWARE.
 
-import React from 'react';
-import PropTypes from 'prop-types';
+import * as React from 'react';
 import classnames from 'classnames';
-import {withRipple} from '@material/react-ripple';
+import * as Ripple from '@material/react-ripple';
+
+export interface TabRippleProps extends React.HTMLProps<HTMLDivElement>, Ripple.InjectedProps<HTMLDivElement> {
```
appsforartists: I'm confused, because I don't see `InjectedProps` in `master`, but I don't see a PR for `typescript-ripple`.

packages/tab/index.tsx
```diff
@@ -0,0 +1,241 @@
...
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+import * as React from 'react';
+import * as classnames from 'classnames';
+import TabIndicator from '@material/react-tab-indicator';
+// No mdc .d.ts files
```
appsforartists: This comment should probably be on the `@ts-ignore` line.

packages/tab/index.tsx
```diff
@@ -0,0 +1,241 @@
...
+// @ts-ignore
+import {MDCTabFoundation} from '@material/tab/dist/mdc.tab';
+import TabRipple, {TabRippleProps} from './TabRipple';
+
+export {
```
appsforartists: By convention, I place the exports at the bottom of the file, since that's where I expect a reader to look for them.

packages/tab/index.tsx
```diff
@@ -0,0 +1,241 @@
...
+  TabRipple,
+  Tab,
+  TabRippleProps,
+};
+
+export interface TabProps extends React.HTMLProps<HTMLButtonElement> {
+  active?: boolean;
+  className?: string;
```
appsforartists: Do you need to explicitly type `className` here, if you're extening `HTMLProps`?

packages/tab/index.tsx
```diff
@@ -0,0 +1,241 @@
...
+  isFadingIndicator?: boolean;
+  indicatorContent?: React.ReactNode;
+  minWidth?: boolean;
+  isMinWidthIndicator?: boolean;
+  stacked?: boolean;
+  previousIndicatorClientRect?: ClientRect;
+}
+
+interface TabAttributes {
```
appsforartists: I know I helped you refactor this, but one problem with this structure is that someone might reasonably expect to augment this with other attributes (e.g. `className`), even though they don't belong in `state`.  I wonder what a more specific name would be.

packages/tab/index.tsx
```diff
@@ -0,0 +1,241 @@
...
+
+interface TabState extends TabAttributes {
+  classList: Set<string>;
+  activateIndicator: boolean;
+  previousIndicatorClientRect?: ClientRect;
+}
+
+export default class Tab extends React.Component<TabProps, TabState> {
+  foundation?: MDCTabFoundation;
+  tabElement: React.RefObject<HTMLButtonElement> = React.createRef();
```
appsforartists: These aren't elements; they are refs, so I would name them accordingly:

```typescript
tabRef,
tabContentRef,
tabIndicatorRef,
tabRippleRef,
```

packages/tab/index.tsx
```diff
@@ -0,0 +1,241 @@
...
+  get classes() {
+    const {classList} = this.state;
+    const {className, minWidth, stacked} = this.props;
+    return classnames('mdc-tab', Array.from(classList), className, {
+      'mdc-tab--min-width': minWidth,
+      'mdc-tab--stacked': stacked,
+    });
+  }
+
+  get adapter() {
```
appsforartists: I presume you've made this a getter instead of a property because execution order matters.  That's implicit and not obvious to a casual observer.

I'd recommend instead to pass this literal into the `new MDCTabFoundation({ })` call in `componentDidMount`, since I don't see `this.adapter` used anywhere else.

packages/tab/index.tsx
```diff
@@ -0,0 +1,241 @@
...
+    return {
+      addClass: (className: string) => {
```
appsforartists: If you locally alias these:

```typescript
const {
  state,
  setState,
  tabRef,
  tabContentRef,
} = this;
```

you can make these methods instead of arrow functions.  That would hem closer to the ts-style's recommendations that discourage arrow properties.

moog16: These actually need to access `this.state` every time the method is called. Creating a variable `state` is a reference to state at that moment in time.

PR #535, opened 2018-12-17
feat(tab-bar): typescript

moog16:
related: #48


packages/tab-bar/index.tsx
```diff
@@ -0,0 +1,193 @@
...
+export interface TabBarProps extends React.HTMLProps<HTMLDivElement> {
+  indexInView?: number;
+  activeIndex: number;
+  handleActiveIndexUpdate?: (index: number) => void;
+  className?: string;
+  isRtl?: boolean;
+};
+
+interface TabBarState {
+  previousActiveIndex: any;
```
patrickrodee: Can we make this more strict than `any`?

packages/tab-bar/index.tsx
```diff
@@ -0,0 +1,193 @@
...
+};
+
+export default class TabBar extends React.Component<
+  TabBarProps, TabBarState
+  > {
+  tabBarElement_: React.RefObject<HTMLDivElement> = React.createRef();
+  tabScroller_: React.RefObject<TabScroller> = React.createRef();
+  tabList_: Tab[] = [];
+  foundation_: MDCTabBarFoundation | null = null;
```
patrickrodee: Can this be optional instead of nullable?

packages/tab-bar/index.tsx
```diff
@@ -0,0 +1,193 @@
...
+};
+
+interface TabBarState {
+  previousActiveIndex: any;
+};
+
+export default class TabBar extends React.Component<
+  TabBarProps, TabBarState
+  > {
+  tabBarElement_: React.RefObject<HTMLDivElement> = React.createRef();
```
patrickrodee: Can these be private?

packages/tab-bar/index.tsx
moog16: ah good catch

packages/tab-bar/index.tsx
moog16: certainly!

packages/tab-bar/index.tsx
moog16: This is also referenced in `/test/unit/tab-bar/index.test.tsx`.

PR #540, opened 2018-12-17
feat(select): typescript 

moog16:
related: #48


packages/select/index.tsx
```diff
@@ -19,48 +19,83 @@
...
+  className: string;
+  disabled: boolean;
+  floatingLabelClassName?: string;
+  isRtl: boolean;
+  label: string;
+  lineRippleClassName: string;
+  nativeControlClassName: string;
+  notchedOutlineClassName: string;
+  outlined: boolean;
+  options: (string | React.HTMLProps<HTMLOptionElement>)[];
```
clocksmith: when would options be a string vs an element? Can it be restricted to make its handling easier?

packages/select/index.tsx
```diff
@@ -96,60 +129,57 @@ export default class Select extends React.Component {
...
     };
-
     const notchedOutlineAdapter = {
       notchOutline: () => this.setState({outlineIsNotched: true}),
       closeOutline: () => this.setState({outlineIsNotched: false}),
       hasOutline: () => !!this.props.outlined,
     };
-
-    return Object.assign({},
+    return Object.assign(
```
clocksmith: spread?

test/screenshot/select/index.tsx
```diff
@@ -37,23 +62,9 @@ class SelectTest extends React.Component {
...
-  {isRtl: true},
-];
-
-const disabledMap = [
-  {},
-  {value: 'pomsky'},
-  {disabled: true},
-];
-
+const variants = [{}, {box: true}, {outlined: true}];
```
clocksmith: This is a bit hard to follow. 

The first 3 arrays of objects, as is, are not maps themselves. Should they be renamed?
Where is variantKey coming from? 
Are any of the values being used (e.g. true or 'pomsky'?)

packages/select/index.tsx
moog16: If people just want to do 
```
options = ['poodle', 'golden_retreiver', 'alaskan']

or

options = [{
  label: 'Poodle', value: 'poodle'
}, {
 label: 'Golden Retriever', value: 'golden_retriever'
}, ...]
```

Any suggestions? If we changed it, it would be a breaking change

test/screenshot/select/index.tsx
moog16: ya it is a bit hacked together. I wanted to programatically add:

```
<Select box ... />
<Select outlined ... />
```

in an array format.

moog16: variantKey is either `''`, `box`, or `outlined`

moog16: I updated the screenshot and moved the {value: pomsky} out, which in turn created a different screenshot. But I the `maps` make more sense.

PR #547, opened 2018-12-19
feat(infrastructure): build JS from TS

moog16:
Builds TSX files into ES5 and .d.ts typings files.


appsforartists:
It smells that these files are being imported from a particular path in the file system (`/dist/…`).  I see that there in a similar pattern already committed, but why can't you `import {MDCFloatingLabelFoundation} from '@material/floating-label`?

moog16:
This is a pattern is going to be fixed in a later release. MDC Web started pointing to the /dist/... files after 0.42.0. Currently React is pointing at 0.41.1. We point to the /dist/... because we don't transpile the node_modules, which can lead to unintended consequences if people are expecting ES5

moog16:
tests passing, CLA was already met. Merging

PR #548, opened 2018-12-19
feat(top-app-bar): typescript

moog16:
related #48


packages/top-app-bar/FixedAdjust.tsx
```diff
@@ -19,22 +19,28 @@
...
-    prominent,
-    short,
-    ...otherProps
-  } = props;
+export interface TopAppbarFixedAdjustProps {
+  tag?: string;
+  className?: string;
+  dense?: boolean;
+  prominent?: boolean;
+  short?: boolean
```
clocksmith: ;

packages/top-app-bar/index.tsx
```diff
@@ -19,28 +19,60 @@
...
   MDCTopAppBarFoundation,
   MDCShortTopAppBarFoundation,
+// no mdc .d.ts file
+// @ts-ignore
 } from '@material/top-app-bar/dist/mdc.topAppBar';
 
-export default class TopAppBar extends React.Component {
-  foundation_ = null;
+export interface TopAppBarProps {
+  actionItems?: React.ReactElement<any>[];
```
clocksmith: Usually any is not allowed when using --noImplicitAny, which google3 uses. Is there any reason the generic type cannot be specified ahead of time?

packages/top-app-bar/index.tsx
```diff
@@ -19,28 +19,60 @@
...
+type VariantType = 'dense' | 'fixed' | 'prominent' | 'short' | 'shortCollapsed';
 
-  state = {
+export default class TopAppBar extends React.Component<
+  Props,
+  TopAppBarState
+  > {
+  topAppBarElement: React.RefObject<HTMLElement> = React.createRef();
+  foundation_?: MDCFixedTopAppBarFoundation | MDCTopAppBarFoundation | MDCShortTopAppBarFoundation;
+
```
clocksmith: Is this union useful for a 3p developer? If so, consider making a type for MDCTopAppBarFoundation which is the union of these 3

test/screenshot/top-app-bar/prominentToShortCollapsed.tsx
```diff
@@ -12,7 +21,7 @@ class TopAppBarProminentToShortCollapsedScreenshotTest extends React.Component {
     window.addEventListener('resize', this.updateTopAppBarVariant);
   }
 
-  shouldComponentUpdate(_, nextState) {
+  shouldComponentUpdate(_: {}, nextState: TopAppBarProminentToShortCollapsedScreenshotTestState) {
```
clocksmith: Consider @override annotations (for readability)

packages/top-app-bar/index.tsx
moog16: I think --noImplicitAny will allow you to imply any types, but defining it as `any` is ok. However it is the lazy way out. I think in this case, it'll be an <i> or <svg> or .

test/screenshot/top-app-bar/mainContent.tsx
```diff
@@ -0,0 +1,22 @@
+import * as React from 'react';
+import {
+  TopAppBarFixedAdjust,
+  TopAppbarFixedAdjustProps, // eslint-disable-line no-unused-vars
+} from '../../../packages/top-app-bar';
+
+const MainTopAppBarContent: React.FunctionComponent<TopAppbarFixedAdjustProps> = (props) => {
+  return (
```
clocksmith: nit: dont need parens

PR #571, opened 2018-12-28
fix(infrastructure): condense webpack configuration build step into one json file 

moog16:
The change to ts-loader increased the amount of memory the webpack config needed. Changing the webpack config to use the multiple entries pattern allows webpack to perform a lot faster.

BREAKING CHANGE: Remove `<package-name>.es.js` files. The alternate is now to use the `index.tsx` files.

### DIFF
--- a/packages/webpack.config.js
+++ b/packages/webpack.config.js
@@ -44,30 +44,34 @@ const getAbsolutePath = (url) => path.resolve(__dirname, url);
 const filename = '[name]';
 
 function getWebpackConfigs() {
-  const webpackConfigs = [];
+  const webpackEntries = {};
+  const cssWebpackEntries = {};
+  const cssWebpackEntriesMin = {};
+  const webpackConfig = getJavaScriptWebpackConfig();
+  const webpackConfigCss = getCssWebpackConfig();
+  const webpackConfigCssMin = getCssWebpackConfig(true);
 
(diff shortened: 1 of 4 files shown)


packages/webpack.config.js
```diff
@@ -97,21 +101,36 @@ function getMaterialExternals() {
...
+          '@material/textfield/constants': `@material/textfield/constants.js`,
         },
-        getMaterialExternals(),
+        materialExternals,
       ),
       module: {
         rules: [{
           test: /\.ts(x)?$/,
           loader: 'ts-loader',
+        }, {
```
moog16: this was here before, but i realized we need it back, since we do still reference .js files

tsconfig.json
```diff
@@ -17,10 +16,5 @@
   "exclude": [
     "node_modules",
     "build"
-  ],
-  "baseUrl": ".",
```
moog16: the repo still works without this

tsconfig.json
```diff
@@ -2,7 +2,6 @@
   "extends": "gts/tsconfig-google.json",
   "compilerOptions": {
     "outDir": "./types/",
-    "declaration": true,
```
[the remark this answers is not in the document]
moog16: Yes but its included in gts/tsconfig-google

Issue #568, opened 2018-12-28
remove prop-types packages


moog16:
prop-types is no longer needed in any of the packages. Issue is to remove all prop-types.

Issue #569, opened 2018-12-28
add @types/react to packages


moog16:
according to this doc https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html, we should add the @types/package-name to the package.json deps.

PR #573, opened 2018-12-28
fix: update package.json for all typescript components

moog16:
fixes https://github.com/material-components/material-components-web-react/issues/568 and https://github.com/material-components/material-components-web-react/issues/569


moog16:
closing due to conflicts with rc0.8.0

PR #584, opened 2019-01-02
Support null as a child element for ListItem

gugu:
Last few years React supports null as a valid child (it just ignores this element). Used mostly for conditional statements. This pull request makes ListItem support nulls


moog16:
@gugu could you please provide your use case? I see what you're saying, but there might be a better way of doing what you're trying to do. 

Also this change requires a unit test. Thanks!

gugu:
Here is an example:

```jsx
return <AppListItem key={domain.id}>
    <DomainNameText primaryText={ domain.hostname } />
    { this.props.editPermission ? <DomainListActions>
        <AppIconButton onClick={this.handleDomainEdit(domain)} type="button"><MaterialIcon icon="edit" /></AppIconButton>
        <AppIconButton onClick={this.handleDeleteDomain(domain)} type="button"><MaterialIcon icon="delete" /></AppIconButton>
    </DomainListActions> : null} 
</AppListItem>;})
```

Here is how I solved the problem (creates unnecessary DOM node):

```jsx
return <AppListItem key={domain.id}>
    <DomainNameText primaryText={ domain.hostname } />
    { this.props.editPermission ? <DomainListActions>
        <AppIconButton onClick={this.handleDomainEdit(domain)} type="button"><MaterialIcon icon="edit" /></AppIconButton>
        <AppIconButton onClick={this.handleDeleteDomain(domain)} type="button"><MaterialIcon icon="delete" /></AppIconButton>
    </DomainListActions> : <div />} 
</AppListItem>;})
```

moog16:
I see...ya that seems to be the easiest way. I was thought the best way to render list items were iterating over an array...which you could use .filter to hide a ListItem. But in the case you have one list item this makes the most sense.

Can you add a unit test for this case? I will approve it then. Thanks for providing the example.

gugu:
Done

moog16:
@gugu thanks! I'm reviewing right now. Could you also sign the PR?

gugu:
One moment, I slightly screwed up your merge with `git push -f`, so I did it by myself

moog16:
Ah thanks...Although I just need you to comment "I signed it" por favor

gugu:
I signed it

PR #586, opened 2019-01-02
docs(checkbox): fix example

moog16:
fixes broken readme example in checkbox


bruqui:
So my comment discord wasn't just for Checkbox, same goes for...

moog16:
@bruqui thanks! Please let me know of any others you find :)

bruqui:
I took a quick peek around and I think the rest are OK.

Issue #588, opened 2019-01-02
Create SSR Tests


moog16:
Many issues have been arising due to some components not being compatible with SSR (server side rendering). We should have some tests to avoid this issue. This would also put us up to par with https://github.com/jamesmfriedman/rmwc who has had SSR tests in place.

4cm4k1:
As I said in #581, I’ll check out RMWC’s testing setup for SSR use cases and report back on it. I don’t have a definite ETA, however. 😄

lucasecdb:
any progress on this? I can't update my packages to versions above v0.7.x because my SSR breaks :-(

let me know if I can help with anything

moog16:
@lucasecdb nothing has been started on this yet. If you'd like to propose something I'm all ears. I currently am not 100% committed to working on this project and am just acting as the maintainer. My plan is to hopefully get back to it in Q2/Q3.

lucasecdb:
@moog16 I have a really bold proposal, to be honest. I see that the majority of the React community and developers like to work more with [writing tests in jest](https://labs.contactually.com/migrating-2-000-specs-from-karma-to-jest-25dd8b0f3cfb) than any other test runner/framework. And it really makes sense when you apply the utilities that come builtin and snapshotting tests that compare what DOM elements the component is "spitting" and *also* run in node, which counts for those SSR tests we want.

The good think about the snapshots is because we could easily compare them with the required DOM structure that the mdc web core module requires, without needing to run any browser and inspecting the elements panel on dev tools.

There are currently a lot of codemods out there which can migrate most of the test suite really quickly, and the ones that are left out can be easily migrated by hand, since there won't be a lot of them. I don't know if you guys already considered rewriting the unit test suite in jest, but I would be willing to work on this migration if you accept this proposal.

moog16:
@lucasecdb I've never used Jest, and I know that the majority of the React community uses it. I understand there are a lot of tools and features that we could use. I am hesitant to switch testing frameworks since it is such a big change. There is also the integration with Travis and the release that would need to be addressed. Can you write out all the different tasks that would be involved? 

Would you be replacing Karma for Jest? I'm not sure how it fits in the picture and haven't looked into the tool. 

TLDR; I'm open to the idea, but need more information.

lucasecdb:
> Would you be replacing Karma for Jest?

[...]

PR #592, opened 2019-01-04
fix(ripple): Adds HTMLElement shim to fix regression for SSR projects

### LINKED ISSUE #581: ripple: HTMLElement is not defined on SSR
Hello,

With the changeover to TS in `v0.8.0`, the following [line](https://github.com/material-components/material-components-web-react/blob/v0.7.1/packages/ripple/index.js#L65) went from being scoped within a function to [not](https://github.com/material-components/material-components-web-react/blob/v0.8.0/packages/ripple/index.tsx#L29):
```js
const MATCHES = util.getMatchesProperty(HTMLElement.prototype);
```

In SSR React apps, this will result in an `HTMLElement is not defined` error, because it's not evaluated in a browser where `HTMLElement` would exist.

The following line, when added just before the line above, resolves the issue locally:

```ts
const HTMLElement = typeof HTMLElement === 'undefined' ? function () {} : HTMLElement;
```

However, I'm not sure this is an appropriate solution outside of my use case. If this addition would suffice, I can submit a PR for it. Let me know! Thanks! 😄

4cm4k1:
This PR fixes #581.

I did not write any tests, as writing tests for SSR use cases is under consideration at #588.

The compiled result of the changed code is as follows, which should fix the regression seen in SSR projects:
```js
var HTMLElementShim = typeof HTMLElement === 'undefined' ? { } : HTMLElement;
var MATCHES = mdc_ripple_1.util.getMatchesProperty(HTMLElementShim.prototype);
```

I’m relatively new to writing TypeScript, so please let me know if `any` as a type should be changed.

Warm regards,
Anthony

### DIFF
--- a/packages/ripple/index.tsx
+++ b/packages/ripple/index.tsx
@@ -26,7 +26,7 @@ import {Subtract} from 'utility-types'; // eslint-disable-line no-unused-vars
 // @ts-ignore no mdc .d.ts file
 import {MDCRippleFoundation, MDCRippleAdapter, util} from '@material/ripple/dist/mdc.ripple';
 
-const HTMLElementShim: any = typeof HTMLElement === 'undefined' ? () => {} : HTMLElement;
+const HTMLElementShim: any = typeof HTMLElement === 'undefined' ? {} : HTMLElement;
 const MATCHES = util.getMatchesProperty(HTMLElementShim.prototype);
 
 export interface RippledComponentProps<T> {
--- a/packages/ripple/index.tsx
+++ b/packages/ripple/index.tsx
@@ -26,7 +26,8 @@ import {Subtract} from 'utility-types'; // eslint-disable-line no-unused-vars
 // @ts-ignore no mdc .d.ts file
 import {MDCRippleFoundation, MDCRippleAdapter, util} from '@material/ripple/dist/mdc.ripple';
 
-const MATCHES = util.getMatchesProperty(HTMLElement.prototype);
+const HTMLElementShim: any = typeof HTMLElement === 'undefined' ? () => {} : HTMLElement;
+const MATCHES = util.getMatchesProperty(HTMLElementShim.prototype);
 
 export interface RippledComponentProps<T> {
   unbounded?: boolean;


packages/ripple/index.tsx
```diff
@@ -26,7 +26,8 @@ import {Subtract} from 'utility-types'; // eslint-disable-line no-unused-vars
 // @ts-ignore no mdc .d.ts file
 import {MDCRippleFoundation, MDCRippleAdapter, util} from '@material/ripple/dist/mdc.ripple';
 
-const MATCHES = util.getMatchesProperty(HTMLElement.prototype);
+const HTMLElementShim: any = typeof HTMLElement === 'undefined' ? () => {} : HTMLElement;
```
moog16: Shouldn't the default be an object instead of a function?
```
const HTMLElementShim: any = typeof HTMLElement === 'undefined' ? {} : HTMLElement;
```

moog16:
This looks like it fixed it, but I left one comment. I tested by using the repo linked in this issue https://github.com/material-components/material-components-web-react/issues/492. I actually found another similar issue using Drawer while testing. 

```
ReferenceError: Element is not defined
    at Object.105 (/usr/local/google/home/mattgoo/Downloads/material-next-template-master/node_modules/@material/drawer/dist/mdc.drawer.js:809:15)
    at __webpack_require__ (/usr/local/google/home/mattgoo/Downloads/material-next-template-master/node_modules/@material/drawer/dist/mdc.drawer.js:35:30)
```

I will open a separate issue.

packages/ripple/index.tsx
4cm4k1: Sure, can you explain why for me? I don’t necessarily understand what the difference is. I’ll make the change real quick. 👍

4cm4k1: Done.

moog16: On the next line you reference HTMLElementShim.prototype. That means that the HTMLElementShim should be an object - not a function.

4cm4k1: Thank you!

moog16:
@4cm4k1 could you please sign the pr

moog16:
tests passing: https://github.com/material-components/material-components-web-react/pull/595

4cm4k1:
I signed it.

moog16:
Thanks! merging :)

PR #600, opened 2019-01-08
fix(text field): adds missing adapter method `notifyIconAction` on Icon 

### LINKED ISSUE #585: TextField/Icon missing adapter methods `notifyIconAction`,`registerInteractionHandler`, and `deregisterInteractionHandler`
Textfield/Icon is missing adapter methods `notifyIconAction`, `registerInteractionHandler`, and `deregisterInteractionHandler`. At the moment the latter two  are there to determine if the Icon has been activated from click or keydown events. 

I also note that it appears the material-components-web project has been removing de/registerInteractionHandlers in a lot of their adapters. With the preference to call register eventHandlers in `initialSyncWithDOM()`. This makes me wonder if these may also be removed in a future release.  Therefore, might only `notifyIconAction` be implemented?

mgr34:
Implements missing icon adapter method `notifyIconAction` on `<Icon/>`. Accomplished via adding optional `onSelect` to `<Icon/>` and corresponding `onLeadingIconSelect` and `onTrailingIconSelect` on `<TextField/>`. Still having trouble with screenshot tests. Should close #585. 

Note: this is the first I've worked with typescript. 

---

I signed it

### DIFF
--- a/packages/text-field/icon/README.md
+++ b/packages/text-field/icon/README.md
@@ -23,7 +23,8 @@ Prop Name | Type | Description
 disabled | Boolean | Toggles the disabled state of the icon.
 children | Element | Required. Expects a single child icon element.
 onSelect | Function() => void | Optional callback for user interaction with icon
-> Note: `onSelect`  fired only for clicked event or when keydown event is enter key.
+> Notes: `onSelect`  fired only for clicked event or when keydown event is enter key.
+> `onSelect` will add tabindex of 0 if tabindex is not previously added to icon
 
 ## Icon
 
--- a/packages/text-field/icon/README.md
+++ b/packages/text-field/icon/README.md
@@ -23,7 +23,7 @@ Prop Name | Type | Description
 disabled | Boolean | Toggles the disabled state of the icon.
 children | Element | Required. Expects a single child icon element.
 onSelect | Function() => void | Optional callback for user interaction with icon
-> Notes: `onSelect`  fired only for clicked event or when keydown event is enter key.
+> Notes: `onSelect`  fired on click event and "Enter key" keydown event.
 > `onSelect` will add tabindex of 0 if tabindex is not previously added to icon
 
 ## Icon
--- a/packages/text-field/icon/README.md
+++ b/packages/text-field/icon/README.md
@@ -22,6 +22,8 @@ Prop Name | Type | Description
 --- | --- | ---
 disabled | Boolean | Toggles the disabled state of the icon.
 children | Element | Required. Expects a single child icon element.
+onSelect | Function() => void | Optional callback for user interaction with icon
+> Note: `onSelect`  fired only for clicked event or when keydown event is enter key.
 
 ## Icon
 
--- a/packages/text-field/icon/index.tsx
+++ b/packages/text-field/icon/index.tsx
@@ -27,6 +27,7 @@ import {MDCTextFieldIconFoundation} from '@material/textfield/dist/mdc.textfield
 export interface IconProps extends React.HTMLProps<HTMLOrSVGElement> {
   disabled: boolean;
   children: React.ReactElement<React.HTMLProps<HTMLOrSVGElement>>;
+  onSelect?: () => void;
 };
 
 interface IconState {
@@ -42,6 +43,7 @@ export default class Icon extends React.Component<
 
   static defaultProps: Partial<IconProps> = {
     disabled: false,
+    onSelect: () => {},
   };
 
   constructor(props: IconProps) {
@@ -94,8 +96,15 @@ export default class Icon extends React.Component<
       removeAttr: (attr: keyof IconState) => (
         this.setState((prevState) => ({...prevState, [attr]: null}))
       ),
+      notifyIconAction: () => this.props.onSelect!(),
     };
   }
+
+  handleClick = (e: React.MouseEvent<HTMLElement>) =>
+    this.foundation_.handleInteraction(e);
+
+  handleKeyDown = (e: React.KeyboardEvent<HTMLElement>) =>
+    this.foundation_.handleInteraction(e)
 
   addIconAttrsToChildren = () => {
     const {tabindex: tabIndex, role} = this.state;
@@ -103,6 +112,8 @@ export default class Icon extends React.Component<
     const className = classnames('mdc-text-field__icon', child.props.className);
     const props = Object.assign({}, child.props, {
       className,
+      onClick: this.handleClick,
+      onKeyDown: this.handleKeyDown,
       tabIndex,
       role,
     });
--- a/packages/text-field/icon/index.tsx
+++ b/packages/text-field/icon/index.tsx
@@ -25,7 +25,7 @@ import * as classnames from 'classnames';
 import {MDCTextFieldIconFoundation} from '@material/textfield/dist/mdc.textfield';
 
 export interface IconProps extends React.HTMLProps<HTMLOrSVGElement> {
-  disabled: boolean;
+  disabled?: boolean;
   children: React.ReactElement<React.HTMLProps<HTMLOrSVGElement>>;
   onSelect?: () => void;
 };
--- a/packages/text-field/icon/index.tsx
+++ b/packages/text-field/icon/index.tsx
@@ -108,8 +108,8 @@ export default class Icon extends React.Component<
       removeAttr: (attr: keyof IconState) => (
         this.setState((prevState) => ({...prevState, [attr]: null}))
       ),
-      notifyIconAction: () => ( this.props.hasOwnProperty('onSelect')
-        ? this.props.onSelect!()
+      notifyIconAction: () => ( this.props.onSelect
+        ? this.props.onSelect()
         : null
       ),
     };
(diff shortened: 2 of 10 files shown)


packages/text-field/icon/index.tsx
```diff
@@ -94,15 +108,27 @@ export default class Icon extends React.Component<
       removeAttr: (attr: keyof IconState) => (
         this.setState((prevState) => ({...prevState, [attr]: null}))
       ),
+      notifyIconAction: () => ( this.props.hasOwnProperty('onSelect')
+        ? this.props.onSelect!()
```
moog16: First time I'm doing a review on my phone... Do you have a default prop set for onSelect? We can get rid of the hasOwnProperty if you set one to an empty function

packages/text-field/icon/index.tsx
```diff
@@ -77,6 +81,16 @@ export default class Icon extends React.Component<
     }
   }
 
+  get tabindex() {
+    // if tabIndex is not set onSelect will never fire.
+    // note that foundation.js alters tabindex not tabIndex
+    if (this.props.children.props.hasOwnProperty('tabIndex')) {
```
moog16: Instead of hasOwnProperty let's check if it is a number

packages/text-field/icon/index.tsx
```diff
@@ -77,6 +81,16 @@ export default class Icon extends React.Component<
...
+      return this.props.children.props.tabIndex;
+    }
+
+    return this.props.hasOwnProperty('onSelect') ? 0 : -1;
```
moog16: We don't need to use hasOwnProperty, but instead just do this.props.onSelect

test/screenshot/text-field/icon/index.tsx
```diff
@@ -15,6 +15,10 @@ const TextFieldIconScreenshotTest = () => {
       <Icon>
         <MaterialIcon tabIndex={0} role='button' icon='favorite' />
       </Icon>
+
```
moog16: Looks like you added a screenshot test! I'll update once the pr looks good. Thanks!

test/screenshot/text-field/TextFieldPermutations.tsx
```diff
@@ -13,7 +13,11 @@ import {
 import TestField from './TestTextField';
 
 const textFields = (variantProps: {variant?: string}) => {
-  return iconsMap.map((icon: {leadingIcon?: React.ReactNode, trailingIcon?: React.ReactNode}) => {
+  return iconsMap.map((icon: {
+    leadingIcon?: React.ReactNode,
```
moog16: something i notice about these tests is that the icon tap target is only 24px x 24px. HIG (human interface guide - I think) says 48x48 is the minimum. I'm not sure what MDC Web does or if this is an issue with MDC Web. I'll follow up with one of our engineers.

packages/text-field/icon/index.tsx
mgr34: I think this might be an appropriate use for `hasOwnProperty`. Shortly I will be pushing a change where I eliminate it from the get tabidex(). 

I consciously chose against the approach of making it a defaultProp. Coincidentally, that is how I first went about it [0]. However, I ran into a problem where onSelect will only fire if tabindex != -1 due to pointer-events being set to none [1]. Since tabindex was not a required property for the passed icon I thought it more straightforward to add it based on if onSelect is present. If its has a default prop it will always be present and tabindex will always be initialized with 0 unless specified. This was the major factor in my decision to drop it as a default property.

I also seem to recall an issue in renderIcon from text-field where It was always passing an onSelect. Alas, I cannot quite recall the issue there and it was never committed. Maybe I was also setting a default prop for onLeadingIconSelect/onTrailingIconSelect?

Take your time on responding. A code review on a phone sounds painful.
 

[0] - https://github.com/material-components/material-components-web-react/pull/600/commits/df8627c239b047c58a3353743b2838369c689516#diff-78f850e429f25a98280e040459f8b54f

[1] https://github.com/material-components/material-components-web/blob/3cc4c6aa48ae8c1a307df542911e28a7808de41f/packages/mdc-textfield/icon/mdc-text-field-icon.scss#L33-L37

test/screenshot/text-field/TextFieldPermutations.tsx
```diff
@@ -13,7 +13,11 @@ import {
 import TestField from './TestTextField';
 
 const textFields = (variantProps: {variant?: string}) => {
-  return iconsMap.map((icon: {leadingIcon?: React.ReactNode, trailingIcon?: React.ReactNode}) => {
+  return iconsMap.map((icon: {
```
moog16: also getting an error when click on the leading-trailing-icon permutation. Try clicking on one of the icons here:
tndrewahb7u

They also should not have `tabIndex` at all. It should be undefined.

packages/text-field/icon/index.tsx
```diff
@@ -27,6 +27,7 @@ import {MDCTextFieldIconFoundation} from '@material/textfield/dist/mdc.textfield
 export interface IconProps extends React.HTMLProps<HTMLOrSVGElement> {
   disabled: boolean;
   children: React.ReactElement<React.HTMLProps<HTMLOrSVGElement>>;
+  onSelect?: () => void;
```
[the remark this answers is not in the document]
mgr34: can do.

packages/text-field/index.tsx
```diff
@@ -45,6 +45,8 @@ export interface Props<T> {
   lineRippleClassName: string;
   notchedOutlineClassName: string;
   outlined: boolean;
+  onLeadingIconSelect?: () => void;
```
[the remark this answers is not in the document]
mgr34: See my previous response about why I did not include a default prop for onSelect. This would hoist the problem. Now renderIcon will always render with an onSelect and would be problematic for setting the tabindex.

packages/text-field/icon/index.tsx
moog16: That makes sense about the defaultProp...but we don't use hasOwnProperty anywhere else in the library, which makes me suspicious. Either we should be doing that everywhere, or we shouldn't do it here - it just breaks the pattern. If we decide to do it, we should instead open another bug to do it system wide.

packages/text-field/index.tsx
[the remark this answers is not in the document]
moog16: sounds good!

packages/text-field/icon/index.tsx
mgr34: I hear you on  that mindset....I guess I could do something like the following. 

```typescript
{ notifyIconAction: (() =>  this.props.onSelect ? this.props.onSelect() : null) }
```

for some reason I seem to rely heavily on hasOwnProperty() in other projects. Old habits and all... Nevertheless, making this exact change doesn't seem to cause any breaking changes. Will this work for you?

moog16: looks good to me!

test/screenshot/text-field/TextFieldPermutations.tsx
moog16: cool - i think one of your other commits must have fixed it

packages/text-field/icon/README.md
```diff
@@ -22,6 +22,9 @@ Prop Name | Type | Description
 --- | --- | ---
 disabled | Boolean | Toggles the disabled state of the icon.
 children | Element | Required. Expects a single child icon element.
+onSelect | Function() => void | Optional callback for user interaction with icon
+> Notes: `onSelect`  fired only for clicked event or when keydown event is enter key.
```
moog16: wording: `fired on click event and "enter key" keydown event.

test/unit/text-field/index.test.tsx
```diff
@@ -478,6 +517,34 @@ test('does not render trailingIcon or trailingIcon if no prop is passed', () =>
   assert.equal(wrapper.find('.test-class-name-icon').length, 0);
 });
 
+test('onTrailingIconSelect is passed to trailingIcion if passed as prop', () => {
```
moog16: trailingIcon spelling mistake

test/unit/text-field/index.test.tsx
```diff
@@ -478,6 +517,34 @@ test('does not render trailingIcon or trailingIcon if no prop is passed', () =>
...
+      trailingIcon={<i className='test-class-name-icon' />}
+    ><Input /></TextField>
+  );
+
+  const trailingIcon = wrapper.find('.test-class-name-icon').parent().props();
+  assert.isFunction(trailingIcon.onSelect);
+  assert.strictEqual(trailingIcon.onSelect, onSelect);
+});
+
+test('onTrailingIconSelect is not passed to trailingIcion if not passed as prop', () => {
```
moog16: Icon spelling

moog16:
@mgr34 please update text-field/icon golden.json to `0bbc8c762d27071e55983e5742548d166864b6fcebc0b9f1e413523fb93b7075`

test/unit/text-field/index.test.tsx
mgr34: fixed!

moog16:
tests are passing and pr is signed in https://github.com/material-components/material-components-web-react/pull/600#issue-243022107. Merging!

Issue #500, opened 2018-12-07
Menu surface are not removed from DOM on unmount


jfily:
Hi,

On mount, MenuSurface components are moved to \<body\> but are not removed from DOM on unmount.

moog16:
This is actually a duplicate of this: https://github.com/material-components/material-components-web-react/issues/455. Your issue is actually more accurate. I will close #455.

Also marking this help wanted if you'd like to open a PR.

moog16:
Closed with #693

PR #617, opened 2019-01-16
fix(menu-surface): remove hoisted element

aluminick:
Remove hoisted element on componentWillUnmount


moog16:
@aluminick - thanks for doing this! I will review in a bit.

fixes https://github.com/material-components/material-components-web-react/issues/500

moog16:
@aluminick it seems to be working and code looks good. Can you please add unit tests  to support your change?

aluminick:
@moog16 sure. I will add it later today. Thanks

aluminick:
I am working on the test but I came across these error `'Warning: render(...): It looks like the React-rendered content of this container was removed without
using React.`.

moog16:
is that during the test?

aluminick:
Yes it's during the test, it's when I unmount the component. I believe it's because react was not used when hoisted. Anyways, I already posted another issue about menu-surface hoisting. Thanks

PR #619, opened 2019-01-16
[Incomplete, preview] material dialog

gugu:
Hi

For my project, I need material-dialog, which was implemented internally. Now I want to backport it to react components.

Can you please check if it is the correct direction? If yes, I'll add missing adapter methods, update docs and tests

Currently it is used by my project with the name @shortcm/react-dialog and the module is public


gugu:
Nevermind, there is already a pull request

PR #622, opened 2019-01-17
MDC button trailing icon support

gugu:
Now MDC-web supports trailing icon

https://github.com/material-components/material-components-web/tree/master/packages/mdc-button#trailing-icon


packages/button/index.tsx
```diff
@@ -37,6 +37,7 @@ export interface ButtonProps<T extends ButtonTypes> extends Ripple.InjectedProps
   className?: string;
   icon?: React.ReactElement<React.HTMLProps<HTMLOrSVGElement>>;
   href?: string;
+  trailingIcon?: boolean,
```
moog16: what if instead you changed this to accept

```
trailingIcon: React.ReactElement<React.HTMLProps<HTMLOrSVGElement>>;
```

packages/button/index.tsx
```diff
@@ -74,16 +76,22 @@ export const Button = <T extends ButtonTypes>(
   if (href) {
     return (
       <a {...props as React.HTMLProps<HTMLAnchorElement>} href={href}>
-        {renderIcon(icon)}
-        {children}
+        {!trailingIcon ? renderIcon(icon) : null}
+        <span className="mdc-button__label">
+          {children}
+        </span>
+        {trailingIcon ? renderIcon(icon) : null}
```
moog16: it could then be renderIcon(trailingIcon)...does that make sense?

packages/button/package.json
```diff
@@ -17,7 +17,7 @@
     "url": "https://github.com/material-components/material-components-web-react.git"
   },
   "dependencies": {
-    "@material/button": "^0.41.0",
+    "@material/button": "^0.43.0",
```
moog16: this also needs to be updated in the main `./package.json`

gugu:
I made these changes and added some simple test. This morning I'll check if there are more tests needed. If yes - I will add them. And retest module in our application

packages/button/index.tsx
moog16: @gugu I meant for the type change to change this line into 
```
{renderIcon(trailingIcon)}
```

test/unit/button/index.test.tsx
```diff
@@ -24,6 +24,12 @@ test('renders an icon', () => {
   assert.isTrue(wrapper.find('.test-icon').hasClass('mdc-button__icon'));
 });
 
+test('renders a trailing icon', () => {
```
moog16: could you also add a test for when it is an anchor tag?

moog16:
Left some comments, but I think once those are addressed it should be enough.

gugu:
Done

moog16:
@gugu please update button golden to the same as this: https://github.com/material-components/material-components-web-react/pull/625

moog16:
https://github.com/material-components/material-components-web-react/pull/625 tests passing...looks good!

PR #630, opened 2019-01-23
docs(dialog): Example variable choice already existed, and 'c' no longer did.

mgr34:
Fix Dialog documentation. Example would not work properly because variable c had been renamed, but not changed in all places.Additionally, it was renamed to a variable that already existed.

> Note: I did not open an issue for this. I can if you prefer. 

---

I signed it


moog16:
merging!

Issue #648, opened 2019-01-31
(Chips) Required property id is marked as optional


mgr34:
In the README.md documentation for Chips it notes that ID is required. However, `ChipProps` indicates that id is optional. The documentation is mainly correct in this case. A unique identifier is necessary for filter and choice variants. I can submit a PR that requires an ID

mgr34:
#649

PR #649, opened 2019-01-31
fix(chips): id shuold be required prop

mgr34:
Documentation claims id should be required, but was an optional property. Documentation mainly correct. A unique identifier is necessary for filter and choice variants. This should close issue #648 .

---

I signed it


mgr34:
@moog16 The problem with throwing an error on componentDidMount is that the chips themselves don't know what sort of variant they are. This information is passed in the `<ChipSet/>` via the boolean props `choice|filter|input`.  I guess the best place to look for this would be in the `renderChip` method of ChipSet?

If that sounds correct to you I will go ahead and make that change.

moog16:
RE #1: I'm not sure if it is possible in TS, that's why I mentioned throwing an error. But then again this will be a breaking change for people if they have choice and didn't have `id`s, but like you mentioned it just wont work without it.

What if we pass down the variant to the Chip so that it can throw an error?

mgr34:
I am thinking the least intrusive change would be to simply catch it in ChipSet's renderChip method. From my understanding, if the Chip is not a part of ChipSet it never matters, nor does it need to know about its variant.

moog16:
Ah...ya that makes sense...let's do that!

mgr34:
ChipSet's renderChip method will now throw error if  ((choice || filter || input) && !chip.props.id).

I was a bit unsure about the unit tests (and spent way too much time on them). If you have a preferred solution they can be changed.

test/unit/chips/ChipSet.test.tsx
```diff
@@ -414,6 +414,31 @@ test('chip is rendered with computeBoundingRect method prop if is not filter var
   assert.equal(chip.props.computeBoundingRect, null);
 });
 
+test('basic variant ChipSet will not throw error if chip missing id', () => {
+  const stub = <ChipSet><Chip /></ChipSet>;
+  const render = () => shallow<ChipSet>(stub);
+  assert.isOk(render);
```
moog16: what about doesnotthrow. I think that is a method

moog16:
@mgr34 I like those tests, but looks like they are failing.

mgr34:
@moog16 Well if that isn't frustrating. They passed in my working copy, but sure enough when I check it out to clean dir it fails. Oddly enough it fails on things I haven't changed. For example here are the first two errors:

```
ERROR in ./packages/text-field/index.tsx                                                                                                                                                      
[tsl] ERROR in /home/matt/material-components-web-react/packages/text-field/index.tsx(352,11)                                                                                                 
      TS2322: Type 'ReactElement<Input<T>, string | ((props: any) => ReactElement<any, string | ... | (new (props: any) => Component<any, any, any>)> | null) | (new (props: any) => Component
<any, any, any>)> | ... 29 more ... | (ReactPortal & ... 1 more ... & ReactElement<...>)' is not assignable to type 'ReactElement<InputProps<T>, string | ((props: any) => ReactElement<any, s
tring | ... | (new (props: any) => Component<any, any, any>)> | null) | (new (props: any) => Component<any, any, any>)>'.                                                                     
  Type 'ReactElement<Input<T>, string | ((props: any) => ReactElement<any, string | ... | (new (props: any) => Component<any, any, any>)> | null) | (new (props: any) => Component<any, any, a
ny>)>' is not assignable to type 'ReactElement<InputProps<T>, string | ((props: any) => ReactElement<any, string | ... | (new (props: any) => Component<any, any, any>)> | null) | (new (props
: any) => Component<any, any, any>)>'.                                                                                                                                                        
    Type 'Input<T>' is missing the following properties from type 'InputProps<T>': className, inputType, disabled, handleValueChange, and 4 more.                                             
                                                                                                                                                                                              
ERROR in ./packages/tab-indicator/index.tsx                                                                                                                                                   
[tsl] ERROR in /home/matt/material-components-web-react/packages/tab-indicator/index.tsx(164,34)                                                                                              
      TS2533: Object is possibly 'null' or 'undefined'.   
```

I am going to have to investigate this a bit more. Unfortunately, I am not sure that I will be able to get to that today.

moog16:
That's odd...maybe if we just restart the tests

test/unit/chips/ChipSet.test.tsx
mgr34: good catch.

mgr34:
Wait, how would I restart the tests? 

to be clear I ran `npm run test:unit`  to receive the previous errors. 

edit: Ok, I checked out everything into yet a different clean directory, but first ran tests on rc0.10.0 and that worked.  I switched to this branch and I received the same errors Travis is reporting. So I guess I am onto something.  Not sure why they're failing exactly. (Investigating...)

moog16:
@mgr34 don't commit the package-lock. Only commit it if you added something to package.json, which you did not. 

Looks like there is an issue with screenshots, I'll try to get that fixed.

mgr34:
@moog16 - sounds good to me. I was a little surprised to see that it was modified.

moog16:
ya that happens a lot, especially if our environments are different (node version, npm version, etc).

mgr34:
makes sense. :)

Issue #650, opened 2019-01-31
(Chips): Chip properties handleSelect/handleRemove/handleInteraction are never called when Chip is child of ChipSet


mgr34:
`<ChipSet>` calls [renderChip](https://github.com/material-components/material-components-web-react/blob/788b426d66bdc5126a9d5448a047e2eb91f56e25/packages/chips/ChipSet.tsx#L162-L180) which overwrites the `handleSelect`, `handleInteraction`, and `handleRemove` properties of the `<Chip/>`. As a result they are never called.  ChipSet should either create a new function which calls both ChipSet's and Chip's `handle[Select|Interaction|Remove]` function. Or have ChipSet accept its own `handle[Select|Interaction|Remove]` function as a prop. 

My inclination is to pursue option one.  This may cause less breakage. The benefit of option two is that when writing `<Chip/>` a user wouldn't need to pass a handler to each chip, but just once on a `<ChipSet/>`. However, the ChipSet interface already has a handleSelect property with a different signature. A user would may need to refactor every usage of ChipSet. Whereas, having `renderChip` create a new function which not require any changes from a users perspective.

...also I have a PR ready to go for option one ;)

moog16:
Thanks for opening this issue and PR..continuing dialog in PR

PR #651, opened 2019-01-31
fix(chips): ChipSet no longer overwrites handle[Interaction|Select|Re…

mgr34:
ChipSet was overwriting handle[Interaction|Select|Remove] of Chip. Therefore, functions passed were never called. Successful merging may close issue #650 

---

I signed it


mgr34:
@moog16 thanks for the comments and the tests! I've merged them and push the additions :)

PR #665, opened 2019-02-08
Update README.md

ZeJ0hn:
The current code doesn't compile.
Need to use currentTarget instead of target (because target is generic now)

### DIFF
--- a/packages/drawer/README.md
+++ b/packages/drawer/README.md
@@ -212,7 +212,7 @@ export default class App extends Component {
       <div className='drawer-container'>
         <TopAppBar title='Inbox'/>
 
-        <TopAppBarFixedAdjust>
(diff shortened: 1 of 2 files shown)


ZeJ0hn:
I signed it!

packages/drawer/README.md
```diff
@@ -212,7 +212,7 @@ export default class App extends Component {
       <div className='drawer-container'>
         <TopAppBar title='Inbox'/>
 
-        <TopAppBarFixedAdjust>
+        <TopAppBarFixedAdjust className='top-app-bar-fix-adjust'>
```
moog16: revert everything in this file since you have it in #666

moog16:
please update

ZeJ0hn:
made a mistake. 
push again with #670

PR #667, opened 2019-02-09
WIP make material-components-web-react to use typescript mdc-web

### LINKED ISSUE #139: Convert to one package
- [ ] update webpack to build one package
- [ ] remove all package.json and .npmignore files from each individual package
- [ ] update packages to use relative paths instead of @material/react-foo packages
- [ ] update how to install documentation
- [ ] create a root index.js to export all packages

gugu:
Currently, I am surprised by the bundle size of the application, written with Material. For example, base Foundation in my app is imported 12 times, ripple 4 times

I'm trying to fix #139

- [x] The package is being built
- [x] Other modules started to bundle @material/ packages
- [x] MDC Components are not included in the bundle
- [ ] MDC base Foundation is still imported 22 times in this package, but it should be fixed as soon as mdc-web will be switched to typescript

Result


moog16:
Yes, I don't think mdc web had tree shaking working, which also resulted in a larger bundle size (duplicate code). I'm working on that right now. Hopefully this will decrease the size.

packages/chips/Chip.tsx
```diff
@@ -23,7 +23,7 @@ import * as React from 'react';
 import * as classnames from 'classnames';
 import * as Ripple from '@material/react-ripple';
 // @ts-ignore no mdc .d.ts file
-import {MDCChipFoundation} from '@material/chips/dist/mdc.chips';
+import {MDCChipFoundation} from '@material/chips/chip/foundation';
```
moog16: I'm not sure if we can make these changes. `@material/chips/chip/foundation` is all ES6 code. People targeting a browser that only supports ES5 might not be able to use this package anymore, since the build step does not transpile mdc web's chip/foundation.  

I think we'll need to wait until MDC Web typescript conversion is complete.

packages/material-components-web-react/package.json
```diff
@@ -0,0 +1,55 @@
+{
+  "name": "material-web-react",
+  "version": "0.9.4",
```
moog16: 0.0.0, since this is a new package.

packages/material-components-web-react/package.json
```diff
@@ -0,0 +1,55 @@
...
+    "@material/react-tab-scroller": "^0.8.0",
+    "@material/react-text-field": "^0.9.0",
+    "@material/react-top-app-bar": "^0.9.0",
+    "@material/react-typography": "^0.8.0"
+  },
+  "publishConfig": {
+    "access": "public"
+  },
+  "devDependencies": {
+    "material-components-web": "^0.44.0"
```
moog16: why do we need this? I dont see it being used anywhere.

package.json
```diff
@@ -136,6 +137,7 @@
     "karma-sourcemap-loader": "^0.3.7",
     "karma-webpack": "^2.0.13",
     "lerna": "^2.9.0",
+    "material-components-web": "^0.44.0",
```
moog16: why do we need this? I dont see it being used anywhere.

packages/webpack.config.js
```diff
@@ -87,16 +85,14 @@ function getReactMaterialExternals() {
   ));
 }
 
-function getMaterialExternals() {
+function getMaterialExternals(withReact) {
```
moog16: where is this withReact being passed as true/false? Is this change unintentional?

packages/chips/Chip.tsx
gugu: This pull request bundles @material/ dependencies into react material. It this case it will be transpiled to ES5

gugu:
The difference is that it uses bundling instead of installing dependencies. This allows using ES6 and TypeScript together.

I think it is better to wait for the first release of MDC in typescript and then recompile it with typescript and without bundling dependencies

moog16:
@gugu  are you saying you want to hold off merging this PR into the RC branch?

gugu:
To be honest I don't know what is correct. Currently, it changes the behaviour of the existing module - instead of depending on another module now it bundles it.

I have 12 modules installed in my project and in total, they are 3.5MB uncompressed. This one is 1.2 MB uncompressed for all modules included, definitely a win

It is up to you, what to do with this pull request - either merge, or we can wait for typescript implementation and I'll update this pull request accordingly. I think total size can be reduced at least twice with typescript

moog16:
Yes I understand the issue. We cannot merge what we have now, since there are some Object.assigns in the bundled (es5) code. 

I think I have solution around the big bundle size:

Right now we (MDC React) only offers UMD. This is not suitable for tree shaking (for people who use rollup and webpack). I think if we use `tsc` we could build and serve esmodules on npm. So your bundle could be 3.5MB, but if you only use drawer or text-field, you would only import that. The current MDC web PR I'm working will include these changes, which we can borrow in MDC React. If you are able to open that PR, which would basically be copying the MDC React PR - we could get this feature. How does that sound?

gugu:
I upgraded versions, will fix build (still keeping ts-ignore for @material/ deps)

gugu:
Too many changes for a single pull request

moog16:
@gugu I'm actually planning on updating to latest MDC Web. Are you currently working on something? Maybe we can split it up.

Issue #663, opened 2019-02-07
List with conditional false children error


madaz:
List component fails on false conditional `&&` children.

Overview

```
// where showDivider = false

<List>
  <ListItem>
     <ListItemText primaryText="Item 1" />
  </ListItem>
  { showDivider && <ListDivider /> }
</List>
```
Example https://codesandbox.io/s/2r86yyvpp

The  `list/index.tsx.isListItem` is not handling `null` element which is what is passed in when a false condition is triggered

```
function isListItem(element: any): element is ListItem {
  return element.type === ListItem;
}
```

Should be 

```
function isListItem(element: any): element is ListItem {
  return element && element.type === ListItem;
}
```

Work around is to create a custom If component e.g. `<If condition={showDivider} truthy={<ListDivider/>} />`

I would send a PR but struggling to get the project working on a Windows 10 development environment.

moog16:
Thanks for opening. Yup definitely a bug. Once we merge your current PR, we can hopefully open a PR to fix this.

moog16:
#674 fixes it

PR #674, opened 2019-02-13
fix(list): renders a list with children which is not DOM

TroyTae:
Now ```<List />```checks children's type using ```isListItem```.
If parameter is null/undefined, an error occurs.
So I add code that checks element's existence.

p.s. I add one dependency(@material/radio) because there is error "cannot find module" when running test code. If you do not need it, I will revert it.

### DIFF
--- a/test/unit/list/index.test.tsx
+++ b/test/unit/list/index.test.tsx
@@ -411,5 +411,5 @@ test('renders a list with children which is not DOM', () => {
     {null}
     {undefined}
   </List>);
-  assert.isNotEmpty(wrapper);
+  assert.isTrue(wrapper.children().length === 1);
 });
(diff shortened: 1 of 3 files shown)


TroyTae:
related to #663

TroyTae:
I signed it!

test/unit/list/index.test.tsx
```diff
@@ -402,3 +402,14 @@ test('renders a list with a nav tag', () => {
...
+
+test('renders a list with children which is not DOM', () => {
+  const wrapper = shallow(<List>
+    {}
+    {false}
+    {'text'}
+    {null}
+    {undefined}
+  </List>);
+  assert.isNotEmpty(wrapper);
```
moog16: I think it it more accurate to find how many listitem nodes would be contained in the wrapper. I think this the length should be 1, correct?

TroyTae: Yeah you are right!
I added code that checks count of child nodes.

Issue #438, opened 2018-11-14
Checkbox: allow props to be passed via props


moog16:
Nov 13 in Discord user reports issues of being able to pass props to <input /> field. He created a codepen. JS on the right is the JSX markup. The HTML on the left is what was output by the rendering engine. 

https://codepen.io/jbsmith969/pen/zMNVEL

What would be better is rendering the component through composition. Instead of doing:

```
<Checkbox ... />
```

we should instead do:

```
<Checkbox ... >
  <CheckboxInput ... />
</Checkbox>
```

jbsmith969:
Oh nice! I like that. Would you do same for the Switch component as well?

moog16:
Yes, especially for consistency that makes sense. I'll open one for that component

moog16:
Need to keep in mind this [bug](https://github.com/material-components/material-components-web-react/issues/308) when fixing this issue.

jbsmith969:
From what I'm reading on that bug, it would seem like the ref issue would be fixed with this wouldn't it?

moog16:
This component actually needs a handle on the ref. So it actually won't be fixed for free. But #308 is something we should have available in some way shape or form to give the developer more control. We may need to proxy the ref to the developer. Currently in text field we have a solution:

https://github.com/material-components/material-components-web-react/blob/master/packages/text-field/Input.js#L109

PR #678, opened 2019-02-15
fix(checkbox): enable name attribute

nicklasisraelsson:
This is a suggested interim solution until issue #438 can be resolved to be able to use Checkbox components in forms.


nicklasisraelsson:
I signed it!

moog16:
cla not passing - signed in https://github.com/material-components/material-components-web-react/pull/678#issuecomment-464049289

Issue #679, opened 2019-02-15
Setting anchor or button attributes on react-button causes type errors


ben-mckernan:
It should be possible to pass standard HTML attributes for buttons or anchors to the Button component. For example `<Button type="submit" />` or `<Button href="http://example.com" download />`. 

However both of these example with cause a type error in typescript, like:
```
TS2322: Type '{ download: true; }' is not assignable to type 'IntrinsicAttributes & IntrinsicClassAttributes<{ foundation?: any; isCo
mponentMounted: boolean; displayName: string; state: RippledComponentState; componentDidMount(): void; componentWillUnmount(): void; ... 20 more ...; refs: { ...; }; }> & Pick
<...> & Partial<...> & Partial<...>'.
  Property 'download' does not exist on type 'IntrinsicAttributes & IntrinsicClassAttributes<{ foundation?: any; isComponentMounted: boolean; displayName: string; state: Rippl
edComponentState; componentDidMount(): void; componentWillUnmount(): void; ... 20 more ...; refs: { ...; }; }> & Pick<...> & Partial<...> & Partial<...>'.
```
This is due to `ButtonProps` only extending `HTMLAttributes<T>`

moog16:
merged #680. Thanks for fixing!

PR #680, opened 2019-02-15
fix(button): Support anchor and button HTML attributes (#679)

ben-mckernan:
I've chosen to extend `React.AnchorHTMLAttributes<T>` and `React.ButtonHTMLAttributes<T>` since these are the possible element types that can be rendered. However looking through the other components you seem to use `HTMLProps<T>` a lot as well. There doesn't seem to be any clear reason why this is used in some places and not in others.

I'm happy to change this to use `HTMLProps<T>` instead if that is the pattern you want to follow. It would make button consistent with icon-button.

### DIFF
--- a/packages/button/index.tsx
+++ b/packages/button/index.tsx
@@ -28,16 +28,17 @@ const BUTTON_CLASS_NAME = 'mdc-button__icon';
 
 type ButtonTypes = HTMLAnchorElement | HTMLButtonElement;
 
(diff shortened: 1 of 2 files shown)


ben-mckernan:
I signed it!

moog16:
CLA was not passing, but signed in https://github.com/material-components/material-components-web-react/pull/680#issuecomment-464043832

Issue #690, opened 2019-02-22
TabBar/Tab should only have one tab stop for accessibility


ben-mckernan:
This is related to this issue: material-components/material-components-web#4429

On first render every tab in the tab-bar is a tab stop, since this is the default behavior for buttons without a tabIndex. When selecting another tab, the previously selected tab has its tabIndex state set to -1 meaning it is no longer possible to tab to.

A more accessible solution would be to ensure every tab except the active tab has tabIndex set to -1. This way there is only one tab stop for the tab-bar (assuming no one sets two or more tabs as active, but that is probably acceptable). Navigation between tabs would then occur with arrow keys as is already possible.

moog16:
merged #691. Thanks for fixing and opening this issue :+1:

PR #691, opened 2019-02-22
fix(tab): Set initial tabIndex state to -1 (#690)

ben-mckernan:
Sets the initial state for the `tabIndex` the same way it is set for `aria-selected`. If a tab is active then the `activate` method on the foundation will update this state on mount.

### DIFF
--- a/test/unit/tab/index.test.tsx
+++ b/test/unit/tab/index.test.tsx
@@ -29,6 +29,16 @@ test('adds the active class if props.active is true on mount', () => {
   assert.isTrue(wrapper.hasClass('mdc-tab--active'));
 });
 
+test('sets the tabIndex to 0 if props.active is true on mount', () => {
+  const wrapper = shallow(<Tab active />);
+  assert.equal(wrapper.prop('tabIndex'), 0);
+});
+
+test('sets the tabIndex to -1 if props.active is false on mount', () => {
+  const wrapper = shallow(<Tab active={false} />);
+  assert.equal(wrapper.prop('tabIndex'), -1);
+});
+
 test('adds a class from state.classList', () => {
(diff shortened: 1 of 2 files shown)


test/unit/tab/index.test.tsx
```diff
@@ -29,6 +29,16 @@ test('adds the active class if props.active is true on mount', () => {
   assert.isTrue(wrapper.hasClass('mdc-tab--active'));
 });
 
+test('sets the tabIndex to 0 if props.active is true on mount', () => {
+  const wrapper = shallow(<Tab active />);
+  assert.equal(wrapper.prop('tabIndex'), 0);
+});
+
+test('sets the tabIndex to -1 if props.active is false on mount', () => {
+  const wrapper = shallow(<Tab active={false} />);
```
patrickrodee: Could we update this to set no active property instead of explicitly setting `active={false}`? Will the test still pass?

moog16: Typically with these props, it wouldn't be hard-coded. It would usually look more like:
```
<Tab active={this.state.active} />
```

In that case it is more helpful to have that prop explicitly defined like that. Otherwise removing it would look like:

```
const props = {};
if (this.state.active) {
  props.active = true;
}
<Tab {...props} />
```

patrickrodee: Fair enough, thanks for explaining. My thinking was that removing the `active={false}` prop tests that a `Tab` has `tab-index=-1` by default without the user providing any active prop. However, that's more a test of `Component.defaultProps` working than of the change here.

moog16:
That totally makes sense...and I had a double take for a second :)

Issue #628, opened 2019-01-22
menu-surface Hoist anti-pattern?


aluminick:
https://github.com/material-components/material-components-web-react/blob/1d0870073bdf2d6753b286f83e1476d6db3f0265/packages/menu-surface/index.tsx#L170-L172

Is this an anti-pattern? I think portal should be used on this one.

moog16:
@aluminick I am not familiar with react.portals, but after reading the [docs](https://reactjs.org/docs/portals.html) I believe you to be correct. Would you like to take a crack at it?

aluminick:
@moog16 sure. Thanks.

aluminick:
@moog16 Sorry, I forgot about this issue.
I've hoisted menu-surface using ReactDOM Portal but it has to be wrapped. And the tests are failing.

moog16:
Oh great - do you have a branch already? Whenever you're ready please feel free to open a PR.

moog16:
Closed with #693

PR #693, opened 2019-02-26
fix(menu-surface): Hoist menu-surface via a portal (#500,#628)

ben-mckernan:
Use `ReactDOM.createPortal` when hoisting the menu surface so that react can handle mounting and unmounting.


moog16:
fixes https://github.com/material-components/material-components-web-react/issues/628
fixes https://github.com/material-components/material-components-web-react/issues/500

packages/menu-surface/index.tsx
```diff
@@ -368,7 +356,8 @@ class MenuSurface extends React.Component<MenuSurfaceProps, MenuSurfaceState> {
         {...otherProps}
       >
         {children}
-      </div>
+      </div>,
+      document.body
```
moog16: are you familiar with SSR? Will this break SSR?

packages/menu-surface/index.tsx
```diff
@@ -160,19 +161,6 @@ class MenuSurface extends React.Component<MenuSurfaceProps, MenuSurfaceState> {
     }
   }
 
-  private hoistToBody(): void {
-    // this deviates from the mdc web version.
```
moog16: please add this comment to line 118:
```
this.foundation.setIsHoisted(true);
```

test/unit/menu-surface/index.test.tsx
```diff
@@ -351,6 +346,8 @@ test('#adapter.getBodyDimensions returns width/height of body', () => {
     .foundation.adapter_.getBodyDimensions();
   assert.isAtLeast(dim.width, 0);
   assert.isAtLeast(dim.height, 0);
+  wrapper.unmount();
```
moog16: thanks for fixing this

packages/menu-surface/index.tsx
ben-mckernan: I've not worked with SSR but from my understanding it doesn't support portals. Does menu-surface currently support SSR? The simple approach would be check if document exists and return null if it doesn't. I'll add a commit which does this.

ben-mckernan:
I've added a fix for the SSR issue but I'm wondering if it would be better to follow the suggestion here https://github.com/facebook/react/issues/13097#issuecomment-405658104 let me know what you think and I can update the PR if needed.

packages/menu-surface/index.tsx
moog16: Cool that sounds good...Currently we have no SSR support, but we have been getting bugs - so I want to be cognizant of that.

I want to add SSR testing to the roadmap.

packages/menu-surface/index.tsx
```diff
@@ -347,6 +350,7 @@ class MenuSurface extends React.Component<MenuSurfaceProps, MenuSurfaceState> {
       children,
       ...otherProps
     } = this.props;
+    if (!document) return null;
```
moog16: @ben-mckernan yes please use a mounted flag as described [here](https://github.com/facebook/react/issues/13097#issuecomment-405658104). Please copy the pattern set in [linear-progress](https://github.com/material-components/material-components-web-react/blob/rc0.11.0/packages/linear-progress/index.tsx#L47). I don't think the property needs to be on state unless you can think of one.

ben-mckernan: It needs to be on state otherwise a re-render will not be triggered. The initial render will be null since it runs before `componentDidMount` is called for the first time.

moog16:
I ran `npm start`. Seems like menusurface does not open menu anymore. That also explains the failing screenshot tests. Can you look into this?

ben-mckernan:
It was working locally but I must have missed something. I'll look into it.

ben-mckernan:
Not sure what is up with the build and why the button screenshot tests keep timing out but seems to be unrelated to my changes as far as I can tell. I have fixed the issue with the menu-surface being open for the initial render.

moog16:
@ben-mckernan button screenshot sometimes does that...it is because there isn't enough time for the webpack-dev-server to start up. Just need to restart the build.

packages/menu-surface/index.tsx
```diff
@@ -128,13 +133,14 @@ class MenuSurface extends React.Component<MenuSurfaceProps, MenuSurfaceState> {
...
+    this.setState({mounted: true});
   }
 
-  componentDidUpdate(prevProps: MenuSurfaceProps) {
-    if (this.props.open !== prevProps.open) {
+  componentDidUpdate(prevProps: MenuSurfaceProps, prevState: MenuSurfaceState) {
+    // if this.props.open was true for the initial render
+    // then it will not be changed but the mounted state will be,
+    // so this.open_() should also be called in that case.
+    if (this.props.open !== prevProps.open || (this.props.open && this.state.mounted !== prevState.mounted)) {
```
moog16: Please add a test for this new case that you just added

ben-mckernan: Will do!

moog16:
Yes please squash. However the tests are failing...it looks like button, which is something you didn't touch. Also I ran them locally and they pass. Can you update the `35s` found in package.json to `40s`. It looks like this:

```
"test:screenshots": "docker run -it --rm --cap-add=SYS_ADMIN -e MDC_GCLOUD_SERVICE_ACCOUNT_KEY=\"${MDC_GCLOUD_SERVICE_ACCOUNT_KEY}\" mdcreact/screenshots /bin/sh -c 'git checkout .; git checkout master; git pull; npm i; /home/pptruser/material-components-web-react/test/screenshot/start.sh; sleep 35s; npm run test:image-diff'",
```

packages/menu-surface/package.json
```diff
@@ -19,7 +19,8 @@
   "dependencies": {
     "@material/menu-surface": "^0.41.0",
     "classnames": "^2.2.5",
-    "react": "^16.4.2"
+    "react": "^16.4.2",
+    "react-dom": "16.4.2"
```
moog16: @ben-mckernan can you also add the `^` to react-dom please?

ben-mckernan: Yes, of course. Good catch!

ben-mckernan:
Changing the sleep didn't help. But from the build log it looks like the `--timeout 30000` should be changed in:
```
"test:image-diff": "MDC_COMMIT_HASH=$(git rev-parse --short HEAD) MDC_BRANCH_NAME=$(git rev-parse --abbrev-ref HEAD) mocha --require ts-node/register --require babel-core/register --ui tdd --timeout 30000 test/screenshot/diff-suite.tsx"
```

moog16:
Hmm not sure if it is the timeout, because its only the first test. This leads me to believe that the startup is not complete. Also the tests all complete within 3s - 10s. If you look [here](https://travis-ci.com/material-components/material-components-web-react/jobs/184284480) you can see the times for each test to complete.

I'm going to restart the test and see if that fixes it.

moog16:
Try increasing the `40s` to `60s`...I'd hate to increase the time to that much, but lets just try it.

moog16:
Can't believe that didn't work...Trying out what you suggested

moog16:
@ben-mckernan not sure what's going on. try increasing the timeout for pupeteer in `test/screenshot/screenshot.tsx`.  Please just add the `timeout` option that looks like this: https://github.com/GoogleChrome/puppeteer/issues/782#issuecomment-329429325

moog16:
I'm wondering if this has something to do with adding react-dom package. That's the only thing I can think of that would slow this down compared to the other PRs that are out.

ben-mckernan:
Increasing `timeout` for puppeteer fixed the tests.

The webpack build for the screenshot tests already includes react-dom in its entry point so just adding react-dom shouldn't affect them, however the addition of `createPortal` does increase the size of the bundle produced by the screenshot tests. This seem to be enough to push the initial page load over the timeout threshold.

I don't know enough about how puppeteer works but I would speculate that this threshold would have eventually been reached regardless as new features or tests are added that increase the size of the bundle produced by the screenshot build. So something that might be worth testing is minifying the screenshot bundle to see if that affects the initial page load time.

But your point about react-dom did make me realize that it is not marked as external for the standard build and so it was being bundled with menu-surface. I have added a fix for this.

moog16:
@ben-mckernan oh man! Thanks for putting that in there (externals). That would have been a huge bundle increase...Anyways, what you last wrote made sense and I agree with it. Sorry this took so long, and I hope it doesn't deter you from contributing again. Thanks so much.

Issue #707, opened 2019-03-01
Snackbar is missing `open` prop for dynamic open/closing


moog16:
Currently there is no way to open/close snackbar without adding/removing from the dom. I propose that we use an `open` prop to allow for dynamic open/close:

```
<Snackbar open={isOpen} />
```

moog16:
fixed by #708

PR #708, opened 2019-03-01
snackbar: dynamic open

aluminick:
added missing componentDidUpdate for dynamic opening


packages/snackbar/index.tsx
```diff
@@ -153,6 +153,14 @@ export class Snackbar extends React.Component<Props, State> {
       this.foundation.open();
     }
   }
+  componentDidUpdate(prevProps: Props) {
+    const {open} = this.props;
+    if (prevProps.open !== open) {
+      if (open) {
+        this.foundation.open();
+      }
```
lucasecdb: shouldn't you call `this.foundation.close()` on an else clause?

moog16:
fixes https://github.com/material-components/material-components-web-react/issues/707

packages/snackbar/index.tsx
moog16: Correct...we need the `this.foundation.close(reason)`. I suspect we would also want a `closedReason` prop that gets passed to `reason` in the close method. I'm open to other ideas too.

lucasecdb: we could use the prop with a default of the dismissed value from the mdc snackbar, what do you think?

lucasecdb: [this one](https://github.com/material-components/material-components-web/blob/master/packages/mdc-snackbar/constants.ts#L40).

aluminick: @lucasecdb I don't think we need to call `this.foundation.close()` because the `foundation` is aware when the snackbar is closed. So why not just utilize the `notifyClosed` method of the `adapter` and use the `onClose` prop to update the component's `isOpen` state. Something like this:
```js
const Component = () => {
  const [isOpen, setIsOpen] = useState(false)
  return (
    <Snackbar open={isOpen} onClose={() => setIsOpen(false)} />
  )
}
```
This method is from the `Dialog` component and I think this makes sense.

lucasecdb: oh, right. but what if the user set the `open` to false _before_ the foundation actually auto-closes the snackbar?

aluminick: @lucasecdb sorry, my bad. Now I understand what you mean.

moog16: The use case is if someone wants to programmatically close the snackbar before it automatically closes. I've attached a [patch](https://stackoverflow.com/questions/5159185/create-a-git-patch-from-the-changes-in-the-current-working-directory/15438863#15438863) to add on to the screenshot tests. 

[snackbar.patch.zip](https://github.com/material-components/material-components-web-react/files/2937257/snackbar.patch.zip)

test/unit/snackbar/index.test.tsx
```diff
@@ -52,6 +52,13 @@ test('renders stacked actions', () => {
   wrapper.unmount();
 });
 
+test('#componentDidUpdate calls foundation.open if props.open is true', () => {
```
moog16: please add the opposite case for foundation.open = false

packages/snackbar/index.tsx
aluminick: @moog16 I've applied the patch. I didn't event know this thing exists. thanks :)
I just called foundation.close with an empty argument if open prop is false.

moog16: Can you add another prop for the reason? I'm thinking something like:
```
const {reason} = this.props;
...
this.foundation.close(reason)
```

moog16:
@aluminick just checking in to see what the status of this PR is.

packages/snackbar/index.tsx
aluminick: @moog16 Sure. Sorry for the late replies. I'll push the update you need in a moment.

packages/snackbar/index.tsx
```diff
@@ -38,6 +38,7 @@ export interface Props {
   leading?: boolean;
   stacked?: boolean;
   open?: boolean;
+  reason?: string;
```
moog16: @aluminick please add this prop to the readme

moog16:
Update golden.json snackbar entry to `3da3c11708088eb563f403f808e0db4d45096e8a098700f95bfcfca13b1c2773`

Issue #703, opened 2019-02-28
chips package relies on classnames 2.2.6 but depends on ^2.2.5


kfranqueiro:
This was originally reported by kittie on discord.

We are relying on the existence of a default export in `classnames` in `ChipSet.tsx`, but the default export was only added in classnames 2.2.6, whereas our package.json lists 2.2.5. If someone somehow installs 2.2.5, they will get an error like this:

```
index.js:590 Uncaught TypeError: classnames_1.default is not a function
    at ChipSet.get [as classes] (index.js:590)
    at Object.hasClass (index.js:603)
    at MDCChipSetFoundation.select (mdc.chips.js:3186)
    at index.js:622
    at forEachSingleChild (react.development.js:881)
    at traverseAllChildrenImpl (react.development.js:785)
    at traverseAllChildrenImpl (react.development.js:801)
    at traverseAllChildren (react.development.js:856)
    at Object.forEachChildren [as forEach] (react.development.js:901)
    at ChipSet.initChipSelection (index.js:618)
```

This likely flew under the radar as we have no package-lock for chips specifically, and the main package-lock is locked to 2.2.6.

moog16:
It occurs to me that instead of upgrading classnames package, we should the `import * as classnames from 'classnames'` syntax. ChipSet, as well as dialog, icon-button, snackbar, tab, tab-bar, tab-indicator/scroller were not converted correctly. I will reproduce this before making changes.

moog16:
TLDR; 
- update all packages to `^2.2.6` 
- classnames should be imported

**RE classnames:**
In <root>/package.json replaced `classnames: ^2.2.6` --> `classnames: 2.2.5`. Received error:
```
TS2345: Argument of type '{ 'mdc-text-field--outlined': boolean; 'mdc-text-field--textarea': boolean; 'mdc-text-field--fullwidth': boolean; 'mdc-text-field--disabled': boolean; 'mdc-text-field--with-trailing-icon': ReactElement<...> | undefined; 'mdc-text-field--with-leading-icon': ReactElement<...> | undefined; 'mdc-text-field--dense': bool...' is not assignable to parameter of type 'ClassValue'.
[...]

PR #709, opened 2019-03-01
fix: classnames@2.2.5 --> classnames@2.2.6 & update imports

moog16:
fixes https://github.com/material-components/material-components-web-react/issues/703#issuecomment-468769014


moog16:
ah! Thanks...I always forget about the example

moog16:
actually...the example isn't using typescript yet. The latest version it is running on is 0.7.0. Created an issue to update the example once I'm using v1-prerelease.

https://github.com/material-components/material-components-web-react/issues/710

PR #711, opened 2019-03-01
feat(ripple) mdc-web typescript conversion

gugu:
I am stuck with some cryptic typescript error, will fix it tomorrow

UPD: fixed


packages/ripple/index.tsx
```diff
@@ -71,12 +72,12 @@ export interface RippledComponentInterface<Surface, Activator = Element> {
...
   handleKeyUp: React.KeyboardEventHandler<Surface>;
   activateRipple: (e: ActivateEventTypes<Surface>) => void;
-  deactivateRipple: (e: ActivateEventTypes<Surface>) => void;
+  deactivateRipple: () => void;
   classes: string;
   style: React.CSSProperties;
   displayName: string;
   createAdapter: (surface: Surface, activator?: Activator) => MDCRippleAdapter;
-  updateCssVariable: (varName: keyof React.CSSProperties, value: string | number) => void;
+  updateCssVariable: (varName: string, value: string | null) => void;
```
moog16: I think you can revert the keyof React.CSSProperties.

packages/ripple/index.tsx
```diff
@@ -147,18 +150,18 @@ export function withRipple <
     this.foundation.init();
   };
 
-  createAdapter: MDCRippleAdapter = (surface: Surface, activator?: Activator) => {
+  createAdapter = (surface: Surface, activator?: Activator): MDCRippleAdapter => {
     return {
       browserSupportsCssVars: () => util.supportsCssVariables(window),
-      isUnbounded: () => this.props.unbounded,
+      isUnbounded: () => (this.props.unbounded || false) as boolean,
```
moog16: shouldn't you be able to do `isUnbounded: () => this.props.unbounded!,` here. `defaultProps` are set here, so `!` should be valid. Same goes for isSurfaceDisabled below.

packages/ripple/index.tsx
```diff
@@ -209,12 +220,12 @@ export function withRipple <
 
   handleFocus = (e: React.FocusEvent<Surface>) => {
     this.props.onFocus && this.props.onFocus(e);
-    this.foundation.handleFocus();
+    this.foundation!.handleFocus();
```
moog16: remove these `!` and initialize `foundation` as `foundation!` up top instead.

packages/ripple/index.tsx
```diff
@@ -209,12 +220,12 @@ export function withRipple <
...
   };
 
   handleBlur = (e: React.FocusEvent<Surface>) => {
     this.props.onBlur && this.props.onBlur(e);
-    this.foundation.handleBlur();
+    this.foundation!.handleBlur();
```
moog16: remove !

packages/ripple/index.tsx
```diff
@@ -247,26 +258,31 @@ export function withRipple <
...
     this.props.onKeyUp && this.props.onKeyUp(e);
-    this.deactivateRipple(e);
+    this.deactivateRipple();
   };
 
   activateRipple = (e: ActivateEventTypes<Surface>) => {
     // https://reactjs.org/docs/events.html#event-pooling
     e.persist();
-    this.foundation.activate(e);
+    this.foundation!.activate(e.nativeEvent);
```
moog16: remove !

packages/ripple/index.tsx
```diff
@@ -247,26 +258,31 @@ export function withRipple <
...
   };
 
-  deactivateRipple = (e: ActivateEventTypes<Surface>) => {
-    this.foundation.deactivate(e);
+  deactivateRipple = () => {
+    this.foundation!.deactivate();
```
moog16: remove !

packages/ripple/index.tsx
```diff
@@ -326,7 +342,7 @@ export function withRipple <
       />
     );
   }
-  };
+  } as any;
```
moog16: What is this for? We shouldn't need to cast the Ripple as any. I changed this and saw that components that use react ripple are throwing an error. Is that the error you are seeing as well? 

This is what I am seeing: https://github.com/mui-org/material-ui/issues/14487

packages/ripple/package.json
```diff
@@ -16,8 +16,8 @@
     "url": "https://github.com/material-components/material-components-web-react.git"
   },
   "dependencies": {
-    "@material/dom": "^0.41.0",
-    "@material/ripple": "^0.41.0",
+    "@material/dom": "^1.0.0-0",
```
moog16: 1.0.0

test/screenshot/stop.sh
```diff
@@ -1,3 +1,3 @@
 #!/bin/bash
 
-kill $(ps aux | grep 'webpack' | awk '{print $2}') || echo 'already stopped'
+kill $(ps aux | grep 'webpack' | grep 'screenshot' | awk '{print $2}') || echo 'already stopped'
```
moog16: Is this so that it also kills the npm screenshot tests running?

test/unit/ripple/index.test.tsx
```diff
@@ -39,6 +40,11 @@ interface RippledComponent extends React.Component<InjectedProps<HTMLDivElement>
   isComponentMounted: boolean;
 }
 
+function getAdapter(foundation: MDCRippleFoundation): MDCRippleAdapter {
+  // @ts-ignore
```
moog16: please add comment here.

packages/tab/TabRipple.tsx
```diff
@@ -30,6 +30,7 @@ import {
 
 export interface TabRippleProps extends React.HTMLAttributes<HTMLDivElement>, Ripple.InjectedProps<HTMLDivElement> {
   className: string;
+  ref?: React.RefObject<any>,
```
moog16: this seems unnecessary

packages/ripple/index.tsx
gugu: MDC-web has string in their interface and when I limit it to CSS properties type verification fails. Is there any better solution? And technically it is css property, which can be different with react css properties (borderBottom vs border-bottom, for example)

test/screenshot/stop.sh
gugu: it should not be there, I'll remove it tomorrow and will make another PR correctly. Currently it kills every webpack project running, including my project

packages/ripple/index.tsx
moog16: ah yes you're right...ok this seems like the correct change

packages/ripple/index.tsx
gugu: Yes, it is. I cast as any return value of this function, but specify a correct return value for withRipple function instead. In this case, it does not visit extra files to determine the file type

moog16: Hmm can you add a comment just saying it is a symlink issue. I investigated into this issue, and it just seems to be a short fall of typescript.

packages/tab/TabRipple.tsx
gugu: I moved it to the withRipple interface, but it is still needed because some components (Tab, for example, provides `ref` attribute)

moog16:
@gugu last comment: https://github.com/material-components/material-components-web-react/pull/711/files#r263426643

packages/ripple/index.tsx
```diff
@@ -147,18 +151,18 @@ export function withRipple <
...
+      isUnbounded: () => this.props.unbounded! as boolean,
       isSurfaceActive: () => {
         if (activator) {
           return matches(activator, ':active');
         }
 
         return matches(surface, ':active');
       },
-      isSurfaceDisabled: () => this.props.disabled,
+      isSurfaceDisabled: () => (this.props.disabled || false) as boolean,
```
moog16: please remove the || false here too.

gugu:
done

moog16:
The build hasn't passed yet, but I think there are some errors when you run `npm start`.

moog16:
@gugu please ping me when this is ready to be reviewed again. Also I opened a floatinglabel PR.

gugu:
@moog16 done

moog16:
Looks good! Thanks :)

moog16:
@gugu - `return new ClientRect();` is causing me errors. Is this even part of the ES5/ES6 library? Not sure how I didn't catch this before.

PR #721, opened 2019-03-07
fix(text-field): input not rendered on SSR

lucasecdb:
The `Input` component isn't being rendered on the first render (thus, not rendering on SSR) which can cause some issues for SEO and usability, such as the user being unable to type in the form while the JS hasn't run yet.

The markup I'm getting on my SSR with the latest version is

```html
<div class="mdc-text-field">
  <label class="mdc-floating-label" for="">Username</label>
  <div class="mdc-line-ripple"></div>
</div>
```

And with the changes on this PR

```html
<div class="mdc-text-field">
  <input type="text" value="" class="mdc-text-field__input" name="username" id=""/>
  <label class="mdc-floating-label" for="">Username</label>
  <div class="mdc-line-ripple"></div>
</div>
```

And the reason for that is because the `TextField` component actually controls whether or not the `Input` will be rendered, which depends on the foundation being initialized.

### DIFF
--- a/packages/text-field/Input.tsx
+++ b/packages/text-field/Input.tsx
@@ -134,7 +134,7 @@ export default class Input<T extends HTMLElement = HTMLInputElement> extends Rea
       handleValueChange && handleValueChange(value, () => {
         // only call #foundation.setValue on programatic changes;
         // not changes by the user. or when we go from no foundation
-        // to having a foundation, since we don't have that guaranteeded
+        // to having a foundation, since we don't have that guaranteed
         // anymore
         if (this.state.wasUserTriggeredChange && prevProps.foundation === foundation) {
           this.setState({wasUserTriggeredChange: false});
--- a/packages/text-field/Input.tsx
+++ b/packages/text-field/Input.tsx
@@ -134,10 +134,8 @@ export default class Input<T extends HTMLElement = HTMLInputElement> extends Rea
 
     if ((value !== prevProps.value || foundationChanged) && foundation) {
       handleValueChange && handleValueChange(value, () => {
-        // only call #foundation.setValue on programatic changes;
-        // not changes by the user. or when we go from no foundation
-        // to having a foundation, since we don't have that guaranteed
-        // anymore
+        // #foundation.setValue should only be called on programatic changes,
+        // not changes by the user, nor when the foundation changes.
         if (this.state.wasUserTriggeredChange && !foundationChanged) {
           this.setState({wasUserTriggeredChange: false});
         } else {
--- a/packages/text-field/Input.tsx
+++ b/packages/text-field/Input.tsx
@@ -98,10 +98,10 @@ export default class Input<T extends HTMLElement = HTMLInputElement> extends Rea
     if (setDisabled && disabled) {
       setDisabled(true);
     }
-    if (handleValueChange && value) {
+    if (handleValueChange && value && foundation) {
       handleValueChange(value, () => foundation.setValue(value));
     }
-    if (isValid !== undefined && foundation !== null) {
+    if (isValid !== undefined && foundation) {
       foundation.setUseNativeValidation(false);
       foundation.setValid(isValid);
     }
@@ -121,7 +121,7 @@ export default class Input<T extends HTMLElement = HTMLInputElement> extends Rea
 
     this.handleValidationAttributeUpdate(prevProps);
 
-    if (disabled !== prevProps.disabled) {
+    if ((disabled !== prevProps.disabled || prevProps.foundation !== foundation) && foundation) {
       setDisabled && setDisabled(disabled!);
       foundation.setDisabled(disabled);
     }
@@ -130,11 +130,13 @@ export default class Input<T extends HTMLElement = HTMLInputElement> extends Rea
       setInputId && setInputId(id!);
     }
 
-    if (value !== prevProps.value) {
+    if ((value !== prevProps.value || prevProps.foundation !== foundation) && foundation) {
       handleValueChange && handleValueChange(value, () => {
         // only call #foundation.setValue on programatic changes;
-        // not changes by the user.
-        if (this.state.wasUserTriggeredChange) {
+        // not changes by the user. or when we go from no foundation
+        // to having a foundation, since we don't have that guaranteeded
+        // anymore
+        if (this.state.wasUserTriggeredChange && prevProps.foundation === foundation) {
           this.setState({wasUserTriggeredChange: false});
         } else {
           foundation.setValue(value);
@@ -142,7 +144,7 @@ export default class Input<T extends HTMLElement = HTMLInputElement> extends Rea
       });
     }
 
-    if (isValid !== prevProps.isValid || foundation !== prevProps.foundation) {
+    if ((isValid !== prevProps.isValid || foundation !== prevProps.foundation) && foundation) {
       if (isValid === undefined) {
         foundation.setUseNativeValidation(true);
       } else {
--- a/packages/text-field/Input.tsx
+++ b/packages/text-field/Input.tsx
@@ -90,6 +90,7 @@ export default class Input<T extends HTMLElement = HTMLInputElement> extends Rea
       setDisabled,
       handleValueChange,
       foundation,
+      isValid,
     } = this.props;
     if (setInputId && id) {
       setInputId(id!);
@@ -99,6 +100,10 @@ export default class Input<T extends HTMLElement = HTMLInputElement> extends Rea
     }
     if (handleValueChange && value) {
       handleValueChange(value, () => foundation.setValue(value));
+    }
+    if (isValid !== undefined && foundation !== null) {
+      foundation.setUseNativeValidation(false);
+      foundation.setValid(isValid);
     }
   }
 
--- a/packages/text-field/Input.tsx
+++ b/packages/text-field/Input.tsx
@@ -121,7 +121,9 @@ export default class Input<T extends HTMLElement = HTMLInputElement> extends Rea
 
     this.handleValidationAttributeUpdate(prevProps);
 
-    if ((disabled !== prevProps.disabled || prevProps.foundation !== foundation) && foundation) {
+    const foundationChanged = prevProps.foundation !== foundation;
+
+    if ((disabled !== prevProps.disabled || foundationChanged) && foundation) {
       setDisabled && setDisabled(disabled!);
       foundation.setDisabled(disabled);
     }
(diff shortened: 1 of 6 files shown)


lucasecdb:
I'll try to fix the tests tomorrow

packages/text-field/index.tsx
```diff
@@ -329,7 +329,7 @@ class TextField<T extends {}> extends React.Component<TextFieldProps<T>, TextFie
         key='text-field-container'
       >
         {leadingIcon ? this.renderIcon(leadingIcon, onLeadingIconSelect) : null}
-        {foundation ? this.renderInput() : null}
+        {this.renderInput()}
```
moog16: This isn't needed anymore?

lucasecdb:
hey @moog16, sorry not explaining the bug I've experienced. I've updated the PR description!

packages/text-field/index.tsx
lucasecdb: this fix is because of this line, actually

lucasecdb:
I need some help to fix the other tests, I don't know why they are failing

moog16:
@lucasecdb which tests specifically are you running into issues with? Are they the ones on travis?

lucasecdb:
@moog16 yeah, those ones.

packages/text-field/Input.tsx
```diff
@@ -102,7 +102,7 @@ export default class Input<T extends {}> extends React.Component<
     if (handleValueChange && value) {
       handleValueChange(value, () => foundation.setValue(value));
```
moog16: Because we no longer have the security of the foundation always being set, we need a guard here to check if the foundation exists. We also need a case in the componentDidUpdate to do this same thing if the foundation changes from `undefined` --> `MDCTextFieldFoundation` (no existing --> existing). This also goes for the same 2 lines below:

```
 if (isValid !== undefined && foundation !== null) {
  foundation.setUseNativeValidation(false);
  foundation.setValid(isValid);
    }	 
```

Does that make sense?

lucasecdb: oh, yeah! I'll try that

lucasecdb:
@moog16 I've fixed the tests, but the screenshot isn't passing

packages/text-field/Input.tsx
```diff
@@ -121,7 +121,7 @@ export default class Input<T extends HTMLElement = HTMLInputElement> extends Rea
 
     this.handleValidationAttributeUpdate(prevProps);
 
-    if (disabled !== prevProps.disabled) {
+    if ((disabled !== prevProps.disabled || prevProps.foundation !== foundation) && foundation) {
```
moog16: what if we change this from `...prevProps.foundation !== foundation) && foundation` to `!prevProps.foundation && foundation`? Seems clearer. If that works can you make that a variable and use it on line 133 as well

packages/text-field/Input.tsx
```diff
@@ -130,19 +130,21 @@ export default class Input<T extends HTMLElement = HTMLInputElement> extends Rea
...
-    if (value !== prevProps.value) {
+    if ((value !== prevProps.value || prevProps.foundation !== foundation) && foundation) {
       handleValueChange && handleValueChange(value, () => {
         // only call #foundation.setValue on programatic changes;
-        // not changes by the user.
-        if (this.state.wasUserTriggeredChange) {
+        // not changes by the user. or when we go from no foundation
+        // to having a foundation, since we don't have that guaranteed
+        // anymore
+        if (this.state.wasUserTriggeredChange && prevProps.foundation === foundation) {
```
moog16: I find this confusing, since right above this we are checking to see if the foundation has changed. Is there some way to make this a little more clear?

packages/text-field/Input.tsx
```diff
@@ -130,19 +130,21 @@ export default class Input<T extends HTMLElement = HTMLInputElement> extends Rea
...
           this.setState({wasUserTriggeredChange: false});
         } else {
           foundation.setValue(value);
         }
       });
     }
 
-    if (isValid !== prevProps.isValid) {
+    if ((isValid !== prevProps.isValid || foundation !== prevProps.foundation) && foundation) {
```
moog16: please use that variable here too.

packages/text-field/Input.tsx
lucasecdb: I extracted this checking to a variable, I think it's better now

packages/text-field/Input.tsx
lucasecdb: I think it's a bit more clear now, with the variable name for the condition

packages/text-field/Input.tsx
```diff
@@ -130,19 +132,19 @@ export default class Input<T extends HTMLElement = HTMLInputElement> extends Rea
...
 
-    if (value !== prevProps.value) {
+    if ((value !== prevProps.value || foundationChanged) && foundation) {
       handleValueChange && handleValueChange(value, () => {
-        // only call #foundation.setValue on programatic changes;
-        // not changes by the user.
-        if (this.state.wasUserTriggeredChange) {
+        // #foundation.setValue should only be called on programatic changes,
+        // not changes by the user, nor when the foundation changes.
+        if (this.state.wasUserTriggeredChange && !foundationChanged) {
```
moog16: This reads better, but can you add a unit test for this case?

lucasecdb: sure!

test/unit/text-field/Input.test.tsx
```diff
@@ -403,3 +405,13 @@ test('#inputElement should return the native input', () => {
   assert.equal(inputElement!.tagName.toLowerCase(), 'input');
   assert.isTrue(inputElement instanceof HTMLInputElement);
 });
+
+test('should call foundation.setValue() when foundation is new', () => {
```
moog16: This test seems pretty similar to this
https://github.com/material-components/material-components-web-react/blob/c485beacc95aeb3e819a1741888ecaaa8614bb3d/test/unit/text-field/Input.test.tsx#L288

I was referring to the new condition testing that `this.setState({wasUserTriggeredChange: false});` was called.

lucasecdb: ohh, okay

moog16: Ya sorry - Also I'm updating this to rc0.11.0_real, which can still go out tomorrow.

test/unit/text-field/Input.test.tsx
```diff
@@ -403,3 +405,38 @@ test('#inputElement should return the native input', () => {
...
+  const wrapper = shallow<Input<HTMLInputElement>>(
+    <Input
+      handleValueChange={handleValueChange}
+      foundation={foundation}
+    />
+  );
+  const event = {target: {value: 'meow'}};
+  wrapper.simulate('change', event);
+  wrapper.setProps({value});
+  td.verify(foundation.setValue(), {times: 0});
```
moog16: this would actually never be called. It needs to be called with foundation.setValue(value)

test/unit/text-field/Input.test.tsx
```diff
@@ -403,3 +405,38 @@ test('#inputElement should return the native input', () => {
...
+  const wrapper = shallow<Input<HTMLInputElement>>(<Input />);
+  const setValue = td.func();
+  const foundation = {setValue, setDisabled: td.func(), setUseNativeValidation: td.func()};
+  const handleValueChange = (_: any, callback: () => void) => callback();
+  const value = 'meow';
+  wrapper.setProps({foundation, value, handleValueChange});
+  td.verify(foundation.setValue(value), {times: 1});
+});
+
+test('shouldn\'t call foundation.setValue() when triggered by user', () => {
```
moog16: Tested an looks good!

moog16:
@lucasecdb looksl ike screenshots are failing

lucasecdb:
@moog16 I don't know how the screenshot works, but maybe the component isn't being fully rendered (i.e. the didMount isn't being called)?

moog16:
@lucasecdb I looked at the 2 different screenshots...this is the [golden](https://storage.cloud.google.com/screenshot-uploads/text-field/standard/be2ea75813583dac8d3ad988282cfa19fa7266a29b095cbd60a34912a1900251.golden.png?organizationId=433637338589&_ga=2.240207811.-1123561096.1535414695). The ones that are failing look like the validation styles aren't appearing quick enough.

Locally they are passing, which leads me to think that the ones on travis are too slow.

lucasecdb:
@moog16 so, I need to increase the screenshot timeout?

moog16:
@lucasecdb I'm seeing if changing networkidle2 --> networkidle0 will work. Can you try changing that in test/screeshot/screenshot.tsx?

lucasecdb:
sure!

moog16:
actually try changing it to `load`...I think `load` comes last. https://github.com/GoogleChrome/puppeteer/blob/master/docs/api.md

I'm reading [this](https://stackoverflow.com/questions/2414750/difference-between-domcontentloaded-and-load-events) for reference...not sure who is right.

lucasecdb:
in [this screenshot tester](https://github.com/burnpiro/puppeteer-screenshot-tester#examples) they say to wait until the `networkidle0`, we could wait to see if that passes before trying the `load`, what do you think?

moog16:
@lucasecdb looks like all of the options fail locally. I think the `foundation` check ensured that the input was fully rendered before the networkidlex event occurred. We might need to update the screenshot tests. Are you on Discord? This might be easier to fix there

lucasecdb:
yes, I am!

moog16:
Talked offline and will push this off for either a patch / next release. Need to fix how we determine when a page is loaded w/ screenshot tests.

moog16:
@lucasecdb have you thought about this one at all? I wanna try getting this into 12.0

lucasecdb:
@moog16 not yet, I'll try to have a look on it later today

lucasecdb:
can someone help us on this?

moog16:
You mean do I have other React resources to help fix the tests? I'm actually the only one on the team that works with React, haha. I'll have to revisit once I get select working.

lucasecdb:
ohh, okayy. it's just that I don't know how to tackle this, tbh, but I didn't wanted to leave this out for 0.12 (which releases next week, right?)

moog16:
Correct, but I actually wanted to get the updated MDC Web v1 components in 12 as well. So I was thinking I would push it off until we get select and text field into the mdcweb update branch.

lucasecdb:
@moog16 I just checked this again, and the input on v0.12 is rendering on SSR!! I think we can close this without any problems

moog16:
That's great! How did it all of a sudden start working? I didn't even realize, but @TroyTae removed the `foundation ?` ternary. I wonder why we didn't run into the same testing issues.

lucasecdb:
yeah, that's strange, but I'm glad it's fine :-)

Issue #734, opened 2019-03-12
TS props required when should be optional


moog16:
related #732

- [x] chips/chip-set/index.tsx
- [x] drawer/index.tsx
- [x] linear-progress/index.tsx
- [x] list/index.tsx
- [x] menu-surface/index.tsx
- [x] notched-outline/index.tsx
- [x] select/index.tsx
- [x] switch/index.tsx
- [x] text-field/helper-text/index.tsx

Issue #732, opened 2019-03-12
Text Field props are all required


moog16:
Many of the Text Field TS props are required, when they should be optional. This results in the error:
```
Type '{ children: Element; label: string; outlined: true; }' is not assignable to type 'Readonly<TextFieldProps<{}>>'.
    Property 'className' is missing in type '{ children: Element; label: string; outlined: true; }'.
```

moog16:
Closing since #735 is in v0.11.0

PR #735, opened 2019-03-12
fix(text-field): make props optional

moog16:
https://github.com/material-components/material-components-web-react/issues/732
related https://github.com/material-components/material-components-web-react/issues/734

### DIFF
--- a/packages/text-field/Input.tsx
+++ b/packages/text-field/Input.tsx
@@ -25,22 +25,20 @@ import * as classnames from 'classnames';
 import {MDCTextFieldFoundation} from '@material/textfield/dist/mdc.textfield';
 
 export interface InputProps<T> {
-  className: string;
-  inputType: 'input' | 'textarea';
-  disabled: boolean;
+  className?: string;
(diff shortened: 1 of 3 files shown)


packages/text-field/Input.tsx
```diff
@@ -25,22 +25,20 @@ import * as classnames from 'classnames';
 import {MDCTextFieldFoundation} from '@material/textfield/dist/mdc.textfield';
 
 export interface InputProps<T> {
-  className: string;
-  inputType: 'input' | 'textarea';
-  disabled: boolean;
```
williamernest: Should this have been removed ?  It looks like it's still referenced.

moog16: Nope!

moog16: OH...yes I meant to remove disabled and id, because its duplicated in `type Props<T> = InputProps<T> & (T extends HTMLInputElement ? InputElementProps : TextareaElementProps);`

PR #737, opened 2019-03-12
fix: make require props optional

moog16:
fixes https://github.com/material-components/material-components-web-react/issues/734

This PR makes all props optional that were required.

### DIFF
--- a/packages/drawer/index.tsx
+++ b/packages/drawer/index.tsx
@@ -61,7 +61,7 @@ class Drawer extends React.Component<DrawerProps, DrawerState> {
 
   state: DrawerState = {classList: new Set()};
 
-  static defaultProps = {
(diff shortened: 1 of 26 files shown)


packages/drawer/index.tsx
```diff
@@ -40,15 +40,13 @@ import {FocusTrap} from 'focus-trap';
 const {cssClasses: listCssClasses} = MDCListFoundation;
 
 export interface DrawerProps extends React.HTMLProps<HTMLElement>{
-  className: string;
-  open: boolean;
-  onOpen: () => void;
-  onClose: () => void;
-  onTransitionEnd: React.TransitionEventHandler<HTMLElement>;
```
moog16: removed these since they are duplicates covered in HTMLProps.

Issue #697, opened 2019-02-26
Use MDC Web Typings


moog16:
- [ ] Update MDC Web select (https://github.com/material-components/material-components-web-react/issues/242)
- [x] Update to updated notched outline (https://github.com/material-components/material-components-web-react/issues/560)
- [x] Update list. Breaking changes in 42, 43, and 44 (#740)
- [x] Update checkbox, fab from breaking changes in 0.42.0 https://github.com/material-components/material-components-web/blob/master/CHANGELOG.md#0420-2018-12-04
- [x] Update snackbar, button, tab-bar from 0.42.0 changes https://github.com/material-components/material-components-web/blob/master/CHANGELOG.md#0430-2019-01-07
- [ ] Update text field from 0.44.0 changes
- [ ] Add Menu to add select
- [x] Update Chips from 0.44.0 changes

Components with breaking changes
- [x] tab-bar
- [x] snackbar
- [x] button
- [x] list
- [x] checkbox
- [x] fab
- [ ] text field
- [ ] select
- [x] notched outline

Other components with no breaking changes
- [x] card
- [x] chips
- [x] dialog
- [x] drawer
- [x] floating-label
- [x] icon-button (@ben-mckernan)
- ~~[ ] layout-grid~~
- [x] line-ripple
- [x] linear-progress
- [x] menu-surface
- [x] radio
- [x] ripple
- [x] switch
- [x] tab
- [x] tab-indicator
- [x] tab-scroller
- [x] top-app-bar
- [x] typography

gugu:
material-icon has no dependency on mdc-web

gugu:
![image](https://user-images.githubusercontent.com/75169/55233034-4f160480-5238-11e9-9a64-25a30a55ebe4.png)
Bundle size before: 1.74 Mb from material-web
![image](https://user-images.githubusercontent.com/75169/55233334-5ab5fb00-5239-11e9-81ee-ce7b5ec6f45a.png)
Current bundle size: 960 kb from material-web and we can save 560 kb more

moog16:
Are you saying by updating to the latest, we're cutting ~70%??

gugu:
Yes. That's why I'm working on it - I want to reduce bundle size

gugu:
But it is mdc-web only, @material/react-* will have the same size

moog16:
[...]

PR #740, opened 2019-03-13
feat(list): update to MDC Web 1.0.0

moog16:
related #697


moog16:
Thanks @abhiomkar - will do!

PR #741, opened 2019-03-13
feat(floating-label): update mdc web to v1.0.0

moog16:
related #697


packages/floating-label/index.tsx
```diff
@@ -83,17 +83,22 @@ export default class FloatingLabel extends React.Component<
...
+  get adapter(): MDCFloatingLabelAdapter {
     return {
       addClass: (className: string) =>
         this.setState({classList: this.state.classList.add(className)}),
       removeClass: this.removeClassFromClassList,
+      // the adapter methods below are effectively useless since React
+      // handles events and width differently
+      registerInteractionHandler: () => undefined,
+      deregisterInteractionHandler: () => undefined,
+      getWidth: () => 0,
```
abhiomkar: why does this need to always return `0`? Add a comment.

packages/floating-label/index.tsx
```diff
@@ -104,8 +109,8 @@ export default class FloatingLabel extends React.Component<
 
   handleWidthChange = () => {
     const {handleWidthChange} = this.props;
-    if (handleWidthChange && this.labelElement_.current) {
-      handleWidthChange(this.labelElement_.current.offsetWidth);
+    if (handleWidthChange && this.labelElement.current) {
+      handleWidthChange(this.labelElement.current.offsetWidth);
```
abhiomkar: Can we directly refer `this.props.handleWidthChange` instead? I read this like `handleWidthChange` method calling by itself.

Issue #401, opened 2018-11-01
Top App Bar needs ability to setScrollTarget


moog16:
top app bar needs ability to scroll from outside of window. Currently like: `getViewportScrollY: () => window.pageYOffset,`, should be 
```
getViewportScrollY: () =>
        this.scrollTarget_[this.scrollTarget_ === window ? 'pageYOffset' : 'scrollTop'],
```

related https://github.com/material-components/material-components-web/pull/2914

moog16:
Closing for #771

PR #744, opened 2019-03-14
feat(top-app-bar): setScrollTarget

TroyTae:
related to #401

I signed it!

There is one question apart from this PR.
```
    if (prevProps.scrollTarget !== this.props.scrollTarget) {
      this.foundation.destroyScrollHandler();
      this.setState({scrollTarget: this.props.scrollTarget});
    }

    if (prevState.scrollTarget !== this.state.scrollTarget) {
      this.foundation.initScrollHandler();
    }
```
Above code is 133~140 lines in the top-app-bar index.tsx.
Is this code safe?
If above expression is true and underlying expression is false, it just destroy scroll hander.
It doesn't matter?


moog16:
@TroyTae I'm actually confused why that issue is still open, since it seems like we have the code to support it. However we don't have any screenshot tests for it so I can't test it quickly.

Are you saying that the current code in `componentdDidUpdate` is not working?

TroyTae:
@moog16 
I don't see screenshot test code yet.
But I'll try make it!

No no maybe componentdDidUpdate is worked.
I guess I ask this question because I don't know about difference between state's scrollTarget and prop's scrollTarget.
I'm worried when (prevProps.scrollTarget !== this.props.scrollTarget) => true and (prevState.scrollTarget !== this.state.scrollTarget) => false, scroll handler is destroyed but not init.

I'm sorry I did not explain well 😅

moog16:
Oh I see...So if the `props.scrollTarget` changes we want to destroy and set `state.scrollTarget` to the new `props.scrollTarget`. The `state.scrollTarget` will update in the next cycle, and call `initScrollHandler`, which will reinit the top app bar. So this does work. :)

moog16:
@TroyTae can we close this PR? I'm not sure what your intent was in here.

TroyTae:
@moog16 
I try to write some screenshot test code.
Would it be better to just close it?

moog16:
@TroyTae - please feel free to write a screenshot test for it. We do have it covered under unit tests

PR #748, opened 2019-03-14
fix(tab): mdc-web v1.0.0 upgrade

gugu:


packages/tab-bar/index.tsx
```diff
@@ -47,7 +47,7 @@ class TabBar extends React.Component<
     this.foundation.init();
     const {activeIndex, indexInView} = this.props;
     if (this.tabList[activeIndex]) {
-      this.tabList[activeIndex].activate({} /* previousIndicatorClientRect */);
+      this.tabList[activeIndex].activate(new DOMRect() /* previousIndicatorClientRect */);
```
moog16: This is actually not IE11 compatible. I also realize we currently don't work on IE11, but we have some issues to be IE11 compatible. Please just change this to:

```
// new DOMRect is not IE11 compatible
const defaultDOMRect = {
bottom: 0
height: 0
left: 0
right: 0
top: 0
width: 0
x: 0
y: 0
};
...activate(defaultDOMRect);
```

You may need to cast it.

packages/tab/index.tsx
```diff
@@ -38,10 +38,11 @@ export interface TabProps extends React.HTMLProps<HTMLButtonElement> {
   isMinWidthIndicator?: boolean;
   stacked?: boolean;
   previousIndicatorClientRect?: ClientRect;
+  onInteration?: () => void;
 }
 
 interface MDCTabElementAttributes {
-  'aria-selected': boolean;
+  'aria-selected': 'false' | 'true' | undefined;
```
moog16: change to `'aria-selected'?: 'false' | 'true'`

packages/tab/index.tsx
```diff
@@ -81,7 +82,7 @@ export default class Tab extends React.Component<TabProps, TabState> {
     const {active, focusOnActivate} = this.props;
     this.foundation = new MDCTabFoundation(this.adapter);
     this.foundation.init();
-    this.foundation.setFocusOnActivate(focusOnActivate);
+    this.foundation.setFocusOnActivate(focusOnActivate as boolean);
```
moog16: change this to `focusOnActivate!`. This has a defaultProp, so we can `!`.

packages/tab/index.tsx
```diff
@@ -94,7 +95,7 @@ export default class Tab extends React.Component<TabProps, TabState> {
   componentDidUpdate(prevProps: TabProps) {
     const {active, focusOnActivate, previousIndicatorClientRect} = this.props;
     if (focusOnActivate !== prevProps.focusOnActivate) {
-      this.foundation.setFocusOnActivate(focusOnActivate);
+      this.foundation.setFocusOnActivate(focusOnActivate as boolean);
```
moog16: same with here.

packages/tab/index.tsx
```diff
@@ -136,12 +137,15 @@ export default class Tab extends React.Component<TabProps, TabState> {
...
+          this.tabContentRef.current.offsetLeft :
+          0,
       getContentOffsetWidth: () =>
-        this.tabContentRef.current &&
-        this.tabContentRef.current.offsetWidth,
+        this.tabContentRef.current ?
+          this.tabContentRef.current.offsetWidth :
+          0,
       focus: () => this.tabRef.current && this.tabRef.current.focus(),
+      notifyInteracted: () => this.props.onInteration,
```
moog16: notifyInteracted: this.props.onInteraction

packages/tab/index.tsx
```diff
@@ -38,10 +38,11 @@ export interface TabProps extends React.HTMLProps<HTMLButtonElement> {
   isMinWidthIndicator?: boolean;
   stacked?: boolean;
   previousIndicatorClientRect?: ClientRect;
+  onInteration?: () => void;
```
moog16: onInteraction

moog16:
Screen Shot 2019-03-18 at 11 46 39 AM

Getting a warning. Please fix

moog16:
@gugu  can you please also add the new prop to the readme

gugu:
@moog16 done

PR #749, opened 2019-03-15
feat(drawer): add innerRef prop

lucasecdb:
this PR add the possibility to pass a `innerRef` property to grab the ref of the `aside` (or the tag you pass in via prop).

this also upgrades the mdc-web dependency to v1.


packages/drawer/index.tsx
```diff
@@ -184,15 +190,32 @@ class Drawer extends React.Component<DrawerProps, DrawerState> {
...
   };
 
   handleTransitionEnd = (evt: React.TransitionEvent<HTMLElement>) => {
     this.props.onTransitionEnd!(evt);
     if (!this.foundation) return;
-    this.foundation.handleTransitionEnd(evt);
+    this.foundation.handleTransitionEnd(evt.nativeEvent);
   };
 
+  attachRef = (node: HTMLElement) => {
```
moog16: I like this pattern - we should actually normalize this across all components. We are doing something similar with textfield but this is a lot cleaner. 
Created an issue:
https://github.com/material-components/material-components-web-react/issues/753

packages/drawer/index.tsx
```diff
@@ -47,15 +48,20 @@ export interface DrawerProps extends React.HTMLProps<HTMLElement>{
...
 };
 
+const isRefObject = function(ref: DrawerProps['innerRef']): ref is React.RefObject<HTMLElement> {
+  return typeof ref !== 'function';
+};
+
 class Drawer extends React.Component<DrawerProps, DrawerState> {
   previousFocus: HTMLElement | null = null;
-  foundation: MDCDismissibleDrawerFoundation | MDCModalDrawerFoundation;
+  foundation!: MDCDismissibleDrawerFoundation | MDCModalDrawerFoundation;
```
moog16: I don't think we should do this in drawer mainly because permanent doesn't have a foundation. I changed the `!` --> `?`. This only changed threw a TS error in componentDidUpdate. Can we add a foundation check there? I think that should be all.

packages/drawer/index.tsx
```diff
@@ -184,15 +190,32 @@ class Drawer extends React.Component<DrawerProps, DrawerState> {
...
   };
 
+  attachRef = (node: HTMLElement) => {
+    const {innerRef} = this.props;
+
+    // https://github.com/facebook/react/issues/13029#issuecomment-410002316
+    // @ts-ignore this is acceptable according to the comment above
+    this.drawerElement.current = node;
+
+    if (innerRef !== undefined) {
```
moog16: Can we make this an early return? Just do:
```
if (!innerRef) return;
```

Does this need to specifically check for undefined?

lucasecdb: about the early return: sure! If I check for anything other than undefined I think typescript won't be able to infer that this isn't a "?" type, but I could try

packages/drawer/index.tsx
lucasecdb: okay, will do

lucasecdb: I think I'll need to update the tests to, adding a `!`.

packages/drawer/index.tsx
```diff
@@ -47,15 +48,20 @@ export interface DrawerProps extends React.HTMLProps<HTMLElement>{
   tag?: string;
   dismissible?: boolean;
   modal?: boolean;
+  innerRef?: RefCallback<HTMLElement> | React.RefObject<HTMLElement>;
```
moog16: please add to readme.

lucasecdb: done :-)

moog16:
@lucasecdb bad news...I'm having to create a 0.10.1 branch so that I can remove the TypeScript updates from tomorrow's release. Do you need this change for tomorrow?

lucasecdb:
@moog16 ohh, yeah I needed it :-(

lucasecdb:
can we cherry-pick the commits that aren't related to the typescript change?

moog16:
Ya...could you create another branch off of v0.11.0_real. you can just cherry-pick what you need into that branch. I would do it myself, but I have no one to do a review for me.

lucasecdb:
okay, will do

moog16:
Ok thanks - sorry for the inconvenience

lucasecdb:
no problem!

PR #750, opened 2019-03-15
fix(chips): upgrade mdc-web to v1.0.0

gugu:


packages/chips/Chip.tsx
```diff
@@ -78,21 +78,21 @@ export class Chip extends React.Component<ChipProps, ChipState> {
     const {selected, shouldRemoveOnTrailingIconClick} = this.props;
     this.foundation = new MDCChipFoundation(this.adapter);
     this.foundation.init();
-    this.foundation.setSelected(selected);
+    this.foundation.setSelected(selected as boolean);
```
moog16: i don't think you need to cast here. I believe it just needs to be `selected!`

packages/chips/Chip.tsx
```diff
@@ -78,21 +78,21 @@ export class Chip extends React.Component<ChipProps, ChipState> {
...
     if (shouldRemoveOnTrailingIconClick !== this.foundation.getShouldRemoveOnTrailingIconClick()) {
-      this.foundation.setShouldRemoveOnTrailingIconClick(shouldRemoveOnTrailingIconClick);
+      this.foundation.setShouldRemoveOnTrailingIconClick(shouldRemoveOnTrailingIconClick as boolean);
```
moog16: same with here

packages/chips/Chip.tsx
```diff
@@ -78,21 +78,21 @@ export class Chip extends React.Component<ChipProps, ChipState> {
...
     }
   }
 
   componentDidUpdate(prevProps: ChipProps) {
     const {selected, shouldRemoveOnTrailingIconClick} = this.props;
 
     if (selected !== prevProps.selected) {
-      this.foundation.setSelected(selected);
+      this.foundation.setSelected(selected as boolean);
```
moog16: selected!

packages/chips/Chip.tsx
```diff
@@ -78,21 +78,21 @@ export class Chip extends React.Component<ChipProps, ChipState> {
...
     }
 
     if (shouldRemoveOnTrailingIconClick !== prevProps.shouldRemoveOnTrailingIconClick) {
-      this.foundation.setShouldRemoveOnTrailingIconClick(shouldRemoveOnTrailingIconClick);
+      this.foundation.setShouldRemoveOnTrailingIconClick(shouldRemoveOnTrailingIconClick as boolean);
```
moog16: shouldRemoveOnTrailingIconClick!

test/unit/chips/ChipSet.test.tsx
```diff
@@ -120,30 +120,30 @@ test(
...
 
 test('#handleSelect calls foundation.handleChipSelection with selectedChipId and selected=true', () => {
   const handleChipSelection = td.func();
-  const foundation = {handleChipSelection};
   const wrapper = shallow<ChipSet>(
     <ChipSet>
       <div id='1' />
     </ChipSet>
   );
+  const foundation = coerceForTesting<any>({handleChipSelection});
```
moog16: Does `Partial<MDCChipFoundation>` work here?

gugu: The same error:
```
TS2322: Type 'import("/home/kostenko/w/material-components-web-react/node_modules/@material/chips/chip-set/foundation").MDCChipSetFoundation' is not assignable to type 'import("/home/kostenko/w/material-components-web-react/packages/chips/node_modules/@material/chips/chip-set/foundation").MDCChipSetFoundation'.
```

It can be fixed globally by adding option `--hoist` to `lerna bootstrap`, but I don't know if it is correct to do this in this pull request

moog16: I'm not sure if we would even want to do that. It sounds like we would only have one version of each component, which could hurt us. This is especially apparent in a case like this TS upgrade. We're about half way with the update, but we release on Tuesday. Its fine to release this way, but hoist might hurt us. 

I'm not 100% about hoist, since I just read about it on lerna docs. Since these are just tests and we're trying to mock the foundation, lets just keep it. Thanks!

PR #756, opened 2019-03-17
chore(switch): remove unused mdcw-ripple dep

lucasecdb:
This dependency isn't being used and, with the next release, this package would break when doing SSR because this version of `@material/ripple` contains direct reference to `HTMLElement`.


moog16:
Thanks @lucasecdb!!

PR #760, opened 2019-03-19
release candidate 0.11.0

moog16:


moog16:
CLA is ok - since all other branches passed.

PR #751, opened 2019-03-15
feat(menu-surface): upgrade to mdc web v1

moog16:
related #697

PR #762, opened 2019-03-20
feat(menu-surface): upgrade to mdc web v1

moog16:
reopening from #751
related #697


packages/menu-surface/index.tsx
```diff
@@ -253,62 +250,60 @@ class MenuSurface extends React.Component<MenuSurfaceProps, MenuSurfaceState> {
       getWindowScroll: () => {
         return {x: window.pageXOffset, y: window.pageYOffset};
       },
-      setPosition: (position: Position) => {
+      setPosition: (position: Partial<MDCMenuDistance>) => {
```
abhiomkar: Previously, `Position` had properties with string type. but `MDCMenuDistance` has properties with number type. You might need to append `px` when setting style.

packages/menu-surface/index.tsx
```diff
@@ -217,15 +212,16 @@ class MenuSurface extends React.Component<MenuSurfaceProps, MenuSurfaceState> {
         this.previousFocus.focus();
       },
       isFirstElementFocused: () =>
-        this.firstFocusableElement &&
+        Boolean(this.firstFocusableElement) &&
```
abhiomkar: Does't typescript has sugar syntax for null checking? 

`this.firstFocusableElement? === document.activeElement`

packages/menu-surface/index.tsx
```diff
@@ -253,62 +250,60 @@ class MenuSurface extends React.Component<MenuSurfaceProps, MenuSurfaceState> {
...
-          if (!this.menuSurfaceElement) return false;
-          if (!this.menuSurfaceElement.current) return false;
-          return window
-            .getComputedStyle(this.menuSurfaceElement.current)
-            .getPropertyValue('direction') === 'rtl';
-        },
-        setTransformOrigin: (transformOrigin: string) => this.setState({transformOrigin}),
+    return {
+      addClass: (className: string) => {
+        const classList = new Set(this.state.classList);
```
abhiomkar: if `this.state.classList` is already of Set type why do we need to instantiate this with `Set` again?

packages/menu-surface/index.tsx
```diff
@@ -253,62 +250,60 @@ class MenuSurface extends React.Component<MenuSurfaceProps, MenuSurfaceState> {
...
       },
-      focusAdapterMethods,
-      dimensionAdapterMethods
-    );
+      removeClass: (className: string) => {
+        const classList = new Set(this.state.classList);
+        classList.delete(className);
+        this.setState({classList});
+      },
+      hasClass: (className: string) => this.classes.split(' ').includes(className),
```
abhiomkar: This should use `classList` (Set) state for consistency.

packages/menu-surface/index.tsx
```diff
@@ -253,62 +250,60 @@ class MenuSurface extends React.Component<MenuSurfaceProps, MenuSurfaceState> {
...
+      hasAnchor: () => !!this.props.anchorElement,
+      notifyOpen: () => {
+        if (this.registerWindowClickListener) {
```
abhiomkar: Why do we need to register window listener here? Seems like this adapter doing more stuff than it should. We're not doing it in vanilla component implementation?

packages/menu-surface/index.tsx
```diff
@@ -253,62 +250,60 @@ class MenuSurface extends React.Component<MenuSurfaceProps, MenuSurfaceState> {
...
+        this.props.onOpen!();
+      },
+      notifyClose: () => {
+        if (this.deregisterWindowClickListener) {
+          this.deregisterWindowClickListener();
+        }
+        this.props.onClose!();
+      },
+      isElementInContainer: (el: HTMLElement) => {
+        if (!this.menuSurfaceElement.current) return false;
```
abhiomkar: `.contains` also includes the root element so you don't need to compare it with root yourself.

The adapter implementation can be shortened to:

```
this.menuSurfaceElement!.current.contains(el);
```

packages/menu-surface/index.tsx
```diff
@@ -253,62 +250,60 @@ class MenuSurface extends React.Component<MenuSurfaceProps, MenuSurfaceState> {
...
+        if (this.menuSurfaceElement.current === el) {
+          return true;
+        }
+        return this.menuSurfaceElement.current.contains(el);
+      },
+      isRtl: () => {
+        if (!this.menuSurfaceElement) return false;
```
abhiomkar: I've seen many places that you're checking if `this.menuSurfaceElement` is null or undefined. Would this every be null or undefined? Can we not use Non-null assertion operator? (`this.menuSurfaceElement!.current`)

packages/menu-surface/index.tsx
moog16: Good catch! It was actually working this way before (as a number). I also verified that it is automatically added on to the value when rendered. Do you think I should still add `px`?

packages/menu-surface/index.tsx
moog16: actually I don't even need the Boolean() part. `firstFocusableElement` and `last` will be an HTMLElement or null. Which means that in all cases `this.firstFocusableElement === document.activeElement` will be valid.

packages/menu-surface/index.tsx
moog16: If we just add to the existing Set, React will not detect a change meaning there won't be a re-render. React does not do a deep comparison.

packages/menu-surface/index.tsx
moog16: So the naming might be causing confusion, but `classList` is from the foundationClasses (what I want to rename all of them to). `this.classes` contains all the classes that are on the menu-surface, including `mdc-menu-surface`,  `mdc-menu-surface--fixed`, and anything the foundation or user add to the component.

packages/menu-surface/index.tsx
moog16: It actually gets registered during the OPENED_EVENT `this.listen(strings.OPENED_EVENT, this.registerBodyClickListener_);`, which eventually calls the `#adapter.notifyOpen` method. Since we don't listen to the events, this is the closest we can get to the vanilla component. What are your thoughts?

packages/menu-surface/index.tsx
moog16: good to know!

packages/menu-surface/index.tsx
moog16: There is a time when this could be undefined/null, which is before componentDidMount, and after componentWillUnmount. So assuming that it is `non-null` is not accurate.

packages/menu-surface/index.tsx
abhiomkar: I feel it is clearer if we encapsulate these event registrations.

```
notifyOpen: () => {
  this.handleOpen_();
  this.props.onOpen!();
}
```

So you can put more stuff in `handleOpen_` in feature if need to be. Same for `notifyClose`.

WDYT?

packages/menu-surface/index.tsx
moog16: It must be, but I couldn't fin any documentation on this.

packages/menu-surface/index.tsx
moog16: Ya that would clean up the adapter. Sounds good to me

moog16:
closing for #774

PR #769, opened 2019-03-24
fix(checkbox): upgrade mdc-web to v1

gugu:


moog16:
related #697

gugu:
image

It was already implemented before

packages/checkbox/index.tsx
```diff
@@ -45,20 +45,22 @@ interface CheckboxState {
...
 
   constructor(props: CheckboxProps) {
     super(props);
     this.state = {
       'checked': props.checked,
       'indeterminate': props.indeterminate,
       'classList': new Set(),
-      'aria-checked': false,
+      'aria-checked': 'false',
+      'disabled': props.disabled!,
```
moog16: what is this new state.disabled variable for?

moog16:
Sorry I didn't realize that we had already update checkbox :)

packages/checkbox/index.tsx
gugu: Adapter now has `setNativeControlDisabled: (disabled) => this.updateState('disabled', disabled),`, as I understand mdc-web now can change disabled state

moog16: ah ok - but we don't use it in the render method.

gugu: done

packages/checkbox/index.tsx
```diff
@@ -91,7 +94,10 @@ export class Checkbox extends React.Component<CheckboxProps, CheckboxState> {
       this.handleChange(checked!, indeterminate!);
     }
     if (disabled !== prevProps.disabled) {
-      this.foundation.setDisabled(disabled);
+      this.setState({
```
moog16: @gugu I dont think we need to setState here. foundation.setDisabled() will call `setNativeControlDisabled`, which updates disabled for us.

packages/checkbox/package.json
```diff
@@ -16,7 +16,7 @@
     "url": "https://github.com/material-components/material-components-web-react.git"
   },
   "dependencies": {
-    "@material/checkbox": "^0.41.0",
+    "@material/checkbox": "^1.0.0",
```
moog16: 1.1.0

PR #778, opened 2019-03-27
fix(typography): upgrade to mdc-web v1

gugu:
Related #697


moog16:
my mistake...its only on 1.0.0

PR #779, opened 2019-03-27
fix(dialog): mdc-web v1 upgrade

gugu:
Related #697

### DIFF
--- a/package.json
+++ b/package.json
@@ -66,7 +66,7 @@
     "@material/card": "^0.41.0",
     "@material/checkbox": "^0.41.0",
     "@material/chips": "^1.0.0",
-    "@material/dialog": "^0.43.0",
+    "@material/dialog": "^1.0.0",
     "@material/dom": "^0.41.0",
     "@material/drawer": "^1.0.1",
     "@material/fab": "^0.41.0",
--- a/package.json
+++ b/package.json
@@ -66,7 +66,7 @@
     "@material/card": "^0.41.0",
     "@material/checkbox": "^0.41.0",
     "@material/chips": "^1.0.0",
-    "@material/dialog": "^1.0.0",
+    "@material/dialog": "^1.1.0",
     "@material/dom": "^0.41.0",
     "@material/drawer": "^1.0.1",
     "@material/fab": "^0.41.0",
--- a/packages/dialog/index.tsx
+++ b/packages/dialog/index.tsx
@@ -23,10 +23,10 @@
 import * as React from 'react';
 import classnames from 'classnames';
 
-// @ts-ignore no .d.ts file
-import {MDCDialogFoundation, MDCDialogAdapter, util} from '@material/dialog/dist/mdc.dialog';
-// @ts-ignore no .d.ts file
-import {ponyfill} from '@material/dom/dist/mdc.dom';
+import {MDCDialogFoundation} from '@material/dialog/foundation';
+import {MDCDialogAdapter} from '@material/dialog/adapter';
+import {createFocusTrapInstance, isScrollable, areTopsMisaligned} from '@material/dialog/util';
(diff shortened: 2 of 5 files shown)


packages/dialog/index.tsx
```diff
@@ -136,15 +136,15 @@ class Dialog<T extends HTMLElement = HTMLElement> extends React.Component<
     const {open, autoStackButtons, escapeKeyAction, scrimClickAction} = this.props;
 
     if (prevProps.autoStackButtons !== autoStackButtons) {
-      this.foundation.setAutoStackButtons(autoStackButtons);
+      this.foundation.setAutoStackButtons(autoStackButtons!);
     }
 
     if (prevProps.escapeKeyAction !== escapeKeyAction) {
-      this.foundation.setEscapeKeyAction(escapeKeyAction);
+      this.foundation.setEscapeKeyAction(escapeKeyAction!);
```
moog16: `escapeKeyAction` doesn't have a defaultProp...We can't use `!` here.

packages/dialog/index.tsx
```diff
@@ -136,15 +136,15 @@ class Dialog<T extends HTMLElement = HTMLElement> extends React.Component<
...
     }
 
     if (prevProps.scrimClickAction !== scrimClickAction) {
-      this.foundation.setScrimClickAction(scrimClickAction);
+      this.foundation.setScrimClickAction(scrimClickAction!);
```
moog16: Same with scrimClickAction...we can't use `!`

packages/dialog/package.json
```diff
@@ -11,8 +11,8 @@
     "dialog"
   ],
   "dependencies": {
-    "@material/dialog": "0.43.0",
-    "@material/dom": "^0.41.0",
+    "@material/dialog": "1.0.1",
```
moog16: can you update these to 1.1.0. Might also need to do that in the other components.

package.json
```diff
@@ -66,7 +66,7 @@
     "@material/card": "^0.41.0",
     "@material/checkbox": "^0.41.0",
     "@material/chips": "^1.0.0",
-    "@material/dialog": "^0.43.0",
+    "@material/dialog": "^1.0.0",
```
moog16: update to v1.1.0

packages/dialog/index.tsx
```diff
@@ -60,15 +61,15 @@ export interface DialogProps<T,
     | React.ReactElement<TitleProps>
   );
   className?: string;
-  escapeKeyAction?: string;
+  escapeKeyAction: string;
```
moog16: make this optional ?

packages/dialog/index.tsx
```diff
@@ -60,15 +61,15 @@ export interface DialogProps<T,
...
   id?: string;
   onClose?: (action: string) => void;
   onClosing?: (action: string) => void;
   onOpen?: () => void;
   onOpening?: () => void;
   open?: boolean;
   role?: 'alertdialog' | 'dialog';
-  scrimClickAction?: string;
+  scrimClickAction: string;
```
moog16: optional

packages/dialog/index.tsx
```diff
@@ -103,6 +104,8 @@ class Dialog<T extends HTMLElement = HTMLElement> extends React.Component<
     id: 'mdc-dialog',
     open: false,
     role: 'alertdialog',
+    escapeKeyAction: strings.CLOSE_ACTION,
```
moog16: Can you document these in the Readme, and also document the defaults.

gugu: Defaults are already written in readme. Do I need to add anything extra?

moog16: Looks good!

PR #780, opened 2019-03-27
fix(top-app-bar): mdc-web v1 upgrade

gugu:
Related #697

### DIFF
--- a/packages/top-app-bar/index.tsx
+++ b/packages/top-app-bar/index.tsx
@@ -30,9 +30,9 @@ import TopAppBarIcon from './Icon';
 import {cssClasses} from './constants';
 import {MDCFixedTopAppBarFoundation} from '@material/top-app-bar/fixed/foundation';
 import {MDCTopAppBarAdapter} from '@material/top-app-bar/adapter';
-import MDCTopAppBarFoundation from '@material/top-app-bar/foundation';
+import {MDCTopAppBarFoundation} from '@material/top-app-bar/standard/foundation';
 import {MDCShortTopAppBarFoundation} from '@material/top-app-bar/short/foundation';
-import {EventType, SpecificEventListener} from '@material/base/types';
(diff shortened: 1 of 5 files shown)


packages/top-app-bar/index.tsx
```diff
@@ -219,6 +219,33 @@ class TopAppBar<T extends HTMLElement = HTMLHeadingElement> extends React.Compon
         }
         return 0;
       },
+      registerResizeHandler: (handler: SpecificEventListener<'resize'>) => {
+        window.addEventListener('resize', handler);
+      },
+      deregisterResizeHandler: (handler: SpecificEventListener<'resize'>) => {
+        window.removeEventListener('resize', handler);
+      },
+      registerNavigationIconInteractionHandler: <K extends EventType>(type: K, handler: SpecificEventListener<K>) => {
```
moog16: For these three adapter methods regarding navigationIcon, can we not implement them (or just have them as empty bodies? And just add a comment linking to this issue: https://github.com/material-components/material-components-web/issues/2813.

No one using the components should need to do this, since they can just add an `onClick` method to the Icon.

packages/top-app-bar/index.tsx
```diff
@@ -219,6 +219,33 @@ class TopAppBar<T extends HTMLElement = HTMLHeadingElement> extends React.Compon
         }
         return 0;
       },
+      registerResizeHandler: (handler: SpecificEventListener<'resize'>) => {
```
moog16: Can you add 2 tests for the register/deregisterResizeHandler methods please :)

moog16:
@gugu I also noticed that the standard top app bar doesn't hide when you scroll down the page in the screenshot tests.

packages/top-app-bar/index.tsx
gugu: will write tomorrow

Issue #242, opened 2018-08-24
New select (drop down menu attached to a text field)


moog16:

moog16:
Merged with #697 enhanced select

moog16:
merged in #697 enhanced select.

PR #786, opened 2019-03-29
feat(menu): add component

moog16:
related #697 & https://github.com/material-components/material-components-web-react/issues/242

Updating enzyme since context API is not supported by older version. See https://github.com/airbnb/enzyme/issues/1553


packages/menu/index.tsx
```diff
@@ -0,0 +1,202 @@
...
+  }
+
+  componentWillUnmount() {
+    if (this.state.foundation) {
+      this.state.foundation.destroy();
+    }
+  }
+
+  get listElements(): Element[] {
+    if (!(this.menuListElement
```
lucasecdb: you don't need to check for `this.menuListElement`

packages/menu/MenuListItem.tsx
```diff
@@ -0,0 +1,54 @@
...
+// THE SOFTWARE.
+
+import * as React from 'react';
+import {ListItem, ListItemProps} from '@material/react-list';
+
+export interface MenuListItemProps<T> extends ListItemProps<T> {
+  children?: React.ReactNode;
+}
+
+class MenuListItem<T extends HTMLElement = HTMLElement> extends React.Component<MenuListItemProps<T>, {}> {
```
lucasecdb: this could be a functional component

packages/menu/index.tsx
```diff
@@ -0,0 +1,202 @@
...
+import {MDCMenuAdapter} from '@material/menu/adapter';
+import MenuSurface, {MenuSurfaceProps} from '@material/react-menu-surface';
+import MenuList, {MenuListProps} from './MenuList';
+import MenuListItem from './MenuListItem';
+
+const {cssClasses} = MDCMenuFoundation;
+
+export interface MenuProps extends Exclude<MenuSurfaceProps, 'ref'> {
+  children: React.ReactElement<MenuList>;
+  handleSelected?: (index: number, item: Element) => void;
```
lucasecdb: shouldn't the prop name be `onSelected`? just like `onChange` for an `input` element

packages/menu/MenuListItem.tsx
moog16: I actually need this to be a Compnonent to pass along the defaultProps. Any idea around that?

packages/menu/index.tsx
moog16: Ya that makes sense. However we use the word `handle` in a few other places. We should normalize that to `on` at some point.

packages/menu/MenuListItem.tsx
lucasecdb: yeah, I don't think we could reuse the default props from the list item :-/

moog16: The defaultProp is the `renderWithListItemProps`, so it can't live on the listItem. Is that what you are suggesting?

lucasecdb: no, my idea would be to use default parameters on the function (which is the recommended way to use default props on function component) like this

```tsx
const MenuListItem = <T extends HTMLElement = HTMLElement>({
  renderWithListItemProps = true,
  ...props
}) => {
  // etc
}
```

but then we would need to redeclare every prop on the `defaultProps` of the `ListItem`. does that make sense?

moog16: I tested it using this as the component: 
```
const MenuListItem: <T extends HTMLElement = HTMLElement>(props: MenuListItemProps<T>)
=> React.ReactElement<MenuListItemProps<T>> = ({
  renderWithListItemProps = true,
  role = 'menuitem',
  children,
  checkboxList = false,
  radioList = false,
  className = '',
  tabIndex = -1,
  onKeyDown = () => {},
  onClick = () => {},
  onFocus = () => {},
  onBlur = () => {},
  tag = 'li',
  ...props
}) => {

  return (
    <ListItem
      role={role}
      {...props}
    >
      {children}
    </ListItem>
  );
}

```

This actually doesn't carry the defaultProp to the ListItem. Is that what you were thinking?

lucasecdb: exactly, but then we have the duplication of code (one in the default parameters of the `MenuListItem` and other on the `defaultProps` of the `ListItem`), but if you are okay with that then it's all good :-)

moog16: It actually doesn't work as we think it would. So its not a viable solution.

lucasecdb: oh, I missed this, but shouldn't you also pass along the props defined? Like
```tsx
const MenuListItem: <T extends HTMLElement = HTMLElement>(props: MenuListItemProps<T>)
=> React.ReactElement<MenuListItemProps<T>> = ({
  renderWithListItemProps = true,
  role = 'menuitem',
  children,
  checkboxList = false,
  radioList = false,
  className = '',
  tabIndex = -1,
  onKeyDown = () => {},
  onClick = () => {},
  onFocus = () => {},
  onBlur = () => {},
  tag = 'li',
  ...props
}) => {

  return (
    <ListItem
      role={role}
      renderWithListItemProps={renderWithListItemProps}
      children={children}
      checkboxList={checkboxList}
      radioList={radioList}
      className={className}
      tabIndex={tabIndex}
      onKeyDown={onKeyDown}
      onClick={onClick}
      onFocus={onFocus}
      onBlur={onBlur}
      tag={tag}
      {...props}
    >
      {children}
    </ListItem>
  );
}
```

lucasecdb: or maybe passing on only what we want to "override" from the default props

```tsx
const MenuListItem: <T extends HTMLElement = HTMLElement>(props: MenuListItemProps<T>)
=> React.ReactElement<MenuListItemProps<T>> = ({
  renderWithListItemProps = true,
  role = 'menuitem',
  ...props
}) => {

  return (
    <ListItem
      role={role}
      renderWithListItemProps={renderWithListItemProps}
      {...props}
    >
      {children}
    </ListItem>
  );
}
```

lucasecdb: sorry I think this ended up being confusing 😅

but I believe this last snippet should achieve the correct behaviour

moog16: ah ya - of course I need to pass that along...however, I don't think we need to pass the same defaultProps to the ListItem...The reason I have `renderWithListItemProps` on the MenuListItem, is so that the List knows to pass along certain props to the MenuListItem. Does that make sense?

lucasecdb: oh, I see. I think that's kind a weird way to do it :smile:

can you add to the check of `element.type === ListItem` the `MenuListItem`?

moog16: Ya I realize that's weird...but it also seemed wierd that MenuListItem is referenced in List. What do you think?

moog16: I guess that is the better more foolproof way :| -- I'll wait to hear back from you if you think of something better.

lucasecdb: you can create a context in the `List` component and use the context consumer in both `ListItem` and `MenuListItem`, and in the context you can put the functions you need in the item. What do you think?

moog16:
Updating enzyme and adapter because of context API not supported in current version of enzyme. See https://github.com/airbnb/enzyme/issues/1553

packages/list/ListItem.tsx
```diff
@@ -81,6 +120,42 @@ export default class ListItem<T extends HTMLElement = HTMLElement> extends React
...
+      const tabIndex = this.props.getListItemInitialTabIndex!(index);
+      this.setState({tabIndex});
+    }
+  }
+
+  getIndex = (listElement: Element) => {
+    return this.listElements.indexOf(listElement);
+  }
+
+  handleClick = (e: React.MouseEvent<any>) => {
```
moog16: These are moved from list/index.tsx

packages/list/index.tsx
```diff
@@ -347,6 +356,24 @@ export default class List<T extends HTMLElement = HTMLElement> extends React.Com
...
 
+  onDestroy = (index: number) => {
+    const {listItemClassNames} = this.state;
+    delete listItemClassNames[index];
+    this.setState({listItemClassNames});
+  };
+
+  listItemProps = {
+    checkboxList: this.props.checkboxList,
+    radioList: this.props.radioList,
```
lucasecdb: this won't work for when the user changes these props, right? I think it would be better to place this in some method and use a [memoization helper](https://github.com/alexreardon/memoize-one#readme) to avoid creating this object on every render

lucasecdb: highly recommended reading [this article](https://overreacted.io/writing-resilient-components/) if you haven't yet :-)

moog16: ah right - this won't be updated. Is this object created every render? Does that article go over it? I thought having it on the class would make sure its the same object every render.

lucasecdb: no, the object itself is only created once, but in order to update the props with the user, it should be at least created once per user change. does that make sense?

lucasecdb: btw is there an ETA for 0.12? I really wanted this component 😢

moog16: Yup that makes sense. I actually just read that article and it had some great points. Although it made me kinda sad since we follow a lot of the patterns that Dan suggested NOT to do...I think we're aware of them, as others have opened issues against some of these patterns. So we'll have to revisit a lot of these components moving forward.

RE listItemProps: Yes you're right. I think we can use the react.memo like you're saying, and recalculate this in the render.

RE ETA: No official ETA. The rate at which we're going, I don't see a release next week. So I'll update to the Tuesday after 4/30. How does that sound?

lucasecdb: 4/30 sounds good!

about the anti-patterns, we could compile a list of them and open an issue with a checklist, so it's easier to see which ones are missing the update. I would definitely help address them.

moog16: 4/30 is 7 business days...so about 1.5weeks. @TroyTae can you also agree to this date? We can push character count to another release.

Ya definitely... I will compile a list and add it as an issue / let you know.

packages/list/index.tsx
```diff
@@ -394,6 +384,23 @@ export default class List extends React.Component<ListProps, ListState> {
       tag: Tag,
       ...otherProps
     } = this.props;
+
+    const getListProps = (checkboxList?: boolean, radioList?: boolean) => ({
```
lucasecdb: I think this needs to be a class property, just like before. Defining this on render every time will reset the `memoize-one` cache 😔

moog16: ah! Thanks for catching that. I will update. Been working on the select. But will get back to this.

packages/list/index.tsx
TroyTae: @moog16 
Sorry late response.
I was planning to work on character count in last weekend.
But I had a sick, so I couldn't do nothing.
If you do not mind I want to push back count character release for stable.

moog16: No need to apologize. That sounds good to me.

Created an issue: https://github.com/material-components/material-components-web-react/issues/817

packages/list/index.tsx
```diff
@@ -364,6 +364,19 @@ export default class List extends React.Component<ListProps, ListState> {
     this.setState({listItemClassNames});
   };
 
+
+  getListProps = (checkboxList?: boolean, radioList?: boolean) => ({
```
lucasecdb: I believe you should place the call to `memoizeOne` directly here. If it's in render, the cache will still be reseted.

(sorry for commenting so much 😅)

lucasecdb: I think my last comment maybe wasn't clear enough, sorry

moog16: I believe you're right again sir! 

(please keep the comments coming)

moog16:
Tests are not passing because of esModuleInterop: false. https://github.com/material-components/material-components-web-react/pull/818 updates this to true. Tests should pass after this is merged.

moog16:
I've been running into this issue https://github.com/airbnb/enzyme/issues/2108. Added a postinstall script to hopefully fix it.

packages/list/ListItem.tsx
```diff
@@ -54,15 +54,54 @@ export default class ListItem<T extends HTMLElement = HTMLElement> extends React
     onBlur: () => {},
     onDestroy: () => {},
     tag: 'li',
+    handleClick: () => {},
```
williamernest: The naming here seems inconsistent. You're using `onBlur` and `handleClick`. I recommend being consistent with your event handler names (and generally prefer `on{EventName}`)

packages/list/ListItem.tsx
```diff
@@ -90,21 +165,51 @@ export default class ListItem<T extends HTMLElement = HTMLElement> extends React
       checkboxList,
       radioList,
       onDestroy,
+      onClick,
```
williamernest: I'm confused, what is the difference between onClick and handleClick ?

packages/list/ListItem.tsx
moog16: `onClick` is a native React click handler. I don't want to collide and override this behavior since its native to React. Therefore I must use a different name (handleClick) so that I don't collide with the naming.

packages/list/ListItem.tsx
moog16: see https://github.com/material-components/material-components-web-react/pull/786/files/f93bcdf660f50ff5a4ce61e65287b25520df2d1c#r279107245

williamernest: :+1:

packages/list/ListItem.tsx
williamernest: :+1:

PR #792, opened 2019-04-02
fix(icon-button): upgrade to mdc-web v1.1

ben-mckernan:
related #697

### DIFF
--- a/packages/icon-button/index.tsx
+++ b/packages/icon-button/index.tsx
@@ -23,8 +23,9 @@
 import * as React from 'react';
 import classnames from 'classnames';
 import * as Ripple from '@material/react-ripple';
-// @ts-ignore no mdc .d.ts file
-import {MDCIconButtonToggleFoundation} from '@material/icon-button/dist/mdc.iconButton';
+import {MDCIconButtonToggleAdapter} from '@material/icon-button/adapter';
+import {MDCIconButtonToggleFoundation} from '@material/icon-button/foundation';
+import {MDCIconButtonToggleEventDetail} from '@material/icon-button/types';
 import IconToggle from './IconToggle';
 const ARIA_PRESSED = 'aria-pressed';
 
@@ -35,6 +36,7 @@ interface ElementAttributes {
 
 type IconButtonTypes = HTMLButtonElement | HTMLAnchorElement;
 export interface IconButtonBaseProps extends ElementAttributes {
+  handleChange?: (evt: MDCIconButtonToggleEventDetail) => void;
   isLink?: boolean;
 };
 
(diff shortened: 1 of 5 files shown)


packages/icon-button/index.tsx
```diff
@@ -78,17 +81,18 @@ class IconButtonBase<T extends IconButtonTypes> extends React.Component<
...
   }
 
-  get adapter() {
+  get adapter(): MDCIconButtonToggleAdapter {
     return {
       addClass: (className: string) =>
         this.setState({classList: this.state.classList.add(className)}),
       removeClass: (className: string) => {
-        const classList = new Set(this.state.classList);
+        const {classList} = this.state;
```
ben-mckernan: This change is done because `hasClass` would otherwise return a false positive if it were called after this. This led to the `isOn` property in the event argument for `notifyChange` always being true.

This pattern seems to be used throughout this library and it is a little worrying as it goes against the first two points that react advises about for using state correctly: https://reactjs.org/docs/state-and-lifecycle.html#using-state-correctly

However, from what I can tell, it needs to work this way in order for `hasClass` to behave correctly.

test/unit/icon-button/index.test.tsx
```diff
@@ -62,13 +63,13 @@ test('aria-pressed is set false if passed as prop but on className is not passed
...
   wrapper.setState({classList: new Set(['test-class'])});
-  wrapper.hasClass('test-class');
+  assert.isTrue(wrapper.hasClass('test-class'));
 });
 
 test('#adapter.addClass adds a class to state.classList', () => {
   const wrapper = shallow<IconButton<HTMLButtonElement>>(<IconButton />);
   wrapper.instance().adapter.addClass('test-class');
-  wrapper.state().classList.has('test-class');
+  assert.isTrue(wrapper.state().classList.has('test-class'));
```
moog16: wow - these tests were basically useless...that's for fixing these.

packages/icon-button/index.tsx
moog16: Gotcha - ya it I realize that setState is async, and that it can become out of sync with hasClass. It does seem wrong that we are relying on changing state directly without calling setState. That is why we were calling `new Set()` to create a new instance. 

We could possibly make the call to `props.handleChange` call async, by calling it in the componentDidUpdate. We could use a boolean indicating `state.classList` is updating (let's call it `state.classListIsUpdating`. When `notifyChange` gets trigger, we could set another boolean `state.hasChangeOccured`. (hopefully you're following). The componentDidUpdate statement would look like:

```js
if (!state.classListIsUpdating && prevState.classListIsUpdating && state.hasChangeOccured) {
  this.props.handleChange(...);
  this.setState({hasChangeOccured: false}); // clean up to setup for the next notifyChange 
}
```

Just a thought as a fix for this. It definitely doesn't belong in this PR, but maybe we can open another issue to fix this.

moog16: And we should actually be consistent in the removeClass and addClass methods...
Created an issue https://github.com/material-components/material-components-web-react/issues/793

ben-mckernan: @moog16 thanks for the feedback and creating a new issue for this.

Issue #795, opened 2019-04-04
Dialog and drawer have different focus-trap dependency versions.


evankim390:
[Drawer depends on ^3.0.0](https://github.com/material-components/material-components-web-react/blob/rc0.12.0/packages/drawer/package.json)
[Dialog depends on ^4.0.2](https://github.com/material-components/material-components-web-react/blob/rc0.12.0/packages/dialog/package.json)

Would it be possible to sync the two packages to the same version? This would avoid pulling in multiple copies of the same library, and works slightly nicer with yarn (right now I have to manually resolve `focus-trap` to `^4.0.2`)

moog16:
@evankim390 - yes definitely...I didn't even realize that they were different, but I know how mismatched npm modules can break you environment. Would you be will to open a PR against rc0.12.0?

evankim390:
Sure thing https://github.com/material-components/material-components-web-react/pull/797

moog16:
Closing for #797

PR #797, opened 2019-04-05
chore(drawer): Update focus-trap dependency to ^4.0.2

evankim390:
Currently the drawer and dialog components use different focus-trap versions. This
pull request syncs them to the later one (^4.0.2)

https://github.com/material-components/material-components-web-react/issues/795


evankim390:
@googlebot Think I handled the registration issues.

PR #799, opened 2019-04-07
Correction to README file

aarnadlr:
Hello, there is an error in this documentation. At line 23, reader is advised to add:
```js
@import "@material/ripple/index.scss";
```

This file does not exist.

User should be advised to add:

```js
@import "@material/ripple/mdc-ripple.scss";
```

Thank you
a

### DIFF
--- a/packages/ripple/README.md
+++ b/packages/ripple/README.md
@@ -27,6 +27,7 @@ You'll also need to include these sass mixins on the element. Please also refer 
   @include mdc-ripple-surface;
   @include mdc-ripple-radius-bounded;
   @include mdc-states;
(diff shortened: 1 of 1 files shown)


aarnadlr:
Second commit to this PR:

At Line 30 of README documentation: Advise reader to add `overflow: hidden` property to `.ripple-icon-component` to render properly bounded ripple.

The current documentation leads the reader to create an unbounded ripple, extending outside the dimensions of the component (even when the `unbounded` prop is removed). The `overflow: hidden` property must be added to render the expected bounded ripple effect.

Thank you
a

PR #818, opened 2019-04-23
chore(infrastructure): add esmoduleinterop to tsconfig 

moog16:
needed for memoize-one not working with typescript issue https://github.com/alexreardon/memoize-one/issues/37. Adding esModuleInterop is also used in MDC Web.


packages/list/ListItem.tsx
```diff
@@ -35,6 +35,7 @@ export interface ListItemProps<T> extends React.HTMLProps<T> {
   activated?: boolean;
   selected?: boolean;
   onDestroy?: () => void;
+  ref?: React.Ref<any>;
```
patrickrodee: Why did this change?

packages/list/index.tsx
```diff
@@ -49,6 +49,7 @@ export interface ListProps<T extends HTMLElement> extends React.HTMLProps<HTMLEl
   wrapFocus?: boolean;
   tag?: string;
   children: ListItem<T> | ListItem<T>[] | React.ReactNode;
+  ref?: React.Ref<any>;
```
patrickrodee: Same question as above

packages/list/ListItem.tsx
moog16: ah good catch! This is actually a change from a future PR. This shouldn't be in here.

PR #819, opened 2019-04-27
chore: reduce unused deps in package.json

moog16:


package.json
```diff
@@ -65,32 +65,23 @@
     "@material/base": "^1.0.0",
     "@material/button": "^1.1.0",
     "@material/card": "^1.1.0",
+    "@material/chips": "^1.1.1",
```
williamernest: maybe this is wrong....

PR #823, opened 2019-04-27
feat(select): enhanced select

moog16:
related #697


packages/menu/index.tsx
```diff
@@ -194,8 +187,8 @@ export {
   ListDivider as MenuListDivider,
   ListGroup as MenuListGroup,
   ListGroupSubheader as MenuListGroupSubheader,
-  ListItemGraphic as MenuListGraphic,
-  ListItemMeta as MenuListMeta,
+  ListItemGraphic as MenuListItemGraphic,
```
moog16: change this in readme

packages/select/EnhancedSelect.tsx
```diff
@@ -0,0 +1,207 @@
...
+      isInvalid,
+    } = this.props;
+
+    const {'aria-expanded': ariaExpanded, selectedValue, selectedItem} = this.state;
+
+    const selectedTextAttrs: {[key: string]: string} = {};
+    if (required) {
+      selectedTextAttrs['aria-required'] = required.toString();
+    }
+    if (ariaExpanded && ariaExpanded !== 'false') {
```
moog16: convert 'false' and 'true' to constants.

moog16:
Running into https://github.com/Microsoft/TypeScript/issues/28762. Updating to typescript@3.3.3 seems to fix issue.

packages/select/NativeSelect.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+  get classes() {
+    return classnames('mdc-select__native-control', this.props.className);
+  }
+
+  attachRef = (node: HTMLSelectElement | null) => {
+    const {innerRef} = this.props;
+
+    // https://github.com/facebook/react/issues/13029#issuecomment-410002316
+    // @ts-ignore this is acceptable according to the comment above
```
williamernest: Can you replace `this is acceptable according to the comment above` with the reason why it's required ? Maybe `incorrectly typed as readonly - see github comment`

packages/select/README.md
```diff
@@ -49,11 +50,42 @@ class MyApp extends React.Component {
...
+}
+```
+
 #### Shorthand options
 
-If the option elements do not require anything unique, you can omit passing `this.props.children` and set the `option` attribute.
+If the option elements do not require anything unique, you can omit passing `this.props.children` and set the `options` prop.
 
-This shorthand option is most useful when you're select options come in the form of JSON. Typically when building a select, the options will be backed by an array of objects or strings (either from an API endpoint or a JSON File). If the label/value key names, you can run a `map` function over your options.
+This options shorthand is most useful when you're select options come in the form of JSON. Typically when building a select, the options will be backed by an array of objects or strings (either from an API endpoint or a JSON File). If the label/value key names, you can run a `map` function over your options.
```
williamernest: `you're` -> `your`
Typically when -> Typically, when

`If the label/value key names, you can run a map function over your options`...... what?

PR #824, opened 2019-04-27
feat(select): add helper text

moog16:
related #697

### DIFF
--- a/packages/select/helper-text/index.tsx
+++ b/packages/select/helper-text/index.tsx
@@ -1,0 +1,130 @@
+// The MIT License
+//
+// Copyright (c) 2019 Google, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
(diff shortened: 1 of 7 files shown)


packages/select/helper-text/index.tsx
```diff
@@ -0,0 +1,130 @@
...
+
+interface SelectHelperTextState extends ElementAttributes {
+  classList: Set<string>;
+};
+
+export class SelectHelperText extends React.Component<SelectHelperTextProps, SelectHelperTextState> {
+  foundation?: MDCSelectHelperTextFoundation;
+
+  state: SelectHelperTextState = {
+    'aria-hidden': undefined,
```
moog16: spoke offline - set to undefined since foundation should programmatically set this. Also need to define this to be of type SelectHelperTextState

packages/select/helper-text/index.tsx
```diff
@@ -0,0 +1,130 @@
...
+        // not implmenting because developer should would never call `setContent()`
+      },
+    };
+  }
+
+  render() {
+    const {
+      children,
+      /* eslint-disable no-unused-vars */
+      persistent,
```
moog16: destructuring these allows us to pass all ..otherProps without having to pass persistent, classname, etc.

PR #825, opened 2019-04-27
feat(select): add icon

moog16:
related #697

### DIFF
--- a/packages/select/icon/README.md
+++ b/packages/select/icon/README.md
@@ -20,4 +20,4 @@ const MyComponent = () => {
 
 Prop Name | Type | Description
 --- | --- | ---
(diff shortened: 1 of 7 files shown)


packages/select/icon/README.md
```diff
@@ -0,0 +1,23 @@
...
+    </SelectIcon>
+  );
+}
+```
+
+## Props
+
+Prop Name | Type | Description
+--- | --- | ---
+tag | keyof React.ReactHTML | Sets the element tag. Defaults to `<i />`.
```
moog16: | string (keyof React.ReactHTML) |

moog16: `i` becomes `<i />` or `div` becomes `<div />`

Issue #827, opened 2019-04-29
Refactor literal class name to constants


TroyTae:
- [x] Button (#838)

- [x] Card (#839)

- [x] Checkbox (#982)

- [x] Chips (#982)

- [x] Drawer (#982)

- [x] Fab (#844)

- [x] FloatingLabel (#845)

- [x] IconButton (#982)

- [x] LayoutGrid (#846)

- [x] LineRipple (#983)

- [x] LinearProgress (#983)

- [x] List (#983)

- [x] MaterialIcon (#983)

- [x] Menu (#983)

- [x] MenuSurface (#983)

- [ ] NotchedOutline

- [x] Radio (#983)

- [x] Select (#983)

- [x] SnackBar (#983)

- [x] Switch

- [x] TabBar

- [x] TabIndicator

- [x] TabScroller

- [x] Tab

- [x] Text Filed

- [x] TopAppBar

abhiomkar:
I created `feat/constants` branch on upstream for future PRs so we can merge to `master` branch at once. Thanks!

TroyTae:
@abhiomkar 
That's nice. Thank you!

TroyTae:
It's going to take a lot of time to merge https://github.com/material-components/material-components-web/pull/4761.
Close this issue and I'll reopen after merging!

PR #838, opened 2019-05-02
feat(button): define CSS_CLASSES

TroyTae:
related to https://github.com/material-components/material-components-web-react/issues/827


packages/button/constant.ts
```diff
@@ -0,0 +1,33 @@
+// The MIT License
+//
+// Copyright (c) 2018 Google, Inc.
```
moog16: 2019

PR #839, opened 2019-05-02
chore(card): define CSS_CLASSES

TroyTae:
related to #827


packages/card/constant.ts
```diff
@@ -0,0 +1,40 @@
+// The MIT License
+//
+// Copyright (c) 2018 Google, Inc.
```
moog16: 2019

Issue #224, opened 2018-08-16
Can only hide label using fullWidth mode


oste:
When using the TextField component it is only possible to hide the label using `fullWidth` option seen here https://github.com/material-components/material-components-web-react/blob/master/packages/text-field/index.js#L245

I propose that that the label should not be tied to this property and be independent on a new property such as `hideLabel`.

There is a full width example seen here https://material-components.github.io/material-components-web-catalog/#/component/text-field but it seems to just add a `width: 100%` to the `.mdc-text-field--fullwidth` class and not control the floating label. 

I think allowing the floating label to be toggled is definitely useful when using just a `placeholder` works best.

moog16:
label is no longer required as of v0.12.0. This shouldn't be an issue anylonger

PR #840, opened 2019-05-02
feat(text-field): add placeholder property

rbare1:
related to #224

I opened this PR to start a discussion around issue #224 that was opened last year. This feature branch added support for the `mdc-text-field--no-label` class, so it seems like a good time to support placeholder text when a label isn't present. The examples for text fields without labels on [Material IO](https://material.io/design/components/text-fields.html) display text fields without labels still having placeholder text.

### DIFF
--- a/packages/text-field/Input.tsx
+++ b/packages/text-field/Input.tsx
@@ -38,6 +38,7 @@ export interface InputProps<T extends HTMLElement = HTMLInputElement> {
   setDisabled?: (disabled: boolean) => void;
   setInputId?: (id: string | number) => void;
   handleFocusChange?: (isFocused: boolean) => void;
+  placeholder?: string;
 };
 
 type InputElementProps = Exclude<React.HTMLProps<HTMLInputElement>, 'ref'>;
@@ -268,6 +269,7 @@ export default class Input<T extends HTMLElement = HTMLInputElement> extends Rea
       onMouseDown,
       onTouchStart,
       onChange,
+      placeholder,
       /* eslint-enable no-unused-vars */
       ...otherProps
     } = this.props;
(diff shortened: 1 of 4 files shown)


packages/text-field/index.tsx
```diff
@@ -58,6 +58,7 @@ export interface Props<T extends HTMLElement = HTMLInputElement> {
   onTrailingIconSelect?: () => void;
   textarea?: boolean;
   trailingIcon?: React.ReactElement<React.HTMLProps<HTMLOrSVGElement>>;
+  floatLabel?: boolean;
```
moog16: I think this should be `noLabel`

packages/text-field/index.tsx
```diff
@@ -158,7 +161,7 @@ class TextField<T extends HTMLElement = HTMLInputElement> extends React.Componen
       // TODO change literal to constant
       'mdc-text-field--fullwidth': fullWidth,
       'mdc-text-field--with-trailing-icon': trailingIcon,
-      'mdc-text-field--no-label': !this.labelAdapter.hasLabel(),
+      'mdc-text-field--no-label': !this.labelAdapter.hasLabel() || !floatLabel,
```
moog16: || noLabel

packages/text-field/Input.tsx
```diff
@@ -38,6 +38,7 @@ export interface InputProps<T extends HTMLElement = HTMLInputElement> {
   setDisabled?: (disabled: boolean) => void;
   setInputId?: (id: string | number) => void;
   handleFocusChange?: (isFocused: boolean) => void;
+  placeholder?: string;
```
moog16: no change is needed as `placeholder` is part of `InputElementProps`

rbare1:
I won't get to this this week, so I'll reopen it later when I've made the changes. Thanks for the feedback!

PR #846, opened 2019-05-03
 feat(layout-grid): define CSS_CLASSES & DEVICE_TYPE

TroyTae:
related to #827

### DIFF
--- a/packages/layout-grid/constant.ts
+++ b/packages/layout-grid/constant.ts
@@ -1,0 +1,34 @@
+// The MIT License
+//
+// Copyright (c) 2018 Google, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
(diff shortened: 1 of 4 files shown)


packages/layout-grid/constant.ts
```diff
@@ -0,0 +1,40 @@
...
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+const CSS_CLASSES = {
```
moog16: These belong in the MDC Web lib, and we should import from there.

packages/layout-grid/constant.ts
```diff
@@ -0,0 +1,40 @@
+// The MIT License
+//
+// Copyright (c) 2018 Google, Inc.
```
moog16: 2019

packages/layout-grid/constant.ts
moog16: change to cssClasses

PR #810, opened 2019-04-15
fix(fab): Made initRipple and unbounded props optional

Maxim-Mazurok:
This PR is a follow up to this conversation: https://github.com/material-components/material-components-web-react/commit/b62c36fbcee3e6aecab00ca949a93080f4254719#r33164756

`initRipple` and `unbounded` props should be optional.

PR #854, opened 2019-05-06
fix(fab): Made initRipple and unbounded props optional

Maxim-Mazurok:
Reopening #810 

> This PR is a follow up to this conversation: b62c36f#r33164756
> 
> `initRipple` and `unbounded` props should be optional.

### DIFF
--- a/packages/fab/index.tsx
+++ b/packages/fab/index.tsx
@@ -33,8 +33,8 @@ export interface FabProps extends InjectedProps<HTMLButtonElement>,
     icon?: React.ReactElement<HTMLElement>;
     textLabel?: string;
     className?: string;
-    initRipple: React.Ref<HTMLButtonElement>;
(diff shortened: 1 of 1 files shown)


Maxim-Mazurok:
rc0.13.0 branch is also failing on screenshot tests on my local environment without this PR changes, so I'll just wait till tests are fixed in rc0.13.0.
The only error in Travis was "Error: Timeout of 60000ms exceeded", so the test might be flaky and rerunning build might fix the issue.

moog16:
Just restarted it. Sometimes travisCI gets slow. And sometimes we need to add some more time to the startup of the webpack server. I hope it doesn't come to that, as we're waiting around 45 seconds for startup now.

moog16:
Thanks @Maxim-Mazurok !!

PR #855, opened 2019-05-07
fix: ClientRect/DOMRect is not IE11 compatible

### LINKED ISSUE #853: ClientRect is not defined
I've noticed that v.0.12 uses `new ClientRect()` in some places...

Well, I've been attempting to implement `mdc-slider` in React, and tried returning a `new ClientRect()` and got a `ClientRect is not defined` error.

Googling took me to: https://github.com/Microsoft/TypeScript/issues/18615

It seems that while TypeScript's typings still have `ClientRect`, Chrome has replaced it with `DOMRect` .

Further, `DOMRect` itself is experimental according to MDN: https://developer.mozilla.org/en-US/docs/Web/API/DOMRect

I'm not sure if these are getting properly polyfilled by Webpack and Babel either.

Was there a specific reason to switch from the explicit `{ top: 0, right: 0, bottom: 0, left: 0, width: 0, height: 0 }` to `new ClientRect()`?

cpboyd:
Fixes #853.

@moog16 I copied the usage from `ripple` and `tab-bar` so that it'd be easier to do a find & replace in the future.

Let me know if there's anything else I need to do.

### DIFF
--- a/packages/chips/Chip.tsx
+++ b/packages/chips/Chip.tsx
@@ -134,12 +134,38 @@ export class Chip extends React.Component<ChipProps, ChipState> {
           .getPropertyValue(propertyName);
       },
       getRootBoundingClientRect: () => {
-        if (!this.chipElement) return new ClientRect();
+        if (!this.chipElement) {
+          // new DOMRect is not IE11 compatible
+          const defaultDOMRect = {
+            bottom: 0,
+            height: 0,
+            left: 0,
+            right: 0,
+            top: 0,
+            width: 0,
+            x: 0,
+            y: 0,
+          };
+          return defaultDOMRect;
+        }
         return this.chipElement.getBoundingClientRect();
       },
       getCheckmarkBoundingClientRect: () => {
         const {chipCheckmark} = this.props;
-        if (!chipCheckmark) return new ClientRect();
+        if (!chipCheckmark) {
+          // new DOMRect is not IE11 compatible
+          const defaultDOMRect = {
+            bottom: 0,
+            height: 0,
+            left: 0,
+            right: 0,
+            top: 0,
+            width: 0,
+            x: 0,
+            y: 0,
+          };
+          return defaultDOMRect;
+        }
         return chipCheckmark.props.getBoundingClientRect();
       },
       setStyleProperty: (propertyName: keyof React.CSSProperties, value: string | null) => {
--- a/packages/chips/Chip.tsx
+++ b/packages/chips/Chip.tsx
@@ -152,7 +152,7 @@ export class Chip extends React.Component<ChipProps, ChipState> {
       },
       getCheckmarkBoundingClientRect: () => {
         const {chipCheckmark} = this.props;
-        if (!chipCheckmark) {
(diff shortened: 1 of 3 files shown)


packages/tab-scroller/index.tsx
```diff
@@ -151,14 +151,40 @@ export default class TabScroller extends React.Component<
...
-          this.areaElement.current.getBoundingClientRect() :
-          new ClientRect(),
-      computeScrollContentClientRect: () =>
-        this.contentElement.current ?
-          this.contentElement.current.getBoundingClientRect() :
-          new ClientRect(),
+      computeScrollAreaClientRect: () => {
+        if (!this.areaElement.current) {
+          // new DOMRect is not IE11 compatible
+          const defaultDOMRect = {
```
moog16: Could you make this a variable that both `computeScrollAreaClientRect` and `computeScrollContentClientRect` reference, instead of defining it twice?

cpboyd: I created a function `getBoundingClientRectOf` within `tab-scroller`.

I considered a shared variable (either within the class or the `adapter`) but if code happened to modify the returned object, then it would result in these functions no longer returning the expected default object.

packages/chips/Chip.tsx
```diff
@@ -134,12 +134,38 @@ export class Chip extends React.Component<ChipProps, ChipState> {
...
+            y: 0,
+          };
+          return defaultDOMRect;
+        }
         return this.chipElement.getBoundingClientRect();
       },
       getCheckmarkBoundingClientRect: () => {
         const {chipCheckmark} = this.props;
-        if (!chipCheckmark) return new ClientRect();
+        if (!chipCheckmark) {
```
cpboyd: @moog16 Should this `if` also check `chipCheckmark.props.getBoundingClientRect`?

moog16: Yes, that doesn't hurt to have the check there.

moog16: I won't merge it if you want to update this...you may also want to check unit tests.

packages/tab-scroller/index.tsx
```diff
@@ -151,19 +151,35 @@ export default class TabScroller extends React.Component<
...
+      },
+      computeScrollContentClientRect: () => {
+        return this.getBoundingClientRectOf(this.contentElement.current);
+      },
       computeHorizontalScrollbarHeight: () =>
         computeHorizontalScrollbarHeight(document),
     };
   }
 
+  getBoundingClientRectOf = (element: HTMLElement) => {
```
moog16: This looks good.

packages/chips/Chip.tsx
cpboyd: Added new commit.

Couldn't run tests due to fsevents failing to build on Node v12 on Mac.

I'll try to see about switching to nvm tomorrow and running the tests.

cpboyd:
@moog16 The new commit fixed the unit tests.

Let me know if you want me to rebase these changes down to a single commit.

PR #858, opened 2019-05-08
Release rc0.13.0 for pach release v0.12.1

moog16:


packages/button/constant.ts
```diff
@@ -0,0 +1,33 @@
...
+  ROOT: 'mdc-button',
+  ICON: 'mdc-button__icon',
+  LABEL: 'mdc-button__label',
+  DENSE: 'mdc-button--dense',
+  RAISED: 'mdc-button--raised',
+  OUTLINED: 'mdc-button--outlined',
+  UNELEVATED: 'mdc-button--unelevated',
+};
+
+export {CSS_CLASSES};
```
abhiomkar: `cssClasses`?

moog16: I spoke to @williamernest about this. We decided that we should follow what is in the Google TypeScript style guide, which states enums should be all uppercase and `_` separated.

PR #863, opened 2019-05-15
chore: add prettier

### LINKED ISSUE #860: Prettier
can I add a prettier config for more consistent code style? does Google have a default configuration we could use?

lucasecdb:
resolves #860 

I think its best to review this pr by going through each commit individually:

46c7dd2: adds the prettier configuration files, removes conflicting rules in the eslintrc, and switch the parser to `@typescript-eslint/parser`.

ff73061: since the parser came with its own `no-unused-vars` rule, I needed to update all `eslint-disable`s of the rule with this one, which is `@typescript-eslint/no-unused-vars` (I also disabled the `no-unused-vars` in the first commit)

a665c2d: this fixes some eslint issues that came with the new parser that weren't treated before. basically it's just adding a few more `eslint-disable`s to the `@typescript-eslint/no-unused-vars` rule and switch some `let` variables with `const`.

ea30ab1: I took the liberty of adding the `skipUndeclared` option to the `react/prop-types` rule, which makes it skip checking components that don't declare the static `propTypes` property, so I removed all disables of this rule since its not needed anymore.

139c675: this is the actual prettier changes, this commit consist only of running `npm run fix` and a one line fix on the material-icons package (because it moved the comment with a eslint disable comment)


moog16:
@lucasecdb is it possible to make them 5 different PRs? Then we can merge it into 1 feature branch?

package.json
```diff
@@ -149,7 +154,6 @@
     "ts-loader": "^3.5.0",
     "ts-node": "^7.0.1",
     "typescript": "^3.3.3",
-    "typescript-eslint-parser": "^21.0.1",
```
moog16: nice - this was deprecated https://github.com/eslint/typescript-eslint-parser!

.eslintrc.json
```diff
@@ -1,39 +1,36 @@
...
+    "no-unused-vars": "off",
+    "@typescript-eslint/no-unused-vars": "warn",
     "strict": 0,
     "require-jsdoc": "off",
     "valid-jsdoc": "off",
     "switch-colon-spacing": 0,
-    "max-len": ["error", 120],
-    "indent":  ["error", 2, {"SwitchCase":1}],
+    "max-len": "off",
+    "indent": "off",
```
moog16: now handled by prettier (2 by default)

.prettierrc
```diff
@@ -0,0 +1,7 @@
+{
```
moog16: I think we might need to turn this on https://prettier.io/docs/en/options.html#arrow-function-parentheses

lucasecdb:
is it ok to lower the `printWidth` to 80 (the default value)? I think 120 it making the lines too  long

moog16:
@lucasecdb thanks for pushing those changes. Can you also add a precommit to run npm run lint?

moog16:
The precommit is slow...can we change this?

lucasecdb:
sure, I'll move that to pre-push

moog16:
confirmed validate-message still works. I'll merge once tests pass

Issue #258, opened 2018-09-03
Slider?


madeleineostoja:
I know that aiming for parity with mainline MDC web is an ongoing project, but just wondering if there's any immediate plans to add a wrapper for the Slider component? One of the more useful MDC components IMO since native range inputs are such a hassle to style, and one of the more beneficial to have a prebuilt React wrapper for, since it's not just a plain UI component.

moog16:
@seaneking forgive the late response. We're actually working on building a new/improved slider in MDC Web (https://github.com/material-components/material-components-web/tree/feat/slider). Once this is complete it'll most likely be put on the roadmap.

PR #868, opened 2019-05-19
add slider

cpboyd:
I've been using this in my own project and felt I've gotten it to a point worth sharing.

I just updated the track markers portion to be more React-oriented. Previously, I just had the functions doing manual DOM manipulations, since I wasn't actually using discrete sliders in my project.

I still need to update the README, specifically in relation to the Props section.

### DIFF
--- a/packages/slider/index.scss
+++ b/packages/slider/index.scss
@@ -1,0 +1,23 @@
+// The MIT License
+//
+// Copyright (c) 2019 Google, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+@import '@material/slider/mdc-slider';
--- a/packages/slider/index.tsx
+++ b/packages/slider/index.tsx
@@ -11,16 +11,16 @@ type MouseLikeEvent = React.MouseEvent | React.PointerEvent | React.TouchEvent;
 
 export interface SliderProps extends React.HTMLProps<HTMLDivElement> {
   value: number;
-  discrete?: boolean;
-  displayMarkers?: boolean;
+  discrete: boolean;
+  displayMarkers: boolean;
   min: number;
   max: number;
-  step?: number;
-  disabled?: boolean;
+  step: number;
+  disabled: boolean;
   ariaLabel: string;
-  isRtl?: boolean;
-  onValueChange?: (value: number) => void;
-  onValueInput?: (value: number) => void;
+  isRtl: boolean;
+  onValueChange: (value: number) => void;
+  onValueInput: (value: number) => void;
 }
 
 interface SliderState {
@@ -34,6 +34,33 @@ interface SliderState {
 type SliderElementNames = 'trackStyleProperty' | 'thumbContainerStyleProperty';
 
 class Slider extends React.Component<SliderProps, SliderState> {
+  foundation!: MDCSliderFoundation;
+  sliderElement: React.RefObject<HTMLDivElement> = React.createRef();
+
+  static defaultProps: Partial<SliderProps> = {
+    ariaLabel: 'Select Value',
+    className: '',
+    open: false,
+    min: 0,
+    max: 100,
+    value: 0,
+    step: 0,
+    discrete: false,
+    displayMarkers: false,
+    disabled: false,
+    tabIndex: 0,
+    isRtl: false,
+    onValueChange: () => {},
+    onValueInput: () => {},
+  };
+
+  state: SliderState = {
+    classList: new Set(),
+    markerStyles: [],
+    trackStyleProperty: {},
+    thumbContainerStyleProperty: {},
+  };
+
   get classes() {
     const {classList} = this.state;
     const {className, discrete, displayMarkers} = this.props;
@@ -47,12 +74,12 @@ class Slider extends React.Component<SliderProps, SliderState> {
     return {
       hasClass: (className: string): boolean =>
         this.classes.split(' ').includes(className),
-      addClass: (className: string): void => {
+      addClass: (className: string) => {
         const {classList} = this.state;
         classList.add(className);
         this.setState({classList});
       },
-      removeClass: (className: string): void => {
+      removeClass: (className: string) => {
         const {classList} = this.state;
         classList.delete(className);
         this.setState({classList});
@@ -63,13 +90,13 @@ class Slider extends React.Component<SliderProps, SliderState> {
         }
         return this.sliderElement.current.getAttribute(name);
       },
-      setAttribute: (name: string, value: string): void => {
+      setAttribute: (name: string, value: string) => {
         if (!this.sliderElement.current) {
           return;
         }
         this.sliderElement.current.setAttribute(name, value);
       },
-      removeAttribute: (name: string): void => {
+      removeAttribute: (name: string) => {
         if (!this.sliderElement.current) {
           return;
         }
@@ -92,21 +119,19 @@ class Slider extends React.Component<SliderProps, SliderState> {
         }
         return this.sliderElement.current.getBoundingClientRect();
       },
-      getTabIndex: (): number => this.props.tabIndex || 0,
+      getTabIndex: (): number => this.props.tabIndex!,
       // React handles events differently:
       registerInteractionHandler: () => undefined,
       deregisterInteractionHandler: () => undefined,
       registerThumbContainerInteractionHandler: (
         evtType: EventType,
         handler: SpecificEventListener<EventType>
-      ): void => {
+      ) => {
         if (evtType === this.transitionendEvtName) {
           this.onThumbEnd = handler as any;
         }
       },
-      deregisterThumbContainerInteractionHandler: (
-        evtType: EventType
-      ): void => {
+      deregisterThumbContainerInteractionHandler: (evtType: EventType) => {
         if (evtType === this.transitionendEvtName) {
           this.onThumbEnd = undefined;
         }
@@ -114,44 +139,39 @@ class Slider extends React.Component<SliderProps, SliderState> {
       registerBodyInteractionHandler: (
         evtType: EventType,
         handler: SpecificEventListener<EventType>
-      ): void => {
+      ) => {
         document.body.addEventListener(evtType, handler);
       },
       deregisterBodyInteractionHandler: (
         evtType: EventType,
         handler: SpecificEventListener<EventType>
-      ): void => {
+      ) => {
         document.body.removeEventListener(evtType, handler);
       },
-      registerResizeHandler: (
-        handler: SpecificEventListener<'resize'>
-      ): void => {
+      registerResizeHandler: (handler: SpecificEventListener<'resize'>) => {
         window.addEventListener('resize', handler);
       },
-      deregisterResizeHandler: (
-        handler: SpecificEventListener<'resize'>
-      ): void => {
+      deregisterResizeHandler: (handler: SpecificEventListener<'resize'>) => {
         window.removeEventListener('resize', handler);
       },
-      notifyInput: (): void => {
+      notifyInput: () => {
         const callback = this.props.onValueInput;
         if (typeof callback === 'function') {
           callback(this.foundation.getValue());
         }
       },
-      notifyChange: (): void => {
+      notifyChange: () => {
         const callback = this.props.onValueChange;
         if (typeof callback === 'function') {
           callback(this.foundation.getValue());
         }
       },
-      setThumbContainerStyleProperty: (prop: string, value: string): void =>
+      setThumbContainerStyleProperty: (prop: string, value: string) =>
         this.setStyleToElement(prop, value, 'thumbContainerStyleProperty'),
-      setTrackStyleProperty: (prop: string, value: string): void =>
+      setTrackStyleProperty: (prop: string, value: string) =>
         this.setStyleToElement(prop, value, 'trackStyleProperty'),
-      setMarkerValue: (value: number): void =>
-        this.setState({markerValue: value}),
-      appendTrackMarkers: (numMarkers: number): void => {
+      setMarkerValue: (value: number) => this.setState({markerValue: value}),
+      appendTrackMarkers: (numMarkers: number) => {
         this.setState((prevState) => {
           let {markerStyles} = prevState;
           // Shallow copy of array:
@@ -162,13 +182,13 @@ class Slider extends React.Component<SliderProps, SliderState> {
           return {markerStyles};
         });
       },
-      removeTrackMarkers: (): void => {
+      removeTrackMarkers: () => {
         this.setState({markerStyles: []});
       },
       setLastTrackMarkersStyleProperty: (
         propertyName: string,
         value: string
-      ): void => {
+      ) => {
         this.setState((prevState) => {
           let {markerStyles} = prevState;
           const last = markerStyles.length - 1;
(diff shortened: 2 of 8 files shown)


moog16 (changes requested):
@cpboyd thanks for adding this component! It would help greatly if you were to add a screenshot test for me to test/poke the code. Could you please add one? 

Also I will be asking for unit tests, but that can happen after the code review. How does that sound?

cpboyd:
@moog16 I just fixed some lint and typing issues.

I'm reading up on the screenshot tests at the moment.

cpboyd:
@moog16 I added a screenshot test, but I'm still not completely clear on everything involved.

I saw the list in **screenshot-test-urls.tsx**, but adding `'slider'` there didn't seem to add it to the test.  Is that in the Docker image itself?

Also, `switch` and `ripple` screenshot tests don't seem to currently be running either.
They're also not in the **screenshot-test-urls.tsx**.

Is there a way for me to view and validate the "golden" image? And where am I supposed to submit that, since [the instructions](https://github.com/material-components/material-components-web-react/blob/master/docs/screenshot-tests.md#step-2-run-a-docker-container) seem to indicate that I can't upload them via the Docker.

moog16:
@cpboyd I have to actually create the golden. I'll get around to this today. The instructions is for you to test everything out. I need to update the contribute doc though (as this might be where the confusion is coming from)

packages/slider/index.scss
```diff
@@ -0,0 +1,23 @@
+// The MIT License
+//
+// Copyright (c) 2018 Google, Inc.
```
moog16: 2019

packages/slider/index.tsx
```diff
@@ -0,0 +1,455 @@
...
+        return this.sliderElement.current.getBoundingClientRect();
+      },
+      /**
+       * Returns the tab index of the slider Element
+       */
+      getTabIndex: (): number => this.props.tabIndex || 0,
+      /**
+       * Registers an event handler on the root element for a given event.
+       */
+      registerInteractionHandler: (
```
moog16: We omit the register/deregister methods in mdc react and just use synthetic event handling. Please see https://github.com/material-components/material-components-web-react/blob/master/packages/floating-label/index.tsx#L94 as an example

packages/slider/index.tsx
```diff
@@ -0,0 +1,455 @@
...
+  }
+
+  private elementByClassName(className: string): HTMLElement | null {
+    if (!this.sliderElement.current) {
+      return null;
+    }
+    // need to use getElementsByClassName since tabIndicator could be
+    // a non-semantic element (span, i, etc.). This is a problem since refs to a non semantic elements
+    // return the instance of the component.
+    return this.sliderElement.current.getElementsByClassName(
```
moog16: Can you use ref's instead of this?

packages/slider/package.json
```diff
@@ -0,0 +1,28 @@
+{
+  "name": "@material/react-slider",
+  "version": "0.12.0",
```
moog16: 0.0.0

moog16 (changes requested):
Getting warning in screenshots:
Screen Shot 2019-05-29 at 10 10 15 AM

Looks like you need to remove this from the render method's otherProps

packages/slider/index.tsx
cpboyd: Do I need to re-implement [these handlers](https://github.com/material-components/material-components-web/blob/master/packages/mdc-slider/foundation.ts#L147-L153) in React?

cpboyd:
@moog16 Commit aa0a057 basically replicates a substantial portion of the [mdc-slider/foundation.ts](https://github.com/material-components/material-components-web/blob/master/packages/mdc-slider/foundation.ts)

Further, I left the `registerBodyInteractionHandler` as I'm not sure there's a more React-like way to add handlers to the document body.

cpboyd:
I guess the rebase to master confused @googlebot.

moog16:
Changed the base to rc0.14.0

packages/slider/index.tsx
```diff
@@ -0,0 +1,677 @@
...
+    const {className, discrete, displayMarkers} = this.props;
+    return classnames('mdc-slider', Array.from(classList), className, {
+      'mdc-slider--discrete': discrete,
+      'mdc-slider--display-markers': discrete && displayMarkers,
+    });
+  }
+
+  get adapter(): MDCSliderAdapter {
+    return {
+      /**
```
moog16: can you remove all the adapter comments? I appreciate the thoroughness however it makes the adapter 200 lines.

packages/slider/index.tsx
```diff
@@ -0,0 +1,677 @@
...
+        }
+      },
+      /**
+       * Sets a style property of the thumb container element to the passed value
+       */
+      setThumbContainerStyleProperty: (
+        propertyName: string,
+        value: string
+      ): void => {
+        this.setStylePropertyOn(
```
moog16: do we know that this won't be overridden on a rerender? This updates the style property on the DOM Node. I'm not sure if it will persist.

packages/slider/index.tsx
```diff
@@ -0,0 +1,677 @@
...
+          this.thumbContainerElement.current,
+          propertyName,
+          value
+        );
+      },
+      /**
+       * Sets a style property of the track element to the passed value
+       */
+      setTrackStyleProperty: (propertyName: string, value: string): void => {
+        this.setStylePropertyOn(this.trackElement.current, propertyName, value);
```
moog16: Here too...does this persist?

packages/slider/index.tsx
```diff
@@ -0,0 +1,677 @@
...
+   */
+  private setInTransit(inTransit: boolean) {
+    this.inTransit = inTransit;
+    this.toggleClass(cssClasses.IN_TRANSIT, this.inTransit);
+  }
+
+  /**
+   * Conditionally adds or removes a class based on shouldBePresent
+   */
+  private toggleClass(className: string, shouldBePresent: boolean) {
```
moog16: we shouldn't be calling adapter methods in our component.

packages/slider/index.tsx
```diff
@@ -0,0 +1,677 @@
...
+    this.foundation.setValue(value);
+
+    if (this.inTransit) {
+      this.onThumbEnd = () => {
+        this.setInTransit(false);
+        this.onThumbEnd = undefined;
+      };
+    }
+
+    if (shouldFireInput) {
```
moog16: We shouldn't be using adapter methods, but instead relying on the foundation.

packages/slider/index.tsx
```diff
@@ -0,0 +1,677 @@
...
+    };
+
+    const moveEventType = MOVE_EVENT_MAP[downEvent.type as DownEventType];
+
+    // Note: upHandler is [de]registered on ALL potential pointer-related release event types, since some browsers
+    // do not always fire these consistently in pairs.
+    // (See https://github.com/material-components/material-components-web/issues/1192)
+    const upHandler = () => {
+      this.handleUp();
+      this.adapter.deregisterBodyInteractionHandler(moveEventType, moveHandler);
```
moog16: This should be handled by the foundation. The general rule is that we need to implement all adapter methods (as you've done so above), and use all foundation methods. If we don't use the foundation methods, then that needs to be commented. 

Please refer to the vanilla web component to check this.

packages/slider/index.tsx
```diff
@@ -0,0 +1,677 @@
...
+  onKeyDown = (evt: React.KeyboardEvent) => {
+    const keyId = this.getKeyId(evt);
+    const value = this.getValueForKeyId(keyId);
+    if (isNaN(value)) {
+      return;
+    }
+
+    // Prevent page from scrolling due to key presses that would normally scroll the page
+    evt.preventDefault();
+    this.adapter.addClass(cssClasses.FOCUS);
```
moog16: You should most likely be calling foundation.handleDown_. I know this is a private method, however you can refer to  https://github.com/material-components/material-components-web/issues/2813 in a comment when using a private event handler method from the foundation in this slider package.

packages/slider/index.tsx
```diff
@@ -0,0 +1,677 @@
...
+          ref={this.thumbContainerElement}
+          onMouseDown={this.onThumbDown}
+          onPointerDown={this.onThumbDown}
+          onTouchStart={this.onThumbDown}
+          onTransitionEnd={this.onThumbEnd}
+        >
+          {discrete && (
+            <div className='mdc-slider__pin'>
+              <span className='mdc-slider__pin-value-marker'>
+                {markerValue}
```
moog16: The markerValue in the demo is a float/has many decimal places. I believe this is a bug. Please address

moog16 (changes requested):

TroyTae:
#258

packages/slider/index.tsx
cpboyd: I just based it on the [tab-indicator](https://github.com/material-components/material-components-web-react/blob/master/packages/tab-indicator/index.tsx#L118).

I'll look at refactoring this.

cpboyd:
@moog16 How do you suggest I handle the `onTransitionEnd` within 
[updateUIForCurrentValue_](https://github.com/material-components/material-components-web/blob/master/packages/mdc-slider/foundation.ts#L507)?

This gets called by `setValue_`, and thus most of the events.

packages/slider/index.tsx
cpboyd: I don't see this when rendering in Chrome, are you saying it has decimal places in the screenshot?

moog16:
@cpboyd I see what you're saying about updateUIForCurrentValue. What if you do implement `[de]registerThumbContainerInteractionHandler`. You can set a flag to true in the register and set the flag to false in the deregister. Then in this.onThumbEnd, say:

```
onThumbEnd: () => {
  if (!this.flag) return;
  // ...
}
```

packages/slider/index.tsx
moog16: Yes, as you move the slider it displays decimal places.

cpboyd:
@moog16 That's effectively what I tried to do in https://github.com/material-components/material-components-web-react/pull/868/commits/b333d05c08152eb700c4a6fd62d3e7e1881b783c

packages/slider/index.tsx
cpboyd: This is what I see in Chrome on macOS:
image

If it's still happening after I removed the `setValue` function in https://github.com/material-components/material-components-web-react/pull/868/commits/92d43bec99de0cdcda193cca2fefb78a3722d86d, then I can try to add code to truncate the value displayed by React.

I guess we'll have to account for the fact that the step itself can be a floating point number.

moog16:
ah you're right...Sorry, it was a lot to sift through. But what you did makes sense.

moog16:
Is this ready for review again?

cpboyd:
@moog16 I rebased with rc0.14.0.

Should I also squash the slider commits down?

moog16:
no need! We'll squash merge this puppy when its ready. I'll take anotehr look. Thanks!

packages/slider/index.tsx
```diff
@@ -0,0 +1,380 @@
+import {getCorrectEventName} from '@material/animation/util';
+import {EventType, SpecificEventListener} from '@material/base/types';
+import {MDCSliderAdapter, MDCSliderFoundation} from '@material/slider';
+import classnames from 'classnames';
+import React from 'react';
+
+const convertDashToCamelCase = (propName: string) =>
+  propName.replace(/-(\w)/g, (_, v) => v.toUpperCase());
+
+type MouseLikeEvent = React.MouseEvent | React.PointerEvent | React.TouchEvent;
```
moog16: Let's call this InteractionEvent

packages/slider/index.tsx
```diff
@@ -0,0 +1,380 @@
...
+  tabIndex?: number;
+  isRtl?: boolean;
+  onValueChange?: (value: number) => void;
+  onValueInput?: (value: number) => void;
+}
+
+interface SliderState {
+  classList: Set<string>;
+  markerValue?: number;
+  markerStyles: React.CSSProperties[];
```
moog16: why is this an array, but the others are not?

packages/slider/index.tsx
```diff
@@ -0,0 +1,380 @@
...
+  trackStyleProperty: React.CSSProperties;
+  thumbContainerStyleProperty: React.CSSProperties;
+}
+
+type SliderElementNames = 'trackStyleProperty' | 'thumbContainerStyleProperty';
```
moog16: SliderElementNames --> ElementStyleNames

packages/slider/index.tsx
```diff
@@ -0,0 +1,380 @@
...
+
+class Slider extends React.Component<SliderProps, SliderState> {
+  get classes() {
+    const {classList} = this.state;
+    const {className, discrete, displayMarkers} = this.props;
+    return classnames('mdc-slider', Array.from(classList), className, {
+      'mdc-slider--discrete': discrete,
```
moog16: can you use the constants? `cssClasses.DISCRETE`, as well as below

packages/slider/index.tsx
```diff
@@ -0,0 +1,380 @@
...
+            top: 0,
+            width: 0,
+            x: 0,
+            y: 0,
+          };
+          return defaultDOMRect;
+        }
+        return this.sliderElement.current.getBoundingClientRect();
+      },
+      getTabIndex: (): number => this.props.tabIndex || 0,
```
moog16: This defaults to 0. Change this to `number -> this.props.tabIndex`

packages/slider/index.tsx
```diff
@@ -0,0 +1,380 @@
...
+      // React handles events differently:
+      registerInteractionHandler: () => undefined,
+      deregisterInteractionHandler: () => undefined,
+      registerThumbContainerInteractionHandler: (
+        evtType: EventType,
+        handler: SpecificEventListener<EventType>
+      ): void => {
+        if (evtType === this.transitionendEvtName) {
+          this.onThumbEnd = handler as any;
```
moog16: can you change `any` to a specific event listener?

packages/slider/index.tsx
```diff
@@ -0,0 +1,380 @@
...
+        this.setState({markerValue: value}),
+      appendTrackMarkers: (numMarkers: number): void => {
+        this.setState((prevState) => {
+          let {markerStyles} = prevState;
+          // Shallow copy of array:
+          markerStyles = markerStyles.slice();
+          for (let i = 0; i < numMarkers; i++) {
+            markerStyles.push({});
+          }
+          return {markerStyles};
```
moog16: I don't think you need to return here.

moog16: Type in adapter is `appendTrackMarkers(numMarkers: number): void;`

packages/slider/index.tsx
```diff
@@ -0,0 +1,380 @@
...
+        });
+      },
+      removeTrackMarkers: (): void => {
```
moog16: remove this void...its unnecessary

packages/slider/index.tsx
```diff
@@ -0,0 +1,380 @@
...
+          const clone = Object.assign({}, markerStyles[last]);
+          // we need to cast prop from string (interface requirement) to React.CSSProperties;
+          const typedProp = propertyName as keyof React.CSSProperties;
+          // https://github.com/Microsoft/TypeScript/issues/11914
+          clone[typedProp] = value;
+          markerStyles[last] = clone;
+          return {markerStyles};
+        });
+      },
+      isRTL: (): boolean => this.props.isRtl || false,
```
moog16: || false is unnecessary as you have it defined in defaultProps.

packages/slider/index.tsx
```diff
@@ -0,0 +1,380 @@
...
+    const {step, value, min, max, disabled} = this.props;
+    this.foundation.setDisabled(disabled!);
+    this.foundation.setStep(step!);
+    this.foundation.setValue(value!);
+    this.foundation.setMin(min!);
+    this.foundation.setMax(max!);
+    this.foundation.layout();
+    this.foundation.setupTrackMarker();
+
+    this.onDown = (evt) => (this.foundation as any).handleDown_(evt);
```
moog16: Why do you need to cast as `any`?

packages/slider/index.tsx
```diff
@@ -0,0 +1,380 @@
...
+import {MDCSliderAdapter, MDCSliderFoundation} from '@material/slider';
+import classnames from 'classnames';
+import React from 'react';
+
+const convertDashToCamelCase = (propName: string) =>
+  propName.replace(/-(\w)/g, (_, v) => v.toUpperCase());
+
+type MouseLikeEvent = React.MouseEvent | React.PointerEvent | React.TouchEvent;
+
+export interface SliderProps extends React.HTMLProps<HTMLDivElement> {
```
moog16: Is it more appropriate for the HTMLDivElement to be HTMLInputElement?

packages/slider/index.tsx
```diff
@@ -0,0 +1,380 @@
...
+
+    return (
+      <div
+        className={this.classes}
+        role='slider'
+        aria-valuemin={min}
+        aria-valuemax={max}
+        aria-valuenow={value}
+        data-step={step}
+        aria-label='Select Value'
```
moog16: we should allow for this to be overridable:

```
aria-label={this.props.ariaLabel}
```

set as a defaultProp to 'Select Value'

moog16 (changes requested):
Hey looking good! Two things I noticed:

- the onValueChange doesn't fire when I programatically change `value` on one of the screenshot tests. I think there isn't a way to achieve this with the current foundation. 
- also need unit tests. We aim for 95% coverage. If you run npm run test:unit, this will build a coverage directory. If you open the index.html file within there, it will show you the coverage.

packages/slider/index.tsx
cpboyd: Hmm.  That's a good point.  The type of `step`, `min`, `max`, and `value` would need to change to `string`, but it might be good from a uniformity standpoint.

I think I'd considered `HTMLInputElement`, but went with `HTMLDivElement` because it doesn't actually use an `<input>` element, so we could potentially end up passing `HTMLInputElement`-specific props to the `<div>` via `otherProps` unless they're explicitly removed in the `props` destructuring.

packages/slider/index.tsx
cpboyd: To eliminate the Typescript complaints in VS Code about not knowing the `private` function.

It seems the `lint` check doesn't care, though, so I can remove it... I was just trying to resolve the majority of red underlines in VS Code.

packages/slider/index.tsx
cpboyd: Should I remove the other void return specifications?

packages/slider/index.tsx
cpboyd: Had to force with `this.props.tabIndex!` since `HTMLDivElement` marks `tabIndex` as optional.

packages/slider/index.tsx
cpboyd: Removed optional types for props per https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-0.html#support-for-defaultprops-in-jsx

packages/slider/index.tsx
cpboyd: I'll cast it to `() => void` since that's the type of the function.

cpboyd:
@moog16 I think I've addressed all the new comments.
I'll work on the unit tests.

moog16:
Great! Let me know when its ready for review :)

cpboyd:
@moog16 I rebased to rc0.15 and added some unit tests.

Note: I'm currently using `@material/slider` 2.3.0.  I'm not overly familiar with `material-components-web` versioning.  Some of the React components are still using the mdc v1 versions.  Should I downgrade (or optionally upgrade to v3)?

Issue #873, opened 2019-05-21
[TextField] autoFocus label is not floated


maloguertin:
## Bug report
The TextField component does not behave properly when its Input children has `autoFocus` prop.

### Steps to reproduce
```jsx
<TextField label="foo">
 <Input autoFocus/>
</TextField>
```
### Actual behavior
The input has the focus but the label is not floated, nor does it float uppon entering text.

### Expected behavior
The input should be floated.

### Screenshots
![Screenshot from 2019-05-21 14-29-20](https://user-images.githubusercontent.com/6407064/58121222-e1051280-7bd4-11e9-9fcc-c6bf83cbfeb1.png)

### Your Environment:

| Software         | Version(s) |
| ---------------- | ---------- |
| MDC Web         | 2.2.0  |
| Browser         |  Chrome (latest)|
| Operating System | Ubuntu 19 |

### Possible solution
When using `autofocus="true"` instead of `autoFocus` it works, it might be because the adapter uses the attribute `autofocus="true"` and React does not add the attribute, it only calls `.focus()` on mount. 

Also doing something like:
```jsx
<TextField label="foo">
 <Input ref={ref=>ref.inputElement.focus()}/>
</TextField>
```
yields the same result. Probably because the `handleFocusChange` prop passed to the InputComponent only sets the isFocused state but doesn't trigger anything on the adapter?

moog16:
@maloguertin please see stackBlitz https://stackblitz.com/edit/react-afnzkv?file=TextField.js. Let me know if this is still an issue for you. Also you said you're on MDC Web version 2.2.0. This is the MDC React repo. Also you're missing a `d` in `<TextField ...`.

Closing as its not an issue, but feel free to ping this issue. I will respond.

maloguertin:
It works on stackBlitz but no matter what I do can't get it to work on my build. Also I said MDC Web because were using it alongside the MDC React Repo (I have version 0.12.1 for the textfield).

maloguertin:
Also if I alt-tab twice to leave and come back to the page then the label is properly floated

moog16:
created a fix targeted for rc0.13.0

moog16:
Closed in #875

PR #875, opened 2019-05-22
 fix(text-field): autofocus floats label

moog16:
fixes https://github.com/material-components/material-components-web-react/issues/873


packages/text-field/index.tsx
```diff
@@ -284,7 +284,15 @@ class TextField<
     const {props} = child;
     return Object.assign({}, props, {
       foundation: this.state.foundation,
-      handleFocusChange: (isFocused: boolean) => this.setState({isFocused}),
+      handleFocusChange: (isFocused: boolean) => {
+        this.setState({isFocused});
+        if (!this.state.foundation) return;
+        if (isFocused) {
+          this.state.foundation.activateFocus();
```
abhiomkar: Why do we have to have foundation as state? We can make text field not accept foundation prop to keep it simple.

moog16: we actually depend on the foundation in the `<Input />` component, and sometimes require a rerender when the foundation updates or is created -- hence the state. Are you saying to not have `<Input />` accept the `foundation` prop?

packages/text-field/Input.tsx
```diff
@@ -144,6 +144,13 @@ export default class Input<
       setDisabled,
     } = this.props;
 
+    if (
```
abhiomkar: This condition looks complicated :) 

Do you need to check the previous state too?

moog16: You need to check previous state because you only want this to trigger when both are set. This only occurs when one was changed from unset --> set.

moog16: And yeah...it is complicated, but it is a common react pattern to check previous state or props.

packages/text-field/Input.tsx
```diff
@@ -178,6 +185,23 @@ export default class Input<
...
+  private initializeComponentWithFoundation = () => {
+    const {handleFocusChange, foundation, autoFocus, isValid} = this.props;
+    if (autoFocus) {
+      handleFocusChange && handleFocusChange(true);
+    }
+    // there is no reason for this to be in Input.tsx
+
+    if (foundation && isValid !== undefined) {
+      foundation.setUseNativeValidation(false);
+      foundation.setValid(!!isValid);
```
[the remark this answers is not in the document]
moog16: ah true...good catch!

PR #877, opened 2019-05-23
fix(select): componentDidUpdate handles disabled prop change

gugu:


packages/select/index.tsx
```diff
@@ -148,6 +148,9 @@ export default class Select<
     if (this.state.foundation && this.state.value !== prevState.value) {
       this.state.foundation.handleChange(true);
     }
+    if (this.state.foundation && this.props.disabled !== prevProps.disabled) {
```
moog16: This looks good - can you please add a unit test :)

test/unit/select/index.test.tsx
```diff
@@ -53,6 +53,12 @@ test('state.value updates when props.value changes', () => {
   assert.equal(wrapper.state().value, updatedValue);
 });
 
+test('state.disabled updates when props.disabled changes', () => {
```
moog16: Your test should test your conditional. Please test that foundation.setDisabled() is called.

PR #883, opened 2019-05-25
fix(chips): error with undefined or null children

### LINKED ISSUE #880: ChipSet errors when rendered without children
The following error occurs if the `ChipSet` component renders without any children:

image

lucasecdb:
fixes #880

### DIFF
--- a/packages/chips/package.json
+++ b/packages/chips/package.json
@@ -29,6 +29,7 @@
     "access": "public"
   },
   "devDependencies": {
+    "@types/react": "^16.8.19",
     "@types/react-is": "^16.7.1"
   }
 }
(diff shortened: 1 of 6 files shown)


packages/chips/package.json
```diff
@@ -22,9 +22,13 @@
...
     "classnames": "^2.2.6",
-    "react": "^16.4.2"
+    "react": "^16.4.2",
+    "react-is": "^16.8.6"
   },
   "publishConfig": {
     "access": "public"
+  },
+  "devDependencies": {
+    "@types/react-is": "^16.7.1"
```
moog16: we should also have @types/react here as well. I have an issue to add it to all components, but until we get to that let's add it here too.

lucasecdb: done!

lucasecdb: I added to all packages in #898

moog16: oh nice!

Issue #885, opened 2019-05-26
Upgrade mdc web dependencies to 2.x


lucasecdb:
these are the components that need to be upgraded:

- [x] `@material/react-button`
- [x] `@material/react-card`
- [x] `@material/react-checkbox`
- [x] `@material/react-chips`
- [x] `@material/react-dialog`
- [x] `@material/react-drawer`
- [x] `@material/react-fab`
- [x] `@material/react-icon-button`
- [ ] `@material/react-list` **has breaking change**
- [ ] `@material/react-menu` **has breaking change**
- [ ] `@material/react-radio`
- [ ] `@material/react-ripple`
- [ ] `@material/react-select`
- [ ] `@material/react-snackbar`
- [ ] `@material/react-switch`
- [ ] `@material/react-tab-bar`
- [ ] `@material/react-tab-indicator`
- [ ] `@material/react-tab-scroller`
- [ ] `@material/react-tab`
- [x] `@material/react-text-field`
- [ ] `@material/react-top-app-bar` **has breaking change**

fjorgemota:
Now that 0.14 is released, I want to warn one thing related directly to this issue (which is the reason behind I decided to comment instead of creating a new issue): You should attempt to use 0.13 version instead of upgrading to 0.14 **if** you're using a mdc's react package that was updated + a mdc's react package that was not updated.

For example, I'm using `@material/react-list` and `@material/react-drawer` on a project. After updating to version `0.14`, importing these two libraries inside one project built by webpack means that `@material/list` will be duplicated inside the bundle, because:

1) Webpack cannot tree-shake the library; (given that `@material/list`'s foundation.js only export `MDCListFoundation`)
2) Different versions of the same library will be imported, so webpack cannot import the library just once (see below the `yarn why @material/list` output on my project);

``` > yarn why @material/list
yarn why v1.16.0
[1/4] Why do we have the module "@material/list"...?
[2/4] Initialising dependency graph...
[3/4] Finding dependency...
[4/4] Calculating file sizes...
=> Found "@material/list@2.3.0"
info Has been hoisted to "@material/list"
info Reasons this module exists
   - Hoisted from "@material#react-drawer#@material#list"
   - Hoisted from "@material#react-drawer#@material#drawer#@material#list"
info Disk size without dependencies: "1.5MB"
info Disk size with unique dependencies: "3.16MB"
info Disk size with transitive dependencies: "3.28MB"
info Number of shared dependencies: 10
[...]

PR #886, opened 2019-05-26
chore(button): upgrade to mdcw 2.x

lucasecdb:
related to #885


lucasecdb:
I think the golden for this needs to be updated, since the border width was fixed in material-components/material-components-web#4606

moog16:
I ran the tests locally and they seemed to pass.

PR #891, opened 2019-05-26
chore(drawer): upgrade to mdcw 2.x

lucasecdb:
related to #885


packages/drawer/package.json
```diff
@@ -17,7 +17,7 @@
     "url": "https://github.com/material-components/material-components-web-react.git"
   },
   "dependencies": {
-    "@material/drawer": "^1.0.1",
+    "@material/drawer": "^2.1.1",
     "@material/list": "^1.0.0",
```
moog16: I think we should also update list here.

moog16:
@lucasecdb I just checked https://github.com/material-components/material-components-web-react/issues/885...it looks like list has a breaking change. What is it that we need to fix?

lucasecdb:
I think it's more like a ["usability" breaking change](https://github.com/material-components/material-components-web/blob/master/CHANGELOG.md#breaking-changes), the API is still the same, so I think we are good

moog16:
Yes it looks like this was the PR:
https://github.com/material-components/material-components-web/pull/4568/files.
We are good!

moog16:
i just rebased with master, and it messed this PR up :(

lucasecdb:
oh let me try to fix it

moog16:
can you update once more? I had to rebase again.

moog16:
@lucasecdb pinging the PR :)

lucasecdb:
I think I lost some commits in between those force pushes :-(

moog16:
Ah yeah :( Sorry I made this so complicated.

lucasecdb:
its ok, git is kinda hard when you are trying to rebase into a branch that has been updated

PR #894, opened 2019-05-27
feat(list): add ripple to list-item

### LINKED ISSUE #385: Add ripple to List Item

TroyTae:
resolves #385

There are two lines that uses `any` type 😢 
If you have suggestion, tell me about that.


TroyTae:
OMG I didn't change target branch to rc0.13.0
Please change base branch 😅

packages/list/ListItem.tsx
```diff
@@ -79,9 +82,12 @@ export class ListItemBase<T extends HTMLElement = HTMLElement> extends React.Com
 
   componentDidMount() {
     this.initializeTabIndex();
+    if (this.props.initRipple) {
+      (this.props.initRipple as any)(this.listItemElement.current);
```
moog16: Try updating React Ripple's initRipple type to:
```
  initRipple?: (surface: S | null, activator?: A | null) => void;
```

I don't think the React.Ref is necessary.

TroyTae: I tried to remove `React.Ref`, but there are some code using that.
So I overwrite initRipple in props!

PR #898, opened 2019-05-29
chore: add react types package to all packages

### LINKED ISSUE #569: add @types/react to packages

lucasecdb:
resolves #569

### DIFF
--- a/packages/button/package.json
+++ b/packages/button/package.json
@@ -24,5 +24,8 @@
   },
   "publishConfig": {
     "access": "public"
+  },
+  "devDependencies": {
+    "@types/react": "^16.8.19"
   }
 }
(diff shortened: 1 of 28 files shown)


packages/button/package.json
```diff
@@ -24,5 +24,8 @@
   },
   "publishConfig": {
     "access": "public"
+  },
+  "devDependencies": {
+    "@types/react": "^16.8.19"
```
moog16: should this match the react version? (16.4.2) Or is it ok to have the newer types involved?

moog16: A ton of unit tests typings are failing - might be related to versioning

moog16:
please rebase with rc0.14.0

PR #899, opened 2019-05-30
Rc0.13.0

moog16:


moog16:
CLA is unnecessary as all the other PRs have been signed.

PR #903, opened 2019-06-04
fix: npm vulnerabilities

TroyTae:
Fix vulnerabilities which is reported from npm.
There is one more about `@google-cloud/storage`.
But I don't have any test environment for it.
If you aren't busy, can you update `@google-cloud/storage` and fix some breaking changes?

***ADD***
karma also has breaking change in ci environment.
(Local test(win/mac) is fine)
So I reverted karma/karma-webpack version 😅

***Breaking Changes***
[bundle].min.js is removed in generated files.
You need to change:
`
import Button from '@material/react-button/dist/index.min.js';` -> `import Button from '@material/react-button/dist/index.js';
`


packages/webpack.config.js
```diff
@@ -138,11 +137,6 @@ function getJavaScriptWebpackConfig() {
         },
       ],
     },
-    plugins: [
-      new webpack.optimize.UglifyJsPlugin({
```
moog16: why did you remove this?

moog16:
Getting a few warnings in `npm run build`:

```
    WARNING in configuration
    The 'mode' option has not been set, webpack will fallback to 'production' for this value. Set 'mode' option to 'development' or 'production' to enable defaults for each environment.
    You can also set it to 'none' to disable any default behavior. Learn more: https://webpack.js.org/configuration/mode/

```

I also see:

```
 Entrypoint undefined = extract-text-webpack-plugin-output-filename
```

Can you look into what this is? This wasn't happening before.

packages/webpack.config.js
TroyTae: This is breaking changes in webpack v4.
We can configure related configuration in `optimization`.
(https://webpack.js.org/configuration/optimization/)
In production mode, webpack4 minize bundle file by default.
So I remove this line :)

TroyTae:
@moog16 
Ah thank you for notifying this.
I configured mode with `production`.
And it looks we need to use `mini-css-extract-plugin` instead of `extract-text-webpack-plugin`.
I reflect it!

packages/webpack.config.js
moog16: I'm going to test this, but I don't want to break anyone by changing the the file name

moog16: It looks like both the .min.js and .js files are the same, which means it doesn't break anyone. I think we should eventually deprecate the .min.js files.

TroyTae: You're right.
I'll configure to not generate min file!

moog16: Cool - this will be a breaking change for people...Can you make sure to add that in the PR description.

TroyTae: I reflect it!
If necessary, we can also remove css.min files.

moog16:
can you update with rc0.14.0 please!

TroyTae:
@moog16 
rebase complete!

Issue #931, opened 2019-06-13
[Bug]: Ripple doesn't activate when clicked if it was previously activated with touch


Arthie:

moog16:
I see your PR and will continue the convo there. Thanks!

moog16:
Closing with #932

PR #932, opened 2019-06-13
(fix:ripple) not activating when clicked after touch

Arthie:
[Bug]: Ripple does not activate when clicked if it was previously activated with touch
fixes: #931

### DIFF
--- a/packages/ripple/index.tsx
+++ b/packages/ripple/index.tsx
@@ -264,6 +264,8 @@ export function withRipple<
       this.props.onMouseDown && this.props.onMouseDown(e);
       if (!this.isTouched) {
         this.activateRipple(e);
+      } else {
+        this.isTouched = false;
       }
     };
 
--- a/test/unit/ripple/index.test.tsx
+++ b/test/unit/ripple/index.test.tsx
@@ -75,6 +75,22 @@ test('mouseDown event triggers activateRipple', () => {
   mockRaf.restore();
 });
 
+test('mouseDown sets isTouched to false if isTouched is true', () => {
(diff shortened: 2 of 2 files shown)


moog16:
How do you reproduce this? I just want to test this.

Arthie:
@moog16 I have a touchscreen laptop makes it easy.
But u can simulate touch events with dev tools when u enable responsive / mobile view.
So to reproduce:
1) go in mobile view with dev tools,  click the button. (this sets the state of `isTouched` to `true`)
2) exit mobile view, click the button again and ripple wont activate. (`isTouched` is never set to `false` again)

moog16:
Change looks good! Could you please add a unit test?

Arthie:
> Change looks good! Could you please add a unit test?

I tried to add the test but after struggling to install the dependencies and when I finally was able to run the tests I got: `(chromium-browser:18737): Gtk-WARNING **: 15:03:11.370: cannot open display:` because the linux environment (WSL) has no gui interface. Really frustrating. I gave up after that not worth spending more time on it.

Hope you understand, I am not able to add tests.

moog16:
@Arthie thanks for trying...I'll add a test and post it here for you to commit. Once that happens let's merge it!

Arthie:
@moog16 Thanks! I have added the test.

moog16:
Arthie you added it to the Button test. It should instead be added to `test/unit/ripple/index.test.tsx`

Arthie:
@moog16 Yea, sorry 😅 fixed it and cleaned the commit history.

PR #934, opened 2019-06-15
fix(text-field): label isn't float when set value with setState

### LINKED ISSUE #895: Floating label on text-field does not move when value prop changes.
image

TroyTae:
fixes #895

**breaking changes**
`handleValueChange` is removed from Input's props.

### DIFF
--- a/packages/text-field/Input.tsx
+++ b/packages/text-field/Input.tsx
@@ -21,14 +21,12 @@
 // THE SOFTWARE.
 import React from 'react';
 import classnames from 'classnames';
-import {MDCTextFieldAdapter} from '@material/textfield/adapter';
 import {MDCTextFieldFoundation} from '@material/textfield/foundation';
 
 export interface InputProps<T extends HTMLElement = HTMLInputElement> {
   className?: string;
   inputType?: 'input' | 'textarea';
   isValid?: boolean;
-  adapter?: MDCTextFieldAdapter;
   foundation?: MDCTextFieldFoundation;
   handleValueChange?: (
     value: string | number | string[] | undefined,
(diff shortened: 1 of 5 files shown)


TroyTae:
Ah I see, then I'll close this PR..!

packages/text-field/Input.tsx
```diff
@@ -163,16 +166,24 @@ export default class Input<
     }
 
     if (value !== prevProps.value) {
```
moog16: Let's reopen it, cause I think this is where the change should happen. I think this `if` block should look like this: 
```
if (value !== prevProps.value) {
      const inputValue = this.valueToString(value);
      const notTriggeredChange = !this.state.wasUserTriggeredChange;
      // only call #foundation.setValue on programatic changes;
      // not changes by the user.
      if (notTriggeredChange) {
        foundation && foundation.setValue(inputValue);
      }

      this.setState({wasUserTriggeredChange: false});
    }
```

packages/text-field/index.tsx
```diff
@@ -287,6 +287,7 @@ class TextField<
     const {props} = React.Children.only(this.props.children);
 
     return Object.assign({}, props, {
+      adapter: this.adapter,
```
moog16: remove the adapter from here

packages/text-field/Input.tsx
TroyTae: I reflected it!
Is handleValueChange a breaking change?
I saw and removed a test code that checks call count of handleValueChange.

moog16: Can you remove handleValueChange from everywhere? https://github.com/material-components/material-components-web-react/pull/934/files#diff-0eddc9611a42aba741f3ee96c819757eR31

This is a "breaking change" however no one should have been using this.

moog16:
please remove handlevalueChange from all tests and source code:  https://github.com/material-components/material-components-web-react/blob/c5ad27e913a8002892d58949d588e5b8fc8c1a10/test/unit/text-field/Input.test.tsx#L183

TroyTae:
@moog16 
I reflected it!
I guess we need to descript about this in CHANGELOG.md :)

moog16:
Changelog is generated - so not worries about CHANGELOG. Changes look good!

Issue #936, opened 2019-06-17
Fix Broken Typescript Production Types


arichiv:
The generated `.d.ts` files have been broken for a couple months now. There are two main errors:
1. [All ripple based components improperly include a relative ripple file that does not exist in production installs](https://github.com/material-components/material-components-web-react/issues/857)
2. Inconsistent imported versions of `@types/react-is` and `@types/react` are causing templating type errors. These can be resolved by making package versions consistent.

See current errors below. [Once fixed, the production types test can be restored to Travis CI](https://github.com/material-components/material-components-web-react/blob/master/.travis.yml)

```
node_modules/@types/react-is/index.d.ts:20:51 - error TS2314: Generic type 'ReactElement<P>' requires 1 type argument(s).
20 export function isAsyncMode(value: any): value is ReactElement;
                                                     ~~~~~~~~~~~~
node_modules/@types/react-is/index.d.ts:21:57 - error TS2314: Generic type 'ReactElement<P>' requires 1 type argument(s).
21 export function isContextConsumer(value: any): value is ReactElement;
                                                           ~~~~~~~~~~~~
node_modules/@types/react-is/index.d.ts:22:57 - error TS2314: Generic type 'ReactElement<P>' requires 1 type argument(s).
22 export function isContextProvider(value: any): value is ReactElement;
                                                           ~~~~~~~~~~~~
node_modules/@types/react-is/index.d.ts:23:49 - error TS2314: Generic type 'ReactElement<P>' requires 1 type argument(s).
23 export function isElement(value: any): value is ReactElement;
                                                   ~~~~~~~~~~~~
node_modules/@types/react-is/index.d.ts:24:52 - error TS2314: Generic type 'ReactElement<P>' requires 1 type argument(s).
24 export function isForwardRef(value: any): value is ReactElement;
                                                      ~~~~~~~~~~~~
node_modules/@types/react-is/index.d.ts:25:50 - error TS2314: Generic type 'ReactElement<P>' requires 1 type argument(s).
25 export function isFragment(value: any): value is ReactElement;
                                                    ~~~~~~~~~~~~
node_modules/@types/react-is/index.d.ts:28:50 - error TS2314: Generic type 'ReactElement<P>' requires 1 type argument(s).
28 export function isProfiler(value: any): value is ReactElement;
                                                    ~~~~~~~~~~~~
[...]

PR #937, opened 2019-06-17
[WIP] Make Package Versions Consistent

arichiv:
Part of #936


arichiv:
Blocked on #939

arichiv:
This rebase will be drastic enough that it'd be better to restart

Issue #781, opened 2019-03-27
Interface 'ButtonProps<T>' cannot simultaneously extend types 'AnchorHTMLAttributes<T>' and 'ButtonHTMLAttributes<T>'


dawsonc623:
Due to a [change in `@types/react`](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/34194), the definition of `ButtonProps` does not work anymore. I have not dived into exactly what it would take to fix it, but the only workaround there is right now is for users to stick to `@types/react@16.8.8` or lower in their project's dependencies.

ben-mckernan:
The easy fix would be to change it to use `HTMLProps` like it is done in icon-button.
https://github.com/material-components/material-components-web-react/blob/rc0.12.0/packages/icon-button/index.tsx#L45-L46

A more complex approach would be to have a conditional generic like what is done for input in text-field.
https://github.com/material-components/material-components-web-react/blob/rc0.12.0/packages/text-field/Input.tsx#L44-L47

moog16:
@ben-mckernan yes you're spot on. It should be extending HTMLProps instead of HTMLAttributes. text-field has a good example of this.

I'm not sure what the exact issue is with extending both button and anchor, but conceptually you can't be both at the same time (so this doesn't make sense). If someone would like to open a PR updating the types feel free.

Jason3S:
The issue is with the `type` property in both interfaces. They are not of the same type, so extends doesn't know how to merge them.

moog16:
Right but this only works for the Button case...if you were using the anchor tag you would need to change this. This issue is to make sure developers can import ButtonProps and it works for both cases. :)

PR #939, opened 2019-06-18
Re-Build Lock File

arichiv:
In order to change package versions the package-lock should be re-built. This breaks a lot of things, so it's been split out as an independent change. #936


arichiv:
Okay, so it looks like it has been a while since this was re-built. I'll try playing around with the package versions a bit, but at some point it's not worth putting off the work of just sitting on edge (especially for `@types/react`).

arichiv:
@moog16 it's alive! The most interesting change here is locking `@types/react` and `react` at `16.7.X` due to [issues with `16.8.X`](https://github.com/material-components/material-components-web-react/issues/781).

arichiv:
@moog16 I'm actually not updating them (much)! The locked-in versions were much newer than the requested version as `^16.0.0` asks for the latest `16.x` release while `~16.0.0` asks for the latest `16.0.x` release.

If you would prefer, I can lock in the older `react(-dom)` versions, but updating to match the `@types/*` packages already in use (by time of release) seemed to make sense.

### `react`
Did ask for `^16.4.2` and locked `16.4.2`.
Will ask for `~16.7.0` and locks `16.7.0`.

### `@types/react`
Did ask for `^16.4.4` and locked `16.7.17`.
Will ask for `~16.7.22` and locks `16.7.22`.

### `react-dom`
Did ask for `^16.4.2` and locked `16.4.2`.
Will ask for `~16.7.0` and locks `16.7.0`.

### `@types/react-dom`
Did ask for `^16.0.11` and locked `16.0.11`.
Will ask for `~16.0.11` and locks `16.0.11`.

moog16:
@arichiv interesting...so you're saying we're "downgrading"? Why not just stick to the version we're on?

arichiv:
@moog16 the table above indicates all packages will be locked at newer (or identical) versions. None have been downgraded.

What is changing is that the primary `package.json` (not requirements for sub-packages) is no longer indicating compatibility the repo can't support. As an example, in `@types/react` we indicate `^16.4.4` is supported, which means any and all versions through the current `16.8.20` work. This is a lie, as there are numerous type errors with the `16.8.x` branch. This hadn't been obvious before as the `package-lock.json` file was forcing npm to pull in `16.7.17`. With this change the `package.json` properly indicates that we support up through `16.7.x` and the version locked in is updated to `16.7.22`.

moog16:
@arichiv gotcha - thanks for fixing those! Looks like there is a conflict with package-lock again. I'll approve once fixed and I can pull in the changes to take a look myself.

packages/chips/Chip.tsx
```diff
@@ -70,8 +70,8 @@ export class Chip extends React.Component<ChipProps, ChipState> {
   };
 
   state = {
-    classList: new Set(),
-    leadingIconClassList: new Set(),
+    classList: new Set<string>(),
```
moog16: are these required? Isn't this inferred by ChipState?

arichiv: Turns out that they aren't anymore :-/

moog16: Ah can't we set the type of the object by doing:

`state: ChipState = {...}`

arichiv:
This diff broke types due to an upgrade to the typescript package. https://github.com/material-components/material-components-web-react/pull/952 is the fix

PR #943, opened 2019-06-20
fix(ripple): focus style alive when disabled

### LINKED ISSUE #916: Button retains mdc-ripple-upgraded--background-focused style even though it's not focused
I have three buttons that I'm using as a search results view type button bank (grid view, list view, map view).

When a particular view type is active, the corresponding `Button` is marked as `disabled` so the user can't reselect it (it's already been applied).

## Expected behaviour
When a user clicks an option, the button relating to that option should be highlighted, other buttons should lose their highlight.

## Actual behaviour
Initial load with list view active:
Screenshot 2019-06-07 at 15 51 40

I think that disabled is the wrong thing. What I really want is a a button group where the buttons work as a radio buttons (eg: https://getbootstrap.com/docs/4.1/components/buttons/#checkbox-and-radio-buttons)
So my first problem with this is that the colours are wrong. The selected one should be highlighted, and the others should be secondary in this case.

It doesn't feel like there's a paradigm for that at the moment...?

After first click:
Screenshot 2019-06-07 at 15 51 50

The selected button retains its focus style, even if I attribute focus to a different element on the screen.

After second click:
Screenshot 2019-06-07 at 15 51 57

The first click-button _still_ retains its focused style.

## Code
```
        <div className="search-tools__format">
          <Button icon={<MaterialIcon icon="map" />} disabled={format === "map"} onClick={e => {changeViewFormat("map");}} />
          <Button icon={<MaterialIcon icon="view_list" />} disabled={format === "list"} onClick={e => {changeViewFormat("list");}} />
          <Button icon={<MaterialIcon icon="view_module" />} disabled={format === "grid"} onClick={e => {changeViewFormat("grid");}} />
        </div>
```

TroyTae:
fixes #916

@poying
I wrote it pretty much like your code.
But I'm not call `handleBlur` and I made `disabledRipplePostProcessor`.
Because there are some possibility that adds other code in`handleBlur`.
If you have any other opinion, tell me about that!

@moog16 
I added `adapter` to a member variable.
Existing member variables do not have an `adapter`, is there any special reason??

If both of you are okay, I'll write unit test code :)

### DIFF
--- a/packages/ripple/index.tsx
+++ b/packages/ripple/index.tsx
@@ -29,6 +29,8 @@ import {supportsCssVariables, applyPassive} from '@material/ripple/util';
 import {SpecificEventListener} from '@material/base/types';
 import {matches} from '@material/dom/ponyfill';
 
+const cssClasses = MDCRippleFoundation.cssClasses;
+
 export interface RippledComponentProps<T> {
   unbounded?: boolean;
   disabled?: boolean;
@@ -110,6 +112,7 @@ export function withRipple<
       RippledComponentState
     >
     implements RippledComponentInterface<Surface, Activator> {
+    adapter!: MDCRippleAdapter;
     foundation!: MDCRippleFoundation;
     isComponentMounted: boolean = true;
     isTouched: boolean = false;
@@ -146,6 +149,12 @@ export function withRipple<
       }
     }
 
+    componentDidUpdate(prevProps: P) {
+      if (this.props.disabled !== prevProps.disabled && this.props.disabled) {
+        this.disabledRipplePostProcessor();
+      }
+    }
+
     componentWillUnmount() {
       if (this.foundation) {
         this.isComponentMounted = false;
@@ -157,8 +166,8 @@ export function withRipple<
     // activator: This element is used to detect whether to activate the ripple. If this is not
     // provided, the ripple surface will be used to detect activation.
     initializeFoundation = (surface: Surface, activator?: Activator) => {
-      const adapter = this.createAdapter(surface, activator);
-      this.foundation = new MDCRippleFoundation(adapter);
+      this.adapter = this.createAdapter(surface, activator);
+      this.foundation = new MDCRippleFoundation(this.adapter);
       this.foundation.init();
     };
 
@@ -323,6 +332,10 @@ export function withRipple<
           style: updatedStyle,
         });
       });
+    };
+
+    disabledRipplePostProcessor = () => {
+      this.adapter.removeClass(cssClasses.BG_FOCUSED);
     };
 
     get classes() {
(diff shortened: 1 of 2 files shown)


packages/ripple/index.tsx
```diff
@@ -325,6 +334,10 @@ export function withRipple<
       });
     };
 
+    disabledRipplePostProcessor = () => {
+      this.adapter.removeClass(cssClasses.BG_FOCUSED);
```
moog16: We shouldn't be calling the adapter directly, which is why we didn't have it on the component. I think we can actually just call handleBlur.

TroyTae: I reflected it and write some test code!

test/unit/ripple/index.test.tsx
```diff
@@ -5,7 +5,9 @@ import {MDCRippleFoundation} from '@material/ripple/foundation';
 import {MDCRippleAdapter} from '@material/ripple/adapter';
 import {SpecificEventListener} from '@material/base/types';
 import {mount} from 'enzyme';
-import {withRipple, InjectedProps} from '../../../packages/ripple/index';
+
+import Button from '../../../packages/button';
+import {withRipple, InjectedProps} from '../../../packages/ripple';
```
moog16: do you need this import here? It looks unused.

moog16:
Tested this by using part of the stackblitz in the original issue:

```

class ButtonScreenshotTest extends React.Component<ButtState, {}> {

  state = {format: 'map'};
  render() {
    const {format} = this.state;

    return (
      <div>
    <div className="search-tools__format">
      <Button unelevated={format === "map"}
        icon={<MaterialIcon icon="map" />}
        disabled={format === "map"}
        onClick={() => {this.setState({ format: "map"})}} />
      <Button unelevated={format === "list"}
        icon={<MaterialIcon icon="view_list" />}
        disabled={format === "list"}
        onClick={() => {this.setState({ format: "list"})}} />
      <Button unelevated={format === "grid"}
        icon={<MaterialIcon icon="view_module" />}
        disabled={format === "grid"}
        onClick={() => {this.setState({ format: "grid"})}} />
    </div>
)
}
}
```

Looks like its working!

test/unit/ripple/index.test.tsx
TroyTae: Yeah this is just my habit.
I always separate node_module and local_module imports!
Does it look better to put it back??

TroyTae: Maybe I misunderstood your comment.
This code already existed before.
`import {withRipple, InjectedProps} from '../../../packages/ripple';`
I'm just moved it down.

moog16: ah! Sorry I didn't notice that. Yes that was there before. Looks good.

PR #945, opened 2019-06-21
fix(list): Only apply aria-orientation with role

### LINKED ISSUE #944: List applies invalid default aria-orientation attribute
[mdc-list documentation](https://github.com/material-components/material-components-web/tree/master/packages/mdc-list#accessibility) says

> Use role="listbox" only for single selection list, without this role the ul element is implicitely role="list". Do not use aria-orientation attribute for standard list (i.e., role="list"), use component's vertical property to set the orientation to vertical.

The React wrapper applies `aria-orientation="vertical"` in default props: https://github.com/material-components/material-components-web-react/blob/master/packages/list/index.tsx#L117

I think the ideal fix is to refactor the `aria-orientation` prop to `orientation` (enum) or `vertical` (bool) and apply this as `aria-orientation` for allowed role values. But, this would be a breaking change.

This could be fixed without renaming props but it would be a little strange to pass an aria prop that is not passed to the root element. Maybe edgy enough not to worry about.

liamcmitchell-sc:
Fixes #944


packages/list/index.tsx
```diff
@@ -146,9 +146,9 @@ export default class List extends React.Component<ListProps, ListState> {
     if (wrapFocus !== prevProps.wrapFocus) {
       this.foundation.setWrapFocus(wrapFocus!);
     }
-    if (this.props[ARIA_ORIENTATION] !== prevProps[ARIA_ORIENTATION]) {
+    if (this.props.orientation !== prevProps.orientation) {
       this.foundation.setVerticalOrientation(
-        this.props[ARIA_ORIENTATION] === VERTICAL
+        this.props.orientation !== HORIZONTAL
```
moog16: Why not use the affirmative here? Meaning `=== VERTICAL`?

packages/list/index.tsx
```diff
@@ -129,7 +129,7 @@ export default class List extends React.Component<ListProps, ListState> {
     }
     this.foundation.setWrapFocus(wrapFocus!);
     this.foundation.setVerticalOrientation(
-      this.props[ARIA_ORIENTATION] === VERTICAL
+      this.props.orientation !== HORIZONTAL
```
moog16: Please leave a comment here as to why you're not using the affirmative (`=== VERTICAL`)

test/unit/list/index.test.tsx
```diff
@@ -434,6 +432,43 @@ test('renders with role=menu if props.role=menu and props.checkboxList', () => {
...
+  const wrapper = shallow<List>(<List>{children}</List>);
+  assert.equal(wrapper.props()['aria-orientation'], undefined);
+});
+
+test('renders without aria-orientation when checkboxList and orientation=vertical (default)', () => {
+  const wrapper = shallow<List>(<List checkboxList>{children}</List>);
+  assert.equal(wrapper.props()['aria-orientation'], undefined);
+});
+
+test('renders with aria-orientation when checkboxList and orientation=horizontal', () => {
```
moog16: Can you add an equivalent test, but swap `checkboxList` --> `radioList`

liamcmitchell-sc:
Done. Yes, tested this in my project. Works as expected.

moog16:
Thanks!

PR #948, opened 2019-06-23
feat(list): Implement getAttributeForElementIndex

### LINKED ISSUE #822: List implement adapter.getAttributeForElementIndex
Missing adapter method from update to mdc web v1.0.0

TroyTae:
fixes #822

### DIFF
--- a/packages/list/index.tsx
+++ b/packages/list/index.tsx
@@ -217,15 +217,16 @@ export default class List extends React.Component<ListProps, ListState> {
       getListItemCount: () => this.listElements.length,
       getFocusedElementIndex: () =>
         this.listElements.indexOf(document.activeElement as HTMLLIElement),
+      getAttributeForElementIndex: (index, attr) => {
+        const listItem = this.listElements[index];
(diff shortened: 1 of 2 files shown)


packages/list/index.tsx
```diff
@@ -217,15 +217,16 @@ export default class List extends React.Component<ListProps, ListState> {
       getListItemCount: () => this.listElements.length,
       getFocusedElementIndex: () =>
         this.listElements.indexOf(document.activeElement as HTMLLIElement),
+      getAttributeForElementIndex: (index, attr) => {
+        const listItem = this.listElements[index];
```
moog16: does this also need an if (listItem) check like `setAttributeForElementIndex`?

moog16: :ping_pong:

TroyTae: Ah i forgot this ;)
I didn’t add check code because error message is very clear. ({index}th element is not exist)
If we catch that, I guess some of developers cannot know cause of error!

PR #952, opened 2019-06-25
 fix(infrastructure): types generated directory problem

TroyTae:
- fix lint/compile error
 - set "declarationDir" in tsconfig
 - add "types" property in dialog
 - add validate code that to check existence of definition file

### DIFF
--- a/scripts/release/verify-pkg-main.js
+++ b/scripts/release/verify-pkg-main.js
@@ -37,15 +37,13 @@ if (!isValidCwd) {
 }
 
 let invalidMains = 0;
+let invalidTypes = 0;
 globSync('packages/*/package.json').forEach((jsonPath) => {
   const packageInfo = JSON.parse(fs.readFileSync(jsonPath));
-  if (!packageInfo.main) {
-    return;
-  }
-
   let isInvalid = false;
+  let isTypesInvalid = false;
 
-  const mainPath = path.join(path.dirname(jsonPath), packageInfo.main);
+  const mainPath = path.join(path.dirname(jsonPath), packageInfo.main || '');
   if (mainPath.indexOf('dist') < 0) {
     isInvalid = true;
     console.error(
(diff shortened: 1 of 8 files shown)


scripts/release/verify-pkg-main.js
```diff
@@ -31,37 +31,66 @@ const isValidCwd = fs.existsSync('packages');
...
     invalidMains++;
   }
+  if (isTypesInvalid) {
+    invalidTypes++;
+  }
 });
-
-if (invalidMains > 0) {
-  console.error(`${invalidMains} incorrect main property values found; please fix.`);
+if (invalidMains > 0 || invalidTypes > 0) {
```
moog16: @TroyTae I updated this - let me know if you disagree with anything.

test/unit/menu/MenuList.test.tsx
```diff
@@ -3,7 +3,7 @@ import {assert} from 'chai';
 import * as td from 'testdouble';
 import {shallow, mount} from 'enzyme';
 import {MenuList} from '../../../packages/menu/index';
-import List from '../../../packages/list/index';
+import List from '../../../packages/list';
```
moog16: I don't think we should be doing this...I see the innerRef error, but this is indicating to us that there is an error with our types

scripts/release/verify-pkg-main.js
TroyTae: I agree to declare invalidTypes.
But we can remove isTypesInvalid and use isInvalid..!

moog16: There are 2 different messages though.

test/unit/menu/MenuList.test.tsx
TroyTae: Ah I see, then let's fix this another PR!

scripts/release/verify-pkg-main.js
TroyTae: I got it now!

PR #953, opened 2019-06-25
fix(ripple) broken relative imports

arichiv:
The issue seems to be at the webpack stage; we need to prevent it looking across packages.
This should be the final fix for #936


moog16:
@arichiv great! Thanks for driving this home. I was going to do a release today. I think I'm going to go ahead with it. This can be a patch release.

arichiv:
For sure! I can rebase on whatever, and the changes from this could use longer to bake IMO

moog16:
agreed! sounds good

arichiv:
Looks like it's not webpack, it's in the typescript build itself

packages/button/index.tsx
```diff
@@ -22,11 +22,13 @@
 
 import React from 'react';
 import classnames from 'classnames';
-import {withRipple, InjectedProps} from '@material/react-ripple';
+import {withRipple, InjectedProps, RippledComponentProps} from '@material/react-ripple';
```
arichiv: @moog16 I traced this to a typescript error not unlike https://github.com/microsoft/TypeScript/pull/24874 (cc: @weswigham)

If we do not explicitly import `RippledComponentProps` the resulting [typefile](https://termbin.com/w4ez) attempts a relative import at the bottom (`import("../ripple").RippledComponentProps`).

If we do explicitly import `RippledComponentProps` (and use/export it as required to avoid lint errors) the resulting [typefile](https://termbin.com/6ow4) is able to directly use the imported type (`RippledComponentProps`).

This appears to be an issue in typescript resolution/expansion itself. The import at the top is left alone, but the generated one at the bottom will default to a relative import.

arichiv: We can wait for a fix in the typescript repo (I do not have the time to dig into that) or add these dummy type exports to allow proper import of `RippledComponentProps` in all 10 affected packages (see this file for an example).

moog16: The TS PR that you linked should already be in the TS library, no? 

I think we'll be better off adding the line you added to all the components needing it.

packages/button/index.tsx
```diff
@@ -22,11 +22,13 @@
...
 
 import {CSS_CLASSES} from './constant';
 
 type ButtonTypes = HTMLAnchorElement | HTMLButtonElement;
+// TODO(https://github.com/material-components/material-components-web-react/pull/953)
+export type removeOnceTypeResolutionIsFixed = RippledComponentProps<ButtonTypes>;
```
moog16: Types should be capitalized: `remove` --> `Remove`.

Also please create an issue to remove this when the TS version is update. Its not clear to me what we're waiting for, or if this is a new TS bug.

arichiv:
@moog16 ready!

moog16:
When I run the tests it creates some files. Please add `./test/types/index.ts` and `./test/types/tsconfig.json` to the `.gitignore`. Looks good though :)

arichiv:
> When I run the tests it creates some files. Please add `./test/types/index.ts` and `./test/types/tsconfig.json` to the `.gitignore`. Looks good though :)

Done!

PR #966, opened 2019-07-07
chore: catalog example

TroyTae:
yeeeeeeeeeeeeeeeeeeeeee


TroyTae:
Oh example directory is also test target!
Is there something that I need to do additionally?

examples/catalog/src/Reset.scss
```diff
@@ -0,0 +1,77 @@
+html, body, div, span, applet, object, iframe,
```
moog16: How did you come up with this list? I think you only need `html, body`

examples/catalog/src/Reset.scss
```diff
@@ -0,0 +1,77 @@
...
+table, caption, tbody, tfoot, thead, tr, th, td,
+article, aside, canvas, details, embed,
+figure, figcaption, footer, header, hgroup,
+menu, nav, output, ruby, section, summary,
+time, mark, audio, video {
+  margin: 0;
+  padding: 0;
+  border: 0;
+  font-size: 100%;
+  font: inherit;
```
moog16: Isn't this by default?

examples/catalog/src/Reset.scss
```diff
@@ -0,0 +1,77 @@
...
+footer, header, hgroup, menu, nav, section {
+  display: block;
+}
+body {
+  font-family: Roboto, sans-serif;
+}
+ol, ul {
+  list-style: none;
+}
+blockquote, q {
```
moog16: Do you use blockquotes?

examples/catalog/src/Reset.scss
```diff
@@ -0,0 +1,77 @@
...
+}
+table {
+  border-collapse: collapse;
+  border-spacing: 0;
+}
+a {
+  text-decoration: none;
+}
+select {
+  appearance: none;
```
moog16: Do you actually need this? Shouldn't mdc-web's styles already have this.

examples/catalog/src/Template.html
```diff
@@ -0,0 +1,12 @@
+<html>
```
moog16: Template.html --> can you make the name less generic? It's not descriptive, since everything can be thought of as a "template"

examples/catalog/src/content/button/ButtonExample.tsx
```diff
@@ -0,0 +1,38 @@
...
+        <Button outlined={true}>outlined</Button>
+      </Cell>
+      <Cell>
+        <Button dense={true}>dense</Button>
+      </Cell>
+      <Cell>
+        <Button icon={<MaterialIcon icon={'stars'} />}>icon</Button>
+      </Cell>
+      <Cell>
+        <Button trailingIcon={<MaterialIcon icon={'cached'} />}>
```
moog16: change to `icon='cached'`. Same with stars above.

examples/catalog/src/content/button/ButtonExample.tsx
```diff
@@ -0,0 +1,38 @@
...
+          trailing-icon
+        </Button>
+      </Cell>
+      <Cell>
+        <Button disabled={true}>disabled</Button>
```
moog16: This can just be `<Buttonst disabled>...`

examples/catalog/src/content/button/ButtonExample.tsx
```diff
@@ -0,0 +1,38 @@
...
+import {Cell, Grid, Row} from '@material/react-layout-grid';
+
+export const ButtonExample = () => (
+  <Grid>
+    <Row>
+      <Cell>
+        <Button>default</Button>
+      </Cell>
+      <Cell>
+        <Button raised={true}>raised</Button>
```
moog16: You can remove `={true}` from all of these

examples/catalog/src/content/card/CardExample.scss
```diff
@@ -0,0 +1,4 @@
+:local .container {
```
moog16: what is this :local?

examples/catalog/src/content/checkbox/CheckboxExample.tsx
```diff
@@ -0,0 +1,49 @@
...
+        />
+        <label htmlFor='checkbox2' className={styles.label}>
+          indeterminate
+        </label>
+      </Cell>
+      <Cell className={styles.checkbox}>
+        <Checkbox
+          nativeControlId='checkbox3'
+          className={styles.check}
+          disabled={true}
```
moog16: remove the `true` arg

examples/catalog/src/content/radio/RadioExample.tsx
```diff
@@ -0,0 +1,36 @@
...
+      </Cell>
+      <Cell>
+        <Radio label='Vue' key='vue'>
+          <NativeRadioControl name='framework' value='vue' id='vue' />
+        </Radio>
+      </Cell>
+      <Cell>
+        <Radio label='Disabled' key='disabled'>
+          <NativeRadioControl
+            disabled={true}
```
moog16: disabled

examples/catalog/src/content/select/SelectExample.tsx
```diff
@@ -0,0 +1,108 @@
...
+class DefaultSelect extends React.Component<any> {
+  state = {value: 'TurkishAir'};
+
+  render() {
+    return (
+      <Select
+        {...this.props}
+        label='Choose Airplane'
+        value={this.state.value}
+        onChange={(evt) =>
```
moog16: Can you move this into its own class method. I saw you did this in a few other places. It is more performant to be a class method.

examples/catalog/src/menu/Menu.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+import {MenuContext, UserContext} from '../Context';
+import {Menus} from '../Router';
+
+export const Menu = () => (
+  <MenuContext.Consumer>
+    {({isOpen, setOpen, selectedIndex, setSelectedIndex}) => (
+      <UserContext.Consumer>
+        {({name, account}) => (
+          <Drawer modal open={isOpen} onClose={() => setOpen(false)}>
```
moog16: please move onClose to its own method outside of the Drawer

examples/catalog/tsconfig.json
```diff
@@ -0,0 +1,25 @@
+{
```
moog16: Can you just use the same tsconfig in the root?

examples/catalog/src/App.tsx
```diff
@@ -0,0 +1,87 @@
...
+          isOpen: isOpen,
+          selectedIndex: selectedIndex,
+          setOpen: this.setOpen,
+          setSelectedIndex: this.setSelectedIndex,
+        }}
+      >
+        <UserContext.Provider
+          value={{
+            name: 'Troy Tae',
+            account: 'tjy970721@gmail.com',
```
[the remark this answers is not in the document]
TroyTae: Oops

examples/catalog/src/App.tsx
```diff
@@ -0,0 +1,87 @@
...
+          }}
+        >
+          <Menu />
+          <TopBar />
+          <Content />
+          <Snackbar
+            timeoutMs={10000}
+            message='You became a frogrammer!'
```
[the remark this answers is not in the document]
TroyTae: I heard it from my friend it's a joke for programmer haha
But it looks not good meaning
I'll change it :)

examples/catalog/src/Reset.scss
TroyTae: Yes some browser is not default this value

examples/catalog/src/Reset.scss
TroyTae: I follow this guide from Eric Meyer's reset.css
Of course I don't use all of tag.
But I didn't touch it because it looks like a library file.

examples/catalog/src/Reset.scss
TroyTae: image
After removing this line.
Native Select appears strange 😢

examples/catalog/src/content/card/CardExample.scss
TroyTae: This keyword generate random className!

TroyTae:
@TroyTae 
It's better to exclude `examples` directory in tsconfig 😅

examples/catalog/src/App.tsx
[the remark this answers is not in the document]
moog16: what is the meaning? I might just live under a rock

examples/catalog/src/Reset.scss
moog16: hmm - seems like an mdc web bug. HOWEVER I think we're in the process of removing native select. We can ignore it for now.

examples/catalog/src/content/card/CardExample.scss
moog16: This is a sass function?

examples/catalog/tsconfig.json
moog16: I was thinking you could remove this file and reference the one in root.

examples/catalog/src/App.tsx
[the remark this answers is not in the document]
TroyTae: http://www.danielsen.com/jokes/frog.txt :)

examples/catalog/src/content/card/CardExample.scss
TroyTae: 👍

examples/catalog/.gitignore
```diff
@@ -0,0 +1,71 @@
...
+
+# Recycle Bin used on file shares
+$RECYCLE.BIN/
+
+# Windows Installer files
+*.cab
+*.msi
+*.msix
+*.msm
+*.msp
```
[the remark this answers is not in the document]
TroyTae: I generate this file in here: https://gitignore.io/
And I write custom ignore list at the bottom.
I cannot sure which is better...
Maybe you want to maintain small for readability, right?

examples/README.md
```diff
@@ -5,3 +5,4 @@ A collection of small apps or pages using MDC React components.
 Example | Link
 --- | ---
 Roses | [./roses](./roses)
+Catalog | [./catalog](./catalog)
```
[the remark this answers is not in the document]
TroyTae: > * I need to download/clone the repo then run `npm start` to see the catalog, but since the package.json deps needs to be bumped manually on each release (instead of referencing the packages in their respective paths since this is a monorepo) I might be looking at other components than the ones in the repo I just downloaded/cloned.
>   
>   * There would be more use for developers on this repo if it referenced the packages in this repo, so they can see how their changes look (I'd love that!)

It's a good idea!
But there is one worried thing.
If we referenced package.json of root directory, this example isn't always stable.
For example, now `Roses` example used version "^0.5.x".
Maybe this example isn't working with latest version.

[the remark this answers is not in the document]
TroyTae: >   * If I run `npm build`, what am I expected to do with the output?
>     
>     * Could this be put online somehow? I guess this is a question to the maintainers and is a much bigger endeavor.

I exported `index.html`/`main.[hash].js`/`[some static files]` in `dist` directory.
Of course you can serve this files in web server.
Just copy this directory and serve html file.
But now I'm not configured about routing system.
I'll make a next.js example as soon as possible.
That example makes more clear about deploy!

examples/catalog/tsconfig.json
TroyTae: Then I'll extends tsconifg.json and overwrite some properties.
Because I don't want to include some properties 😅

moog16: sounds good!

examples/README.md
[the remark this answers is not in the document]
moog16: Yeah agreed - it could be out of date like Roses. We haven't had the resources to focus on the examples. I think having each example as a smaller sub project makes sense since it requires minimal maintenance down the road.

examples/catalog/src/content/card/CardExample.scss
moog16: Gotcha - I've used that before but not in production. Great feature!

examples/catalog/src/menu/Menu.tsx
moog16: @TroyTae did you have a reason for not wanting to do this?

examples/catalog/src/App.tsx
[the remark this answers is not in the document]
moog16: @TroyTae can you please change it to something else :)

examples/catalog/src/Reset.scss
moog16: I agree @seriema...the user(developer using mdc) shouldn't have to do this to a component. But since we're deprecating the native select. I don't think we'll have time to get to it.

examples/README.md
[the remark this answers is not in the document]
moog16: I think publishing it is fairly easy, but it requires maintenance that I don't think the team can take on. I think this is actually a common pattern (having to clone/download the repo and running `npm run start`).

examples/catalog/src/menu/Menu.tsx
TroyTae: Oh where is my comment?
I replied like this:
```
Can you tell me what is the benefit when we split this method to outside?
I written like this because I want to leave it as expression.
To split method, we need to make it a statement.
```

examples/catalog/src/App.tsx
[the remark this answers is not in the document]
TroyTae: I already change it to programmer!

examples/catalog/src/Reset.scss
TroyTae: Then I'll delete this line and remove Native Select.

examples/catalog/src/menu/Menu.tsx
moog16: Oh weird we lost that comment.

So if you keep it as an inline function like that, it will rerender every time. If you move it to a named function outside of the render method, it will not rerender because of the method.

examples/catalog/src/top-bar/TopBar.tsx
```diff
@@ -0,0 +1,44 @@
...
+  <MenuContext.Consumer>
+    {({setOpen}) => (
+      <TopAppBar>
+        <TopAppBarRow>
+          <TopAppBarSection align='start'>
+            <TopAppBarIcon navIcon tabIndex={0}>
+              <MaterialIcon
+                hasRipple
+                icon='menu'
+                onClick={() => setOpen(true)}
```
moog16: move this out to its own function as well.

examples/catalog/webpack.config.js
```diff
@@ -0,0 +1,71 @@
+const {CleanWebpackPlugin} = require('clean-webpack-plugin');
+const HtmlWebpackPlugin = require('html-webpack-plugin');
+const MiniCssExtractPlugin = require('mini-css-extract-plugin');
+const Path = require('path');
```
moog16: can you make this lowercase `path` - I'm just used to it that way.

moog16:
Getting an error when i run:

```
rm -rf node_modules
npm i
npm start
```

```ERROR in /usr/local/google/home/mattgoo/Documents/material-components-web-react/examples/catalog/tsconfig.json
[tsl] ERROR
      TS18003: No inputs were found in config file 'tsconfig.json'. Specified 'include' paths were '["**/*"]' and 'exclude' paths were '["../../node_modules","../../examples","../../build"]'.

ERROR in ./src/App.tsx
Module build failed (from ./node_modules/ts-loader/index.js):
Error: error while parsing tsconfig.json
```

examples/catalog/src/menu/Menu.tsx
TroyTae: > If you move it to a named function outside of the render method, it will not rerender because of the method.

Can you explain to me about this mean?
I knew that React's Virtual Dom will detect what has changed and reflect only changed part.

TroyTae:
@moog16 
I fixed it!

examples/catalog/src/top-bar/TopBar.tsx
TroyTae: This is also same issue with above, right?

examples/catalog/src/menu/Menu.tsx
moog16: It detects an anonymous function as a new function each render. So it should instead be like this:

```
const myFunc = () => {
  // do something
}

...


render() {
  return (
    <MyComp onSomething={myFunc} />
  );
}
```

examples/catalog/src/top-bar/TopBar.tsx
moog16: correct

examples/catalog/src/menu/Menu.tsx
TroyTae: Ah I see, I'm working on it!

examples/catalog/src/top-bar/TopBar.tsx
TroyTae: I fixed it!

examples/catalog/src/menu/Menu.tsx
```diff
@@ -20,33 +20,35 @@ export const Menu = () => (
...
-            </DrawerHeader>
-
-            <DrawerContent>
-              <List singleSelection selectedIndex={selectedIndex}>
-                {Menus.map((menu, index) => (
-                  <ListItem
-                    key={index}
-                    onClick={() => {
+        {({name, account}) => {
+          const close = () => setOpen(false);
```
moog16: This actually has to be out of the scope of the JSX and render method.

examples/catalog/src/top-bar/TopBar.tsx
```diff
@@ -15,30 +15,29 @@ const styles = require('./TopBar.scss');
...
-            <TopAppBarTitle className={styles.title}>
-              Material Components React
-            </TopAppBarTitle>
-          </TopAppBarSection>
-          <TopAppBarSection align='end'>
-            <a rel='noopener noreferrer' target='_blank' href={Repository}>
-              <TopAppBarIcon actionItem tabIndex={0}>
-                <MaterialIcon aria-label='code' hasRipple icon='code' />
+    {({setOpen}) => {
+      const click = () => setOpen(true);
```
moog16: same here :)

examples/catalog/src/menu/Menu.tsx
TroyTae: But this function used `setOpen`.
And `setOpen` is in the `Context`.
If I move it to out of scope, this method isn't work.

examples/catalog/src/menu/Menu.tsx
moog16: ah gotcha...hmm I guess we can ignore this one for now

examples/catalog/src/content/Content.tsx
```diff
@@ -0,0 +1,15 @@
...
+import {TopAppBarFixedAdjust} from '@material/react-top-app-bar';
+
+import {MenuContext} from '../Context';
+import {Menus} from '../Router';
+
+export const Content = () => (
+  <MenuContext.Consumer>
+    {({selectedIndex}) => (
+      <TopAppBarFixedAdjust>
+        {React.createElement(Menus[selectedIndex].component)}
```
moog16: This seems like an antipattern. Sorry I didn't pick this up before. What is going on here. Is this the recommended usage?

examples/catalog/src/content/chips/ChipsExample.tsx
```diff
@@ -0,0 +1,101 @@
...
+  state = {
+    selectedChipIds: ['choice2'],
+  };
+
+  render() {
+    return (
+      <ChipSet
+        choice
+        selectedChipIds={this.state.selectedChipIds}
+        handleSelect={(selectedChipIds) => this.setState({selectedChipIds})}
```
moog16: Move this handleSelect out into its own method.

examples/catalog/src/content/chips/ChipsExample.tsx
```diff
@@ -0,0 +1,101 @@
...
+  state = {
+    selectedChipIds: ['filter1', 'filter2'],
+  };
+
+  render() {
+    return (
+      <ChipSet
+        filter
+        selectedChipIds={this.state.selectedChipIds}
+        handleSelect={(selectedChipIds) => this.setState({selectedChipIds})}
```
moog16: Move this handleSelect out into its own method.

examples/catalog/src/content/chips/ChipsExample.tsx
```diff
@@ -0,0 +1,101 @@
...
+  };
+
+  render() {
+    return (
+      <div>
+        <label>
+          <input type='text' onKeyDown={this.handleKeyDown} />
+          <span>type & enter & add</span>
+        </label>
+        <ChipSet input updateChips={(chips) => this.setState({chips})}>
```
moog16: Move this updateChips out into its own method.

examples/catalog/src/content/Content.tsx
TroyTae: I don't know this is a recommendation.
But I know that JSX is compiled like this.
```
<MyComponent></MyComponent> => React.createElement(MyComponent, someProps)
```
And this is the doc. (https://reactjs.org/docs/react-without-jsx.html)

TroyTae:
@moog16 
I checked all of files.
And it looks inline function is only in ChipsExample.
Thank you!

examples/catalog/src/content/Content.tsx
moog16: Ok fair enough! :)

moog16:
Looks good!

PR #977, opened 2019-07-22
fix(list): Remove selected & activated from prop spread

liamcmitchell-sc:
`selected` and `activated` are boolean props that toggle class names. They shouldn't be passed down with prop spread. In the case of `activated`, this triggers a React console message.

There are no tests because I didn't see any other tests checking that other props are not rendered.


liamcmitchell-sc:
This react attr warning will show up in unit tests when using `mount` but not `shallow`.

These issues could be caught by using mount instead of shallow and throwing on console error to fail test.

PR #991, opened 2019-08-06
fix(list): meta doesn't work with some element

ZeJ0hn:
Trailing with meta doesn't work with some Element. 
For exemple, if I try to create a a list with some radiobutton (on the right) with this code :
```
<ListItem key={index}>
  <ListItemText primaryText={item} />
  <ListItemMeta meta={<Radio>
    <NativeRadioControl
      name={item}
      {...}
    />
  </Radio>} />
</ListItem>
```
The css class **mdc-list-item__meta** is not present in the source code, so the element is not placed on right.


ZeJ0hn:
@moog16 Hello, 

How can I trigger a rebuild ? I'm not sure that my change makes a build error.

Thank you for your help,

Regards

Issue #933, opened 2019-06-14
feat(snackbar): include dismiss icon and missing foundation method handleActionIconClick


mgr34:
currently the snackbar is missing the ability to add an optional dismiss icon and the related foundation method `handleActionIconClick`. 

>**Dismiss Icon**
Snackbars are intended to dismiss on their own after a few seconds, but a dedicated dismiss icon may be optionally included as well for accessibility purposes. ([[0](https://github.com/material-components/material-components-web/tree/master/packages/mdc-snackbar#dismiss-icon)]

---

I will be moving next week and largely unavailable, but if this is still open I'll try to have something for either 0.14.0/0.15.0 (in case the roadmap is ever updated ;)

moog16:
Great sounds good! Thanks for opening this issue. I think it will probably be waiting for you when you get back. Good luck on your move!

krishnaxv:
Hey! Any plan for this feature in the upcoming release?

mgr34:
@krishnaxv I  do not currently have time to prioritize this at the moment. If someone else would like to implement this feature that would be great! Otherwise, I will try to come back to this when time allows.

emri99:
Hello, I have submitted a pull request adding the dismiss button.
Unfortunately I'm unable to regenerate screenshots to make travis get green, if any one can help me to figure out what I'm doing wrong, this will be highly appreciated.
I post more infos on #react discord channel (tell me if you want me to explain this here)
Thanks

emri99:
Hey, @mgr34 @moog16, could you please point me the direction to achieve this PR ?
Thanks ;)

mgr34:
hey @emri99, I opened the issue and have submitted PR's to this great project in the past, but I am not involved in this project otherwise. @moog16 has been responsive in the past, but I see the project hasn't had much activity lately. Perhaps hop on the discord and see whats happening? Your PR has sat for awhile.

Sorry I couldn't be of more help!

emri99:
I did post a more detailed explanation on the discord channel a week ago, but I didn't get any response about this problem until now. I was on my way to bump subject before trying to get help here ;)
Thanks for your quick response anyway ;)

PR #999, opened 2019-09-16
feat(snackbar): Add dismiss button

emri99:
* Add property `dismissible` (default false) to avoid BC
* Update related test
* Update related screenshot

Related issue: #933


emri99:
@googlebot I signed it!

